PR #159, opened 2019-07-02
Add IRS Plan Pages

cKellyDesign:
This PR creates simple (read unconnected) placeholder pages for new IRS Planning functionality.

@moshthepitt how do you feel about this page architecture with different `pages` for each phase of the Planning lifecycle (eg `new|draft|finalized`)?


moshthepitt:
@cKellyDesign I think we should merge it - but I'll let you do the honors ;-)

Issue #168, opened 2019-07-05
Link Focus Area cell to FA Map


cKellyDesign:
Currently links to both the map view and single FI vew are in the same cell in the Dropdown table. Ideally the cell for the Focus Area will link to the Map view (which will load the FA with active FIs), while the Investigation Name cell will link to the Single FI view.

- [x] Link to Map from Focus Area column
- [x] Remove Map icon/link from Investigation Name column

moshthepitt:
I think this has been overtaken by events i.e. I think this one is now overidden by https://github.com/onaio/reveal-frontend/issues/190

PR #175, opened 2019-07-08
Update Active FI table links to maps

cKellyDesign:
For #168


moshthepitt:
@cKellyDesign I think it is the other way round.  The link under the investigation name should be the one going to the map (because that is an individual investigation) while the focus area lists more than one investigation.  No?

cKellyDesign:
@mberg can you confirm? Should the link to the map view be in the `Focus Area` column or `Investigation Name`?

moshthepitt:
@cKellyDesign  I think this one is now overidden by https://github.com/onaio/reveal-frontend/issues/190.  Should we close this PR?

Issue #144, opened 2019-06-26
IRS Planning - Finalized Plan View


cKellyDesign:
This was discussed as the basic Plan view for Finalized plans.

#### Questions
* Is there a difference between the Plan Assignment view and a basic Plan view?
* Where does Plan Finalization end and Plan Monitoring begin?
* Do we need to consider Syncing in the scenario that a Plan is loaded from a Store is out of date compared to what we get from OpenSRP tasks? Is this even a scenario?

#### Dependencies
- [x] Non-client Plan Store and/or OpenSRP or Superset Plan endpoint

#### Requirements
- [x] Plan data is loaded from Store, OpenSRP, or Superset endpoint
- [ ] Plan drafts are opened in the `draft` view
- [ ] Finalized Plans are opened in the `assign` view
- ~~Assigned Plans are opened in a `monitor` view/module(?)~~

cKellyDesign:
We need to add:

* visual confirmation of success (popup modal? alert banner?)
* visual warning of failure (popup modal? alert banner?)
* re-direct on confirmation of successful save from OpenSRP

cKellyDesign:
Now tracking at https://github.com/onaio/reveal-frontend/issues/308

Issue #137, opened 2019-06-26
View List of All Plans


cKellyDesign:
When navigating to `/irs` we need to render all existing Plans as rows in a table and a button to create a new Plan.

#### Table Columns
- [x] Name - Unique name of the Plan
- [x] Date created - Date_Time of last modified (depends on onaio/canopy#306)
- [x] Status - Draft | Finalized | In Progress | Complete
- ~~Summary* - `[x] houses, [y] resources` (* this depends on Resource Estimation)~~

#### Interactions
- [x] Clicking Plan name navigates to existing Plan page
- [x] Clicking `Create new plan` button navigates to Create Plan page `/irs/new`
- ~~Table is sortable and filterable via column headers (* nice-to-have)~~

moshthepitt:
@cKellyDesign I think for uniformity we need to change the data source of the plan ([these ones](https://github.com/onaio/reveal-frontend/blob/master/src/store/ducks/plans.ts#L138)) from Superset to OpenSRP.  This is informed by [this conversation](https://onaio.slack.com/archives/CBSU8L2SC/p1562932654022100).

PR #178, opened 2019-07-08
IRS Plans List page

cKellyDesign:
PR for #137; starts #144;


moshthepitt:
I think for the purposes of the IRS/FI planning it may not be the bestest idea to get plans from the materialized view.  Much easier imho to use the `plans` table.

Of course this means we either need another "plans" reducer, or some fancy footwork to make it work with one reducer (I prefer it this way).

Issue #186, opened 2019-07-11
Ensure superset user is logged out after logging out


moshthepitt:
When you click the log out button, we need to ensure that we log out of superset by calling the `deAuthZ` method in SupersetConnector

cKellyDesign:
This can simply be accomplished by `fetch`ing `https://reveal-[X].smartregister.org/opensrp/logout.do` from the browser when clicking log out.

moshthepitt:
@FrankApiyo while this works, I think it is not the right way to do this.

A better or more right way to do it would be to extend the [logout component](https://github.com/onaio/js-tools/blob/master/packages/gatekeeper/src/components/logout.tsx) that comes with the gatekeeper package.  One way to do this is to make it take an additional prop that represents the logout URL and then call it.  One benefit of doing it this way is that everyone who uses gatekeeper gets free access to this new feature.

I think you should consider:

- creating an issue to do this on the js-tools repo
- creating an issue to remove this "workaround" from Reveal FE and use an updated gatekeeper package, when ready

Let me know if you agree with what I am saying?

cc: @cKellyDesign @p-netm

FrankApiyo:
This makes sense... I think it should be done. The gatekeeper package should be able to handle all oath2 transactions made by any app that uses it and that includes logout.

Also There could be an enhancement that I just learned that we could add to the logout process where a callback URL is passed to the callback endpoint which it should then redirect to after logout(this ideally should  be our login URL). This will help avoid opening another window. This may not be implemented on openSRP though. I'll find out.

moshthepitt:
@FrankApiyo sounds exciting!  Please create the issues :)

PR #191, opened 2019-07-12
Ensure superset user is logged out after logging out

kahama94:
fix: #186

### DIFF
--- a/src/components/page/Header/Header.tsx
+++ b/src/components/page/Header/Header.tsx
@@ -1,5 +1,6 @@
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { User } from '@onaio/session-reducer';
+import superset from '@onaio/superset-connector';
 import * as React from 'react';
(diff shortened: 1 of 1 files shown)


src/components/page/Header/Header.tsx
```diff
@@ -64,7 +67,15 @@ export class HeaderComponent extends React.Component<HeaderProps, State> {
       isOpen: false,
     };
   }
-
+  public componentWillReceiveProps(nextProps: any) {
```
moshthepitt: We should not do this here.  Instead, we need to supply a custom LogOut component to the `gatekeeper` package.  dditionally, let us try and not use the now deprecated `componentWillReceiveProps`, yes?

cKellyDesign: At the very very least this should be `HeaderProps`, no?

moshthepitt:
Going to close this.  Will be MUCH easier to start afresh when attempting to fix this issue.

Issue #181, opened 2019-07-09
Add more Superset filters


moshthepitt:
- [x] Ensure queries to get jurisdictions from Superset are filtered
- [x] Ensure queries to get tasks from Superset are filtered
- [x] Ensure queries to get goals from Superset are filtered

Issue #154, opened 2019-06-29
Fetch structures from Superset


moshthepitt:
After https://github.com/onaio/reveal-frontend/issues/153 and https://github.com/onaio/js-tools/issues/71 are implemented, we should be able to fetch structures for a particular jurisdiction directly from the structures slice in Superset instead of kind of getting structures through tasks.

- [x] Add structures reducer module
- [x] Get structures from the structures reducer module for use in the focus investigation map component

moshthepitt:
We need to fetch structures from this slice: https://superset.reveal-stage.smartregister.org/superset/slice_json/468

We also need to ensure that every query to that slice is filtered to prevent getting tens of thousands of results

moshthepitt:
Sample structure object from superset:

```
{
    "id": "163650",
    "jurisdiction_id": "3390",
    "geojson": "{"id": "163650", "type": "Feature", "geometry": {"type": "Polygon", "coordinates": [[[32.5395938, -14.1692235], [32.5396213, -14.169247], [32.5395837, -14.1692883], [32.5395563, -14.1692649], [32.5395938, -14.1692235]]]}, "properties": {"uid": "163650", "code": "163650", "name": "163650", "type": "163650", "status": "Active", "version": 0, "server_version": 1545225058891, "jurisdiction_id": "3390", "geographic_level": 4, "effective_end_date": null, "effective_start_date": null}}"
}
```

cKellyDesign:
**Structure.geojson.properties**
```
{
    "uid": "163650", 
    "code": "163650", 
    "name": "163650", 
    "type": "163650", 
    "status": "Active", 
    "version": 0, 
    "server_version": 1545225058891, 
    "jurisdiction_id": "3390", 
    "geographic_level": 4, 
    "effective_end_date": null, 
    "effective_start_date": null
}
````

moshthepitt:
Fixed by https://github.com/onaio/reveal-frontend/pull/195

PR #195, opened 2019-07-15
Structures reducer

kahama94:
fix: #181  #154


kahama94:
```  REACT_APP_SUPERSET_TASKS_SLICE=460
REACT_APP_SUPERSET_STRUCTURES_SLICE=468 ```

**Comment by cKellyDesign** [comment]
@moshthepitt @kahama94 Can this be merged?

**Comment by moshthepitt** [comment]
@cKellyDesign I would really appreciate it if we delay merging for a bit?  I wanted to take another pass at reviewing it if thats okay

File: src/configs/env.ts [hunk 1]
```diff
@@ -49,9 +49,16 @@ export type SUPERSET_JURISDICTIONS_SLICE = typeof SUPERSET_JURISDICTIONS_SLICE;
 export const SUPERSET_STRUCTURES_SLICE = process.env.REACT_APP_SUPERSET_STRUCTURES_SLICE || '0';
 export type SUPERSET_STRUCTURES_SLICE = typeof SUPERSET_STRUCTURES_SLICE;
 
+export const SUPERSET_TASKS_SLICE = process.env.REACT_APP_SUPERSET_TASKS_SLICE || '0';
+export type SUPERSET_TASKS_SLICE = typeof SUPERSET_TASKS_SLICE;
+
 export const SUPERSET_PLANS_TABLE_SLICE = process.env.REACT_APP_SUPERSET_PLANS_TABLE_SLICE || '0';
 export type SUPERSET_PLANS_TABLE_SLICE = typeof SUPERSET_PLANS_TABLE_SLICE;
 
+export const SUPERSET_PLAN_STRUCTURE_PIVOT_SLICE =
```
moshthepitt: Should you add this to the sample .env file?

src/containers/pages/FocusInvestigation/map/active/index.tsx
```diff
@@ -126,21 +142,62 @@ class SingleActiveFIMap extends React.Component<
...
+        async relevantJurisdictions => {
+          if (!relevantJurisdictions) {
+            return new Promise(reject => {
+              reject();
+            });
+          }
+          for (let i = 0; i < relevantJurisdictions.length; i += 1) {
+            const jurId = relevantJurisdictions[i].jurisdiction_id;
+            if (i) {
+              sqlFilterExpression += ' OR ';
```
moshthepitt: Why don't you use an `IN` (e.g. `jurisdiction_id IN (1,2,3)`) here?  It is supposed to be more efficient SQL-wise, and it would mean we don't need this loop here.

src/containers/pages/FocusInvestigation/map/active/index.tsx
```diff
@@ -126,21 +142,62 @@ class SingleActiveFIMap extends React.Component<
...
+          ]);
+          const jurisdictionResults = await supersetFetch(
+            SUPERSET_JURISDICTIONS_SLICE,
+            planJurisdictionSupersetParams
+          );
+          fetchJurisdictionsActionCreator(jurisdictionResults);
+        }
+      );
+      /** define superset params for structures */
+      const structuresparams = superset.getFormData(3000, [{ sqlExpression: sqlFilterExpression }]);
```
moshthepitt: Why not use an `ad hoc filter`?  I think we need to keep some kind of consistency in how we do the filtering - I had imagined this would be that we only use sql expressions when that is necessary, otherwise we always use the `ad hoc filters`.  But I dont mind what we specifically do, as long as we do it consistently the same way in our codebase.

src/store/ducks/structures.ts
```diff
@@ -0,0 +1,134 @@
...
+import SeamlessImmutable from 'seamless-immutable';
+import { FeatureCollection, GeoJSON, UpdateType, wrapFeatureCollection } from '../../helpers/utils';
+
+/** the reducer name */
+export const reducerName = 'structures';
+
+/** Interface for structure.geojson.properties for structure
+ *  as received from the fetch request / superset
+ */
+export interface InitialProperties {
```
moshthepitt: If we renamed this to `Structure` we would not need line 40 below at all.  What's the reason for it being this way?

src/store/ducks/structures.ts
```diff
@@ -0,0 +1,134 @@
...
+  jurisdiction_id: string;
+}
+
+export type Structure = InitialStructure;
+
+// actions
+/** STRUCTURES_SET action type */
+export const STRUCTURES_SET = 'reveal/reducer/structures/STRUCTURES_SET';
+
+/** interface for authorize action */
```
moshthepitt: This docstring is not accurate

src/store/ducks/structures.ts
```diff
@@ -0,0 +1,134 @@
...
+      return state;
+  }
+}
+
+// action creators
+
+/** fetch Structure creator
+ * @param {Structure[]} structuresList - array of structure objects
+ */
+export const setStructures = (structuresList: InitialStructure[] = []): SetStructuresAction => {
```
moshthepitt: Should you name this `fetchStructures`?  Right now the name conflicts with that the docstring is saying.

src/store/ducks/structures.ts
```diff
@@ -0,0 +1,134 @@
...
+          return structure as Structure;
+        }
+      ),
+      structure => structure.id
+    ),
+    type: STRUCTURES_SET,
+  };
+};
+
+// selectors
```
moshthepitt: There are potentially thousands of structures.  All our structure selectors need the ability to filter by at least:

- jurisdiction_id
- type

Please make this change for `getStructuresArray` and `getStructuresFC`

moshthepitt: Do you think we need a `getStructuresById` selector?

kahama94: Am filtering by jurisdiction_id.  from the end point. should I still have a selector that filters by jurisdiction_id and type ?

kahama94: Yah we don't

moshthepitt: Yes but what you are getting from the Redux store is everything.  e.g. if you load 20 plans one after the other you will have structures in the Redux store that belong to more than one jurisdiction, and thus when pulling data out of the store we need to only take what we need.

cKellyDesign:
@kahama94 - I spent some time triaging the Inverted Matrix error and ended up going down a few rabbit holes in trying to figure out what was going wrong. The two commits I've pushed on this branch were minor and might be helpful but the fix itself was in onaio/gisida-react#120.

Please see if you can fix the tests I broke here, otherwise feel free to try running a branch without my last two commits to see if the gisida-react fix is enough to get this working.

kahama94:
@cKellyDesign  Thanks a lot for looking into this issue. The inverted matrix issue was not being experienced on this branch it was on add-icons branch I have cherry picked the two commits to that branch. This branch was to improve how we are get data to render various geometries. I think we should move  inverted matrix related issues to add-icons branch.

Issue #194, opened 2019-07-15
Use Symbol Layers


moshthepitt:
To match the mobile client, we need to use symbol layers for:

- Case confirmation
- Mosquito Collection
- Larval Dipping

githengi:
![breeding](https://user-images.githubusercontent.com/25123762/61277437-d624c380-a7ba-11e9-8caf-d3b7204674dd.png)
![mosquito](https://user-images.githubusercontent.com/25123762/61277455-df159500-a7ba-11e9-82be-39a905c0a000.png)
![icon-index-map](https://user-images.githubusercontent.com/25123762/61277469-e6d53980-a7ba-11e9-93d1-a60b03551ecd.png)

PR #187, opened 2019-07-11
Filter jurisdiction, tasks, structures, goals by plan_id

kahama94:
fix: #181

PR #202, opened 2019-07-17
Add symbol layers as is on the mobile client

kahama94:
fix: #194  #180 #154 #187 
The issue depends on gisida-react v1.2.6


moshthepitt:
> The issue depends on gisida-react v1.2.5 yet to be published

@kahama94 can we publish and update our dependencies in this PR?

kahama94:
> can we publish and update our dependencies in this PR?

@moshthepitt most definitely.

cKellyDesign:
@kahama94 - It looks like the Mosquito Collection layer is rendering both icons and circles, is this intentional? Looks like the circles can be removed if the icons are rendering, and maybe the icons size can increase a little when zooming in.

src/containers/pages/FocusInvestigation/map/active/index.tsx
```diff
@@ -58,18 +61,26 @@ import plansReducer, {
...
+} from '../../../../../store/ducks/structures';
 import tasksReducer, {
   fetchTasks,
   getFCByPlanAndGoalAndJurisdiction,
-  getStructuresFCByJurisdictionId,
   reducerName as tasksReducerName,
   Task,
   TaskGeoJSON,
 } from '../../../../../store/ducks/tasks';
+import { jurisdictions } from '../../../../../store/ducks/tests/fixtures';
```
cKellyDesign: @kahama94 - What's this being used for? Seems like fixtures should stick to tests?

src/helpers/utils.tsx
```diff
@@ -109,13 +109,19 @@ export interface SiteConfigAppMapconfig {
...
+  imageUrl: string;
+}
 
 /** interface to describe Gisida site app configuration object */
 export interface SiteConfigApp {
   accessToken: string;
   apiAccessToken: string;
   appName: string;
   mapConfig: SiteConfigAppMapconfig;
+  mapIcons: MapIcons[];
```
cKellyDesign: @kahama94 - Let's make `mapIcons` optional

Issue #208, opened 2019-07-19
Add OpenSRP Service Module


cKellyDesign:
This service will allow us to connect directly to the OpenSRP server to GET, PUSH, and PUT entities, starting with Plans.

- [x] GET Plans
- [x] GET Plan by Unique ID
- [x] PUSH New Plan
- [x] PUT Updated Plan

Unblocks: #141, #142

PR #215, opened 2019-07-22
Add OpenSRP service

moshthepitt:
Fix: #208


src/services/opensrp/index.ts
```diff
@@ -0,0 +1,180 @@
...
+    const url = getURL(this.generalURL, params);
+    const payload = {
+      ...getPayload(method),
+      'Cache-Control': 'no-cache',
+      Pragma: 'no-cache',
+      body: JSON.stringify(data),
+    };
+    const response = await fetch(url, payload);
+
+    if (!response.ok || response.status !== 200) {
```
githengi: Also check for `201`

Issue #198, opened 2019-07-16
Updates to Focus Investigation "profile" [discussion]


rowo:
![Focus Investigation](https://user-images.githubusercontent.com/1584163/61327513-113cec00-a7e7-11e9-89d7-aad40f9e3d76.png)

![Focus Investigation Complete](https://user-images.githubusercontent.com/1584163/61327524-15690980-a7e7-11e9-8846-bb44a77dbc53.png)

- Add a "mark as complete" button
- Update the progress to match activity target issue https://github.com/onaio/reveal-frontend/issues/169
- Add a flag for active vs. complete and a "reactive investigation" vs. "routine investigation" title.
- The breadcrumbs should be links. 
- ~~Remove pulldown to other investigations (this is more of a standalone investigation rather than a way to browse investigations (not sure if this is/was implemented yet)~~ Edit: nevermind. There is a pulldown again.

moshthepitt:
@rowo I remember there was an involving discussion on the need to change how the indicators are displayed.  I think this was the chief complaint from the Thailand client.

On the mockups I see we have something like:

> Progress 18 of 20 structures visited (90%) 

In this case, 19 is the number of completed tasks, right?

But what does 20 represent?  it could be one of two things:

- The number of total tasks 
- The value of the goal's target from the Plan definition

These two numbers are usually the same, but we've seen cases where they are different in both directions (i.e. both when the total number of tasks is lower and greater than the goal target value).

Additionally, when the goal target is of unit "percent", the goal target value will be a target percentage so in this case it is clear that the denominator will be the total number of tasks.

Which means the part that is unclear to is what numbers to use for goals whose unit is not "percent".

@mberg @craigappl @cKellyDesign @githengi  have we come to an agreement on this?

moshthepitt:
Related: https://github.com/onaio/reveal-frontend/issues/169

moshthepitt:
@rowo regarding the pull down, and sorry if this has been addressed elsewhere, what is shown there?  Is it other focus areas, or other investigations in the current focus area?

rowo:
It is other investigations in the current focus area that would have an investigation view (the active and completed). The investigations should be listed out in chronological order, most recent first, with the current selected investigation and active investigations labeled. As an upgrade, the list could be divided into the same sections as the focus area page i.e. Current reactive, Current routine, Completed reactive, Completed routine, though I'm not actually sure if that's more useful or jsut by date alone. @mberg ?

[...]

PR #222, opened 2019-07-23
Detail view enhancement

kahama94:
fix: #198


src/containers/pages/FocusInvestigation/map/active/index.tsx
```diff
@@ -237,24 +244,76 @@ class SingleActiveFIMap extends React.Component<
...
-      ({
+      return {
         label: namePath,
         url: '',
-      })
-    );
+      };
+    });
     breadCrumbProps.pages = [homePage, basePage, ...pages, secondLastPage];
+    const statusBadge =
```
moshthepitt: Can we have this render the value of `plan.plan_status` please?  I do not see a need to use hard coded strings here.

src/containers/pages/FocusInvestigation/map/active/index.tsx
```diff
@@ -237,24 +244,76 @@ class SingleActiveFIMap extends React.Component<
...
+      ) : null;
+
+    const cols: string[] = Object.keys(plan);
+
+    const alias: FlexObject = {
+      plan_effective_period_end: END_DATE,
+      plan_effective_period_start: START_DATE,
+    };
 
+    const container: FlexObject[] = [];
```
moshthepitt: Can we either rename this to a more descriptive name, or add a comment on what is for?

src/containers/pages/FocusInvestigation/map/active/index.tsx
```diff
@@ -237,24 +244,76 @@ class SingleActiveFIMap extends React.Component<
...
+    const markCompleteButton =
+      plan && plan.plan_status === PlanStatus.ACTIVE ? (
+        <Button color="primary" size="lg" block={true}>
+          Mark as Complete
+        </Button>
+      ) : null;
+
+    const cols: string[] = Object.keys(plan);
+
+    const alias: FlexObject = {
```
moshthepitt: Can we either rename this to a more descriptive name, or add a comment on what is for?

moshthepitt: Why does this need to be of type `FlexObject`?  It seems that we already know what kind of object should go here.  Let us only use FlexObject when we absolutely cannot avoid it.

src/containers/pages/FocusInvestigation/map/active/index.tsx
moshthepitt: Why does this need to be of type `FlexObject`?  It seems that we already know what kind of object should go here.  Let us only use FlexObject when we absolutely cannot avoid it.

src/containers/pages/FocusInvestigation/map/active/index.tsx
```diff
@@ -237,24 +244,76 @@ class SingleActiveFIMap extends React.Component<
...
+        </Badge>
+      );
+    const markCompleteButton =
+      plan && plan.plan_status === PlanStatus.ACTIVE ? (
+        <Button color="primary" size="lg" block={true}>
+          Mark as Complete
+        </Button>
+      ) : null;
+
+    const cols: string[] = Object.keys(plan);
```
moshthepitt: Can we either rename this to a more descriptive name, or add a comment on what is for?

src/containers/pages/FocusInvestigation/map/active/index.tsx
```diff
@@ -237,24 +244,76 @@ class SingleActiveFIMap extends React.Component<
...
-        </h2>
+        <div>
+          <Row>
+            <Col xs="8">
+              <h2 className="page-title mt-4 mb-4">
+                {plan && plan.plan_title} {INVESTIGATION} {statusBadge}
+              </h2>
+            </Col>
+            <Col xs="4">
+              <Select placeholder="Other area Investigations" />
```
moshthepitt: A few things on this:
- This select will be empty.  [Here is what it needs to contain](https://github.com/onaio/reveal-frontend/issues/198#issuecomment-512242921)
- I would move this to a different component and pass the select options to it as props
- What did you consider in using **Select** vs **AsyncSelect**?

src/components/SelectPlan/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+class SelectPlan extends React.Component<SelectPlanProps, {}> {
+  constructor(props: SelectPlanProps) {
+    super(props);
+  }
+  public render() {
+    const { plansArray } = this.props;
+    /** Sort plans by plan_date and build value label key value pairs to populate the select */
+    let options;
+    options = plansArray.sort((a, b) => {
+      if (a.plan_date > b.plan_date) {
```
moshthepitt: Can this sorting be made simpler?

src/containers/pages/FocusInvestigation/map/active/index.tsx
```diff
@@ -182,7 +178,7 @@ class SingleActiveFIMap extends React.Component<
           fetchStructuresActionCreator(structuresResults);
         }
       );
-      await supersetFetch(SUPERSET_PLANS_SLICE, supersetParams).then((result2: Plan[]) => {
+      await supersetFetch(SUPERSET_PLANS_SLICE).then((result2: Plan[]) => {
```
moshthepitt: Why did we stop filtering?

PR #226, opened 2019-07-25
New IRS Planning Page

cKellyDesign:


src/store/ducks/plans.ts
```diff
@@ -13,11 +13,13 @@ export enum InterventionType {
   IRS = 'IRS',
 }
 
+/** interface for plan Objects */
 /** Enum representing the possible intervention types */
 export enum PlanStatus {
   ACTIVE = 'active',
   COMPLETE = 'complete',
   DRAFT = 'draft',
+  NEW = 'new',
```
moshthepitt: I think `new` is not a supported plan status value from the FHIR docs on plan definitions.  (`complete` is also not supported by FHIR but we chose to add it in).  I was talking to @rowo in another issue on this repo and we think we can use the plan date field to kind of loosely determine whether a plan is new or not.

@githengi could you please confirm if I got this right?

src/store/ducks/plans.ts
```diff
@@ -248,6 +255,7 @@ export type ImmutablePlanState = PlanState & SeamlessImmutable.ImmutableObject<P
 const initialState: ImmutablePlanState = SeamlessImmutable({
   planRecordsById: {},
   plansById: {},
+  plansByPlanId: {},
```
moshthepitt: It seems that we aren't using this anywhere?

src/configs/env.ts
```diff
@@ -46,6 +46,10 @@ export const SUPERSET_JURISDICTIONS_SLICE =
   process.env.REACT_APP_SUPERSET_JURISDICTIONS_SLICE || '0';
 export type SUPERSET_JURISDICTIONS_SLICE = typeof SUPERSET_JURISDICTIONS_SLICE;
 
+export const SUPERSET_JURISDICTIONS_DATA_SLICE =
```
moshthepitt: Should we add this to the .env.sample file?

src/containers/pages/InterventionPlan/IRS/plan/index.tsx
```diff
@@ -1,79 +1,1419 @@
...
   supersetService: supersetFetch,
 };
 
+interface TableCrumb {
+  label: string;
+  id: string | null;
+  active: boolean;
+}
+
+interface IrsPlanState {
```
moshthepitt: Should we add docstrings here?

src/containers/pages/InterventionPlan/IRS/plan/index.tsx
```diff
@@ -1,79 +1,1419 @@
...
 
-import { SUPERSET_PLANS_TABLE_SLICE } from '../../../../../configs/env';
-import { HOME, HOME_URL, INTERVENTION_IRS_URL } from '../../../../../constants';
-import { RouteParams } from '../../../../../helpers/utils';
+import { GREY } from '../../.../../../../../colors';
+import {
+  SUPERSET_JURISDICTIONS_DATA_SLICE,
+  SUPERSET_JURISDICTIONS_SLICE,
+  SUPERSET_PLAN_STRUCTURE_PIVOT_SLICE,
+  SUPERSET_PLANS_TABLE_SLICE,
```
moshthepitt: I think `GREY`, `extractPlanRecordFromPlanPayload`, `PlanRecordResponse`, and `SUPERSET_PLANS_TABLE_SLICE` are imported but not used.

src/containers/pages/InterventionPlan/IRS/plan/index.tsx
```diff
@@ -1,79 +1,1419 @@
...
+            .map((j: any) =>
+              !loadedJurisdictionIds.includes(j.jurisdiction_id) ? j.jurisdiction_id : null
+            )
+            .filter((j: string | null) => j && strict.length);
+
+          if (!planJurisdictionIdsToGet.length) {
+            return new Promise(resolve => resolve());
+          }
+
+          // build superset adhoc filter expression
```
moshthepitt: This comment seems to be misleading

src/containers/pages/InterventionPlan/IRS/plan/index.tsx
```diff
@@ -1,79 +1,1419 @@
...
+          let sqlFilterExpression = '';
+          for (let i = 0; i < planJurisdictionIdsToGet.length; i += 1) {
+            const jurId = planJurisdictionIdsToGet[i];
+            if (i) {
+              sqlFilterExpression += ' OR ';
```
moshthepitt: I think in many cases using `IN` vs `OR` is usually faster/more efficient.  Do you think we should do that here?  Source: https://stackoverflow.com/a/3074731

src/containers/pages/InterventionPlan/IRS/plan/index.tsx
```diff
@@ -1,79 +1,1419 @@
...
+        );
+
+        const doInclude = otherJurisdictionIdsToGet.length < otherJurisdictionIdsNotToGet.length;
+        idsToUse = doInclude ? otherJurisdictionIdsToGet : otherJurisdictionIdsNotToGet;
+
+        // build query params
+        for (let i = 0; i < idsToUse.length; i += 1) {
+          const jurId = idsToUse[i];
+          if (i) {
+            sqlFilterExpression += doInclude ? ' OR ' : ' AND ';
```
moshthepitt: I think in many cases using `IN` vs `OR` is usually faster/more efficient.  Do you think we should do that here?  Source: https://stackoverflow.com/a/3074731

src/containers/pages/InterventionPlan/IRS/plan/index.tsx
```diff
@@ -1,79 +1,1419 @@
...
+            };
+            return jurisdiction;
+          });
+          return fetchJurisdictionsActionCreator(jurisdictions);
+        }
+      );
+    }
+  }
+
+  public componentWillReceiveProps(nextProps: IrsPlanProps) {
```
moshthepitt: I feel uncomfortable using `componentWillReceiveProps` given that it is deprecated 🙈 for a good reason IMO.  I think if we have to use it, we should have a corresponding issue to investigate the possibility of replacing it with something else.

src/containers/pages/InterventionPlan/IRS/plan/index.tsx
```diff
@@ -1,79 +1,1419 @@
...
+      (isDraftPlan && isLoadingJurisdictions)
+    ) {
       return <Loading />;
     }
+
+    const pageLabel =
+      (isFinalizedPlan && planById && planById.plan_title) ||
+      (isDraftPlan && planById && `${planById.plan_title} (draft)`) ||
+      (newPlan && newPlan.plan_title) ||
+      'New Plan';
```
moshthepitt: I think we have used this string a number of times in this module.  In the spirit of zero tolerance to magic strings, should we move it to the constants module?

src/containers/pages/InterventionPlan/IRS/plan/index.tsx
```diff
@@ -1,79 +1,1419 @@
...
+                  <Input
+                    id="select-plan-country"
+                    defaultValue={planCountry}
+                    name="select-plan-country"
+                    onChange={onSelectCountryChange}
+                    type="select"
+                  >
+                    <option>...</option>
+                    {/* <option value="BW">Botswana</option> */}
+                    <option value="NA">Namibia</option>
```
moshthepitt: Should we load these options from the settings module, or somewhere that allows for easy configuration?

src/containers/pages/InterventionPlan/IRS/plan/index.tsx
```diff
@@ -1,79 +1,1419 @@
...
+                  Select Jurisdictions
+                </Button>
+              </Form>
+            </Col>
+          </Row>
+        </div>
+      );
+    }
+
+    const onEditNameButtonClick = (e: any) => {
```
moshthepitt: I think we should try very hard not to use `any` here for these events.  Do you agree?

moshthepitt: I also think that we should consider adding docstrings for these functions here and elsewhere in this module

src/containers/pages/InterventionPlan/IRS/plan/index.tsx
```diff
@@ -1,79 +1,1419 @@
...
+        const { geographic_level } = j;
+        if (
+          typeof geographic_level !== 'undefined' &&
+          geoGraphicLevels.indexOf(geographic_level) === -1
+        ) {
+          geoGraphicLevels.push(geographic_level);
+        }
+      }
+      geoGraphicLevels.sort().reverse();
+      const colorMap = ['red', 'orange', 'yellow', 'green', 'blue', 'purple'];
```
moshthepitt: Should we get this from settings?

src/containers/pages/InterventionPlan/IRS/plan/index.tsx
```diff
@@ -1,79 +1,1419 @@
...
+          visible: true,
+        };
+        layers.push(layer);
+      }
+      return layers;
+    }
+
+    const JURISDICTION_FILL_LAYERS = getJurisdictionFillLayers(filteredJurisdictions);
+
+    const drillUpHandler: Handlers = {
```
moshthepitt: Should we move this to the module that has the other Gisida handlers?

moshthepitt:
Screenshot from 2019-07-29 13-26-13

@cKellyDesign I am getting this error:

```
Error {message: "Cannot read property 'geometry' of null"}
```

When I click on the map

src/store/ducks/plans.ts
githengi: I think `new` is not required. New plans should be of status `draft` or what kind of action is needed to move a plan from `new` to `draft`?

src/configs/settings.ts
```diff
@@ -594,6 +594,106 @@ export const circleLayerConfig = {
...
+    layer: string;
+    parentIdField: string;
+    url: string;
+  }>;
+}
+
+export const ZambiaAdmin0: JurisdictionsByCountry = {
+  ADMN0_EN: 'Zambia',
+  ADMN0_PCODE: 'ZM',
+  jurisdictionIds: ['2939', '2940', '2942', '2942', '2953', '2954'],
```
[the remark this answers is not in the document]
cKellyDesign: Agreed, if we can identify which country those parentless Jurisdictions belong to than that would be ideal, rather than hard-coding like this.

src/configs/settings.ts
```diff
@@ -594,6 +594,106 @@ export const circleLayerConfig = {
   visible: false,
 };
 
+export interface ADMN0 {
```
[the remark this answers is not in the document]
cKellyDesign: Oh definitely, I will add the doc-strings and do some clean-up here.

src/helpers/utils.tsx
```diff
@@ -409,3 +412,14 @@ export function roundToPrecision(n: number, precision: number = 0): number {
   const factor = Math.pow(10, precision);
   return Math.round(n * factor) / factor;
 }
+
+export function stopPropagation(e: any) {
```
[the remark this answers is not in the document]
cKellyDesign: I'm forgetting where I saw these events being returned as `any`, may have to differentiate between normal React events and Mapbox events.

src/store/ducks/jurisdictions.ts
```diff
@@ -14,26 +15,48 @@ export interface JurisdictionGeoJSON extends GeoJSON {
...
 export interface Jurisdiction {
-  geojson: JurisdictionGeoJSON;
+  geographic_level?: number;
+  geojson?: JurisdictionGeoJSON;
   jurisdiction_id: string;
+  name?: string | null;
+  parent_id?: string | null;
+}
+
+export interface AllJurisdictionIds {
```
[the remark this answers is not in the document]
cKellyDesign: Yes, definitely! Also maybe some renaming 😊

src/containers/pages/InterventionPlan/IRS/plan/index.tsx
cKellyDesign: Sure thing! ...what's a magic string?

src/containers/pages/InterventionPlan/IRS/plan/index.tsx
cKellyDesign: Yeah, I'll go through and add doc strings to all the things today.

src/containers/pages/InterventionPlan/IRS/plan/index.tsx
cKellyDesign: Mmm, yes absolutely.

src/containers/pages/InterventionPlan/IRS/plan/index.tsx
moshthepitt: Here you go https://deviq.com/magic-strings/

src/containers/pages/InterventionPlan/IRS/plan/index.tsx
cKellyDesign: Not sure since these handlers call private functions of the IRS plan page  which are very specific to the interaction of drilling up and down through the Jurisdiction hierarchies. If we extract the map itself from this page then maybe but then we would probably have to migrate some of this internal component state into redux stores for the separate modules to work together like this.

cKellyDesign:
@moshthepitt the finalized plan view needs a bit more work to use the same drilldown table as the new and draft views, I'll standardize that today as well.

cKellyDesign:
Not sure why the number of Jurisdictions would be increasing like that, the page's internal state should be clearing when moving between plans.

cKellyDesign:
> I think that we should consider (probably via an issue) moving the map stuff out of this component and into another one

@moshthepitt - do you mean decoupling the map from the act of Planning? I think this makes sense and would help make the file a bit more manageable.

src/store/ducks/plans.ts
cKellyDesign: `new` was meant to be more of an internal status representing a plan that has not yet been saved as a `draft`, the only action required is for the user to click the `save as draft` button and the draft plan model is constructed before being saved to OpenSRP with `status: draft`.

src/containers/pages/InterventionPlan/IRS/plan/index.tsx
cKellyDesign: Ooh yeah, I'll make sure to remove any un-used imports

src/containers/pages/InterventionPlan/IRS/plan/index.tsx
cKellyDesign: Oooh, sure thing

src/containers/pages/InterventionPlan/IRS/plan/index.tsx
cKellyDesign: Oh interesting, not sure how to get around this off the top of my head, but if this is the case then we should probably have the issue be for updating the GisidaWrapper component as well.

src/store/ducks/plans.ts
moshthepitt: aaaah, cool.  It makes it a little complicated to deal with this enum coz you ave to keep in mind one of them should not be used/expected with the OpenSRP server and/or Superset

src/containers/pages/InterventionPlan/IRS/plan/index.tsx
moshthepitt: Yeah I'm good with this moving to a technical debt issue, for tackling later.

src/containers/pages/InterventionPlan/IRS/plan/index.tsx
moshthepitt: Maybe that is worthy of another issue?

moshthepitt:
> @moshthepitt - do you mean decoupling the map from the act of Planning? I think this makes sense and would help make the file a bit more manageable.

Yes :)

src/helpers/utils.tsx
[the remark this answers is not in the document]
cKellyDesign: 1c72218 updates the Mapbox specific events and 728f30c updates click event handlers, but I started running into issues with `onChange` event handlers, even when trying the `ChangeEvent`, `FormEvent` and `SyntheticEvent` typings from `react`. Not sure why changing the typings was breaking the linkage between the elements and handlers, but #236 should negate the use of `onChange` event handlers state for the most part.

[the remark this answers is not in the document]
cKellyDesign: Oh but on this line it should probably have an array of acceptable types.

src/store/ducks/plans.ts
cKellyDesign: > you ave to keep in mind one of them should not be used/expected

Good point, I'll pull it out of the type, add the string "new" to a constant, and set the component state to allow either the enum or the constant.

cKellyDesign: @moshthepitt - I tried quickly decoupling `new` out of the `PlanStatus` but started running into some gnarly TS errors about extending / inheritance. I think this should be done but maybe for now we can just wrap the service calls with conditionals to protect making calls to OpenSRP with `new`.

src/containers/pages/InterventionPlan/IRS/plan/index.tsx
cKellyDesign: @moshthepitt If i could get an example of the `IN` syntax being used in SupersetConnector's `sqlExpression` option that would be super helpful.

src/containers/pages/InterventionPlan/IRS/plan/index.tsx
cKellyDesign: Cool deal, issue created at #237

src/containers/pages/InterventionPlan/IRS/plan/index.tsx
cKellyDesign: Cool deal, replace this one but seems like quite a few of my strings could be considered magical, created a new issue at #238 to audit and replace in the near future.

src/containers/pages/InterventionPlan/IRS/plan/index.tsx
cKellyDesign: Issue created hapa: #243

src/containers/pages/InterventionPlan/IRS/plan/index.tsx
cKellyDesign: Done at `21bbddb`

cKellyDesign:
Issue created at #244 for https://github.com/onaio/reveal-frontend/pull/226#pullrequestreview-267696551

PR #227, opened 2019-07-25
Create FI Plans

### LINKED ISSUE #199: Create/Add and Edit FI page
## Add new FI
Add Focus Investigation
- For user-created investigations
- Clicking on the checkbox will make visible a target as described in https://github.com/onaio/reveal-frontend/issues/188#issue-467064059 and https://github.com/onaio/reveal-frontend/issues/188#issuecomment-511298511
- The start date is for display and FI naming only, equivalent to the "case notification" date in reactive investigations. 
- Saving the FI as active will mean this view isn't editable anymore (and the FI isn't editable). Therefore, we probably need a confirmation step if an investigation is saved as Active. What do you think @mberg ?
- Save button updates to the selected status i.e. "Save as draft" or "Save and activate"
- @mberg are all fields required?

## Edit FI (either a draft of a user-created, or a new draft of an incoming case-triggered)
Edit Focus Investigation
- Start date will change to "Case notification date" for case-triggered instances
- Generally the same as above "add FI" form except with a new uneditable name field. I don't know if the name changes if the start date or case notification date changes during edit. @mberg what do you think?

moshthepitt:
Implements a way to add Focus Investigation plans, and fixes #199.  Based on the [plan generation web form specification](https://docs.google.com/spreadsheets/d/1kGPgyF0eJdag19FFL1Oog2gdphdPtKobY4EQ0gMgtgw/edit#gid=1805975071).

A high level goal is to make it re-usable for other types on interventions, although the first version will be focused on FI.

- [x] Add basic plan form
- [x] Add date pickers
- [x] Plan UUID generation
- [x] Plan name & title generation
- [x] Deal with FI section
- [x] Deal with Subsection FI Case Triggered
- [x] Subsection Case Confirmation Action
- [x] Subsection Family Registration Action
- [x] Subsection Blood Screening Action
- [x] Subsection Bednet Distribution Action
- [x] Subsection Larval Dipping Action
- [x] Subsection Mosquito Collection Action
- [x] Subsection IRS Action
- [x] Ensure validation is working as expected
- [x] Deal with jurisdictions
- [x] Deal with version
- [x] Deal with action identifier
- [ ] Add tests
- [x] Add method to construct OpenSRP geojson


moshthepitt:
Here is what the form looks like so far @rowo @craigappl - I know this breaks the mockup that we had, but I think that was not avoidable and  I hope that we can get fresher mockups (though not before Wednesday's deadline)?:

long plan form

mberg:
@moshthepitt - the mockups I think can still be replicated.  All this info per activity can be hard coded via a template. The only thing to modify here is the quantity 

- action = same name as the activity. Hidden property 
- start and end date = same as what is entered in the start of the form. Hidden property 
- description = hardcoded to a template. Doesn’t need to be made visible.  Make it a hidden property. 
- reason = investigation or case triggered - same as above
- priority - can he hardcoded to medium priority for now.  Hidden priority 
- quantity - only thing that needs to be displayed

src/configs/settings.ts
```diff
@@ -92,6 +92,451 @@ export const locationHierarchy: LocationItem[] = [
...
+};
+
+/** UseContext - interface for PlanPayload.useContext[] items */
+export interface UseContext {
+  code: UseContextCodesType;
+  valueCodableConcept: string;
+}
+
+/** interface that describes plan definition objects from OpenSRP */
+export interface PlanDefinition {
```
cKellyDesign: 👍 Looks like we defined similar interfaces in different places, I may cherry pick this for the IRS branch or just update it later.

src/containers/forms/PlanForm/helpers.ts
```diff
@@ -0,0 +1,339 @@
...
+        .required(REQUIRED),
+      timingPeriodEnd: Yup.date().required(REQUIRED),
+      timingPeriodStart: Yup.date().required(REQUIRED),
+    })
+  ),
+  caseNum: Yup.string(),
+  date: Yup.string().required(`${DATE} ${IS} ${REQUIRED}`),
+  end: Yup.date().required(REQUIRED),
+  fiReason: Yup.string().oneOf(FIReasons.map(e => e)),
+  fiStatus: Yup.string().oneOf(fiStatusCodes),
```
cKellyDesign: Little nit-picky, feels like `fiReason` and `fiStatus` should be mapped the same way (here or above) for uniformity.

src/containers/forms/PlanForm/helpers.ts
```diff
@@ -0,0 +1,339 @@
...
+  status: PlanStatus;
+  title: string;
+}
+
+/**
+ * Convert a plan activity object to an object that can be used in the PlanForm
+ * activities section
+ * @param activityObj - the plan activity object
+ */
+export function extractActivityForForm(activityObj: PlanActivity): PlanActivityFormFields {
```
cKellyDesign: Do you think there is a better place for these types of extractor functions? In this case here probably makes sense b/c the extractor is so specific to the needs of the Form. For more general extractors should they live in the ducks or should there be a similar `helpers` for ducks? (no need to decide before the release, just something to keep in mind!)

src/containers/forms/PlanForm/helpers.ts
```diff
@@ -0,0 +1,339 @@
...
+  const initialGoalDue = activityObj.goal.target[0].due;
```
cKellyDesign: Will there always be at least 1 target in array? Or should this be wrapped in a conditional to protect against `TypeError: Cannot read property 'due' of undefined`?

src/containers/forms/PlanForm/helpers.ts
```diff
@@ -0,0 +1,339 @@
...
+
+  return {
+    action: actions,
+    goal: goals,
+  };
+}
+
+/**
+ * Get the plan name and title
+ * @param {any} event - the event object
```
cKellyDesign: should be `@param {FormEvent} event - the event object`

src/containers/forms/PlanForm/helpers.ts
```diff
@@ -0,0 +1,339 @@
...
+    errors && Object.entries(errors[index] || {}).filter(([key, val]) => key === field).length > 0
+  );
+}
+
+/**
+ * Generate an OpenSRP plan definition object from the PlanForm
+ * @param formValue - the value gotten from the PlanForm
+ * @returns {PlanDefinition} - the plan definition object
+ */
+export function generatePlanDefinition(formValue: PlanFormFields): PlanDefinition {
```
cKellyDesign: Really nice

src/containers/forms/PlanForm/helpers.ts
moshthepitt: Hmmm, could you explain further?

src/configs/settings.ts
moshthepitt: Yeah, I dont know how best we'll manage the code before we fully integrate both branches together.  Ideas?

src/containers/forms/PlanForm/helpers.ts
moshthepitt: I dont know where to best put them.  For now since they are only used for PlanForm we have them here, but I think you are right about helpers for ducks.  We should do that.

src/containers/forms/PlanForm/helpers.ts
moshthepitt: I'll add a check here

src/containers/forms/PlanForm/helpers.ts
moshthepitt: Yes it should. Good catch

src/configs/settings.ts
cKellyDesign: I was thinking of making sure both worked together and then [lightly] refactoring IRS code as needed to use shared codes you put together for FI.

src/containers/forms/PlanForm/helpers.ts
cKellyDesign: Just that `FIReasons` is mapping inline while `fiStatusCodes` was mapped previously.

src/containers/forms/PlanForm/helpers.ts
cKellyDesign: Cool, created an issue for this https://github.com/onaio/reveal-frontend/issues/252

src/containers/forms/PlanForm/helpers.ts
moshthepitt: Created an issue: https://github.com/onaio/reveal-frontend/issues/253

src/helpers/utils.tsx
```diff
@@ -392,3 +393,14 @@ export function roundToPrecision(n: number, precision: number = 0): number {
...
 }
+
+/**
+ * Generate a namespaced uuid using uuidv5
+ * @param {string} seedString - the string to be used to generate the uuid. It should
+ * be unique per namespace
+ * @param {string} namespace - the namespace
+ * @returns {string} - the uuid
+ */
+export function generateNameSpacedUUID(seedString: string, namespace: string) {
```
cKellyDesign: Noice!

src/services/opensrp/index.ts
```diff
@@ -31,12 +31,22 @@ type paramsType = URLParams | null;
...
   return Object.entries(obj)
     .map(([key, val]) => `${key}=${val}`)
     .join('&');
 }
 
+/** converts filter params object to string
+ * @param {URLParams} obj - the object representing filter params
+ * @returns {string} filter params as a string
+ */
+export function getFilterParams(obj: URLParams | {}): string {
```
cKellyDesign: Oh sweet, yes!

PR #228, opened 2019-07-25
singlefi view

### LINKED ISSUE #196: Updates to Focus Area "profile" [discussion]
Focus Area

1. Added tables to represent investigations that are: current reactive, current routine, previous reactive, previous routine.
2. Used columns from current investigations (#190), but removed the location, status... (this is the rule for any columns with shared data — if FI status (i.e. A1, B1 etc) is editable, we might want to keep that showing)
3. The Add Focus Investigation button is added to the current routine table.
4. Edit: Added a "delete" and "mark as complete" to the current investigations. This is the only place to delete a draft investigation because the only other independent view of an investigation is an edit state.
5. Edit: Added some columns to the Complete Investigations

kahama94:
fix #196


src/configs/settings.ts
```diff
@@ -241,3 +257,263 @@ export const symbolLayerConfig = {
   },
   type: 'symbol',
 };
+/** Default table props config */
+export const defaultTableProps = {
```
moshthepitt: We should not do this in the `settings` module.  Please do this in a helpers or utils module somewhere else.  Reasons why:

1. The settings module should not have any dependencies that are in the codebase
2. We dont want to run into cyclic dependency issues

I'm fine with putting the status column headers in the settings module, but not anything that requires an import from the Reveal web code base.

src/helpers/utils.tsx
```diff
@@ -409,3 +420,122 @@ export function roundToPrecision(n: number, precision: number = 0): number {
...
+          },
+        ],
+      },
+    ];
+  } else {
+    return [];
+  }
+}
+
+export function buildTableHeader(plansArray: FlexObject[]) {
```
moshthepitt: Lets not use FlexObject here please.

moshthepitt: Can we move this to the components directory?

src/helpers/utils.tsx
```diff
@@ -409,3 +420,122 @@ export function roundToPrecision(n: number, precision: number = 0): number {
   const factor = Math.pow(10, precision);
   return Math.round(n * factor) / factor;
 }
+
+/** Return columns with jsx implementations */
```
moshthepitt: Perhaps nitpicky, but what does "with jsx implementations" mean?  Is it possible to have a better docstring and function name here?

src/helpers/utils.tsx
```diff
@@ -409,3 +420,122 @@ export function roundToPrecision(n: number, precision: number = 0): number {
...
+              Add Focus Investigation
+            </Button>
+          </Col>
+        </Row>
+      </div>
+    );
+  }
+}
+
+export function buildTableWithoutPlanData(
```
moshthepitt: Both `buildTableHeader` and `buildTableWithoutPlanData` seem to me like they are regular react components.  Therefore I would prefer that for both of them we:

- Move them to the components directory
- Add better prop definitions:
    - We certainly do not need to use FlexObject for either `tableProps` or `plansArray`.  Let us not abuse FlexObject, please.
    - can we add comments on what the props represent?
    - do the props of either of these components have default values? 
- Write better docstring
- Add individual tests for each, with snapshots

Finally, and this is up for debate, I think that it is more common to define function components using arrow functions as opposed to what we have here.  I recommend that you consider doing this.  [Please see this article](https://www.robinwieruch.de/react-function-component/).

src/configs/settings.ts
```diff
@@ -1,6 +1,20 @@
 /** This is the main configuration module */
 import { Providers } from '@onaio/gatekeeper';
 import { Expression } from 'mapbox-gl';
+import {
```
moshthepitt: Please dont do this import.  I'm fine with having the string literal here.  We dont want settings to depend on anything in the codebase

src/containers/pages/InterventionPlan/tests/__snapshots__/IRS.test.tsx.snap
```diff
@@ -118,7 +118,7 @@ exports[`containers/pages/IRS renders plans list correctly 1`] = `
           Object {
             "id": "draftPlan-id-2",
             "jurisdiction_depth": 41,
-            "jurisdiction_id": "3378",
+            "jurisdiction_id": "450fc15b-5bd2-468a-927a-49cb10d3bbb",
```
moshthepitt: Why is this changing?

src/helpers/utils.tsx
```diff
@@ -392,3 +404,103 @@ export function roundToPrecision(n: number, precision: number = 0): number {
   const factor = Math.pow(10, precision);
   return Math.round(n * factor) / factor;
 }
+
+/** Returns Table columns Which require external dependencies (Cell, Link, CellInfo)
+ * this columns include focusarea, name and action
+ * @param {colType} accessor column
+ */
+
```
moshthepitt: Please remove this empty line

src/store/ducks/tests/fixtures.ts
```diff
@@ -43,7 +43,7 @@ export const plan2 = {
 export const draftPlan = {
   id: 'draftPlan-id-2',
   jurisdiction_depth: 41,
-  jurisdiction_id: '3378',
+  jurisdiction_id: '450fc15b-5bd2-468a-927a-49cb10d3bbb',
```
moshthepitt: Why is this changing?

PR #235, opened 2019-07-29
test superset connector in FI containers

### LINKED ISSUE #212: Improve tests of components that use SupersetConnector
The components that fetch data from Superset need tests that ascertain the Superset service was called with the correct/expected parameters.

- [ ] FI map page component
- [ ] FI single page component
- [ ] FI active list component

### LINKED ISSUE #240: Typo
i think `provice` should be `province` [here](https://github.com/onaio/reveal-frontend/blob/d3d76aa3a340df9b1a89d6234cf46baa4ad1a0fc/src/containers/pages/FocusInvestigation/single/index.tsx#L161)

p-netm:
Ascertain the Superset service in Focus Investigation components was called with the correct/expected parameters.

- [x] FI map page component
- [x] FI single page component
- [x] FI active list component

closes #212 closes #240


src/containers/pages/FocusInvestigation/map/active/tests/index.test.tsx
```diff
@@ -116,4 +125,45 @@ describe('containers/pages/FocusInvestigation/activeMap', () => {
...
+      </Provider>
+    );
+    // HACK: HACK: to force for asynchronous code in component to execute.
+    await Promise.resolve();
+    await Promise.resolve();
+    await Promise.resolve();
+    await Promise.resolve();
+    await Promise.resolve();
+    expect(supersetMock).toHaveBeenCalledTimes(5);
+    const callList = [[1, {}], [2, {}], [0, {}], [3, {}], [4, {}]];
```
moshthepitt: Why are these objects empty?  I think in the code we call each with some params, no? e.g. `supersetService(SUPERSET_JURISDICTIONS_SLICE, jurisdictionsParams)`

src/containers/pages/FocusInvestigation/active/tests/__snapshots__/index.test.tsx.snap
```diff
@@ -4945,18 +4890,7 @@ exports[`containers/pages/ActiveFocusInvestigation renders ActiveFocusInvestigat
...
-                  "clause": "WHERE",
-                  "comparator": "FI",
-                  "expressionType": "SIMPLE",
-                  "operator": "==",
-                  "subject": "plan_intervention_type",
-                },
-              ],
-              "row_limit": 2000,
-            },
+            Object {},
```
moshthepitt: I think that this change should not happen.  Have we stopped calling the Superset service with this filter?

src/containers/pages/FocusInvestigation/active/tests/__snapshots__/index.test.tsx.snap
```diff
@@ -4889,18 +4845,7 @@ exports[`containers/pages/ActiveFocusInvestigation renders ActiveFocusInvestigat
         "calls": Array [
           Array [
             0,
-            Object {
```
moshthepitt: I think that this change should not happen.  Have we stopped calling the Superset service with this filter?

src/containers/pages/FocusInvestigation/active/tests/__snapshots__/index.test.tsx.snap
```diff
@@ -4862,18 +4829,7 @@ exports[`containers/pages/ActiveFocusInvestigation renders ActiveFocusInvestigat
...
```
moshthepitt: I think that this change should not happen.  Have we stopped calling the Superset service with this filter?

src/containers/pages/FocusInvestigation/active/tests/__snapshots__/index.test.tsx.snap
```diff
@@ -162,18 +140,7 @@ exports[`containers/pages/ActiveFocusInvestigation renders ActiveFocusInvestigat
...
```
moshthepitt: I think that this change should not happen.  Have we stopped calling the Superset service with this filter?

src/containers/pages/FocusInvestigation/active/tests/__snapshots__/index.test.tsx.snap
```diff
@@ -106,18 +95,7 @@ exports[`containers/pages/ActiveFocusInvestigation renders ActiveFocusInvestigat
...
```
moshthepitt: I think that this change should not happen.  Have we stopped calling the Superset service with this filter?

src/containers/pages/FocusInvestigation/active/tests/__snapshots__/index.test.tsx.snap
```diff
@@ -79,18 +79,7 @@ exports[`containers/pages/ActiveFocusInvestigation renders ActiveFocusInvestigat
...
```
moshthepitt: I think that this change should not happen.  Have we stopped calling the Superset service with this filter?

src/containers/pages/FocusInvestigation/map/active/tests/index.test.tsx
p-netm: I mocked out the response of `superset.getFormData` to return an empty object. The thought process was that its functionality is already tested in the js-tools supersetConnector. So essentially I just needed to test `supersetFetch` gets called with the right args irregardless of what `superset.getFormData` response is. 

Calling the __supersetFetch mock__ with  an empty object means that:
 - `superset.getformData` was called, the test assumes the response of `getFormData` is already tested and therefore it mocks out the actual response with an empty object

I think what I could add is to check that a mock of `superset.getFormData` is called the right number of times and with the correct args i.e. e.g for a single call:

```typescript 
    [ 
        3000, [
               { comparator: plan.plan_id, operator: '==', subject: 'plan_id' },
           ]
     ]
``` 

but I still think that I would retain `superset.getFormData` mock implementation to return an empty dictionary, as it currently does.

@moshthepitt  How would you advise?

moshthepitt: > So essentially I just needed to test supersetFetch gets called with the right args irregardless of what superset.getFormData response is.

I agree with this line of thinking.

It seems to me, however, that our current tests do not achieve this goal.

This bit of code `supersetService(SUPERSET_JURISDICTIONS_SLICE, jurisdictionsParams)` results in supersetFetch being called with two params, the first one is the slice, the second one is the filter.

We are therefore not testing that we are indeed using the correct (or any) filters.  We need to ensure that the right value of the filter is captured in the test.  Do you agree?

I think that we need to find a way to mock `superset.getFormData` in the same test so that we have a more useful and accurate test.

src/containers/pages/FocusInvestigation/single/tests/index.test.tsx
```diff
@@ -145,4 +145,64 @@ describe('containers/pages/SingleFI', () => {
...
+          {
+            comparator: fixtures.plan1.jurisdiction_id,
+            operator: '==',
+            subject: 'jurisdiction_id',
+          },
+        ],
+      ],
+    ];
+    expect((superset.getFormData as any).mock.calls).toEqual(getFormDataCallList);
+    const callList = [[0, {}], [3, {}], [1, {}]];
```
moshthepitt: Is there no way to make sure we dont get empty objects here?  I think it makes the test very confusing

moshthepitt:
Nice work

Issue #277, opened 2019-08-01
Re-use jurisdiction geometries instead of Re-downloading them


cKellyDesign:
This ticket is primarily for IRS planning page but should be also be considered any time a page is pulling geojson from the server it should first check local state to see if they exist. 

...UNLESS the geojson properties are different (like for goals) then we will still need to re-download the jurisdiction. Ideally we should have one set of geometry for each GIS entity which gets contextually specified at attached if/when needed.

moshthepitt:
What are you ideas on how to achieve this?

cKellyDesign:
What I'm thinking is basically:
1. determine what Jurisdictions are involved
2. check the store to see which Jurisdictions have geojson
3. make API requests only for those that don't have them
4. then `fetchJurisdictions(Jurisdiction[])` (dispatch / save to store)
5. the page can then move on with the process of rendering them all from geojson in the store

The main barrier that I've run into so far is dealing with the asynchronous-ity of dispatching to the store when a component needs the results of this process to move onto the next step of things (like building GisidaWrapper props). Was thinking of adding `isLoading` or `isFetching`  boolean to the Jurisdiction store (which could be quickly dispatched to true at the beginning of some of these chained async processes).

cKellyDesign:
Closing this as it is currently happening, we only download geojson of administrative jurisdictions if no tilesets are provided.

Issue #252, opened 2019-07-31
Move Planning getters and extractors into Ducks


cKellyDesign:
Certain getter / extractor functions may be better suited as methods included in the Ducks modules of their entity's redux store. This is for both FI and IRS plans.

cKellyDesign:
- [ ] Audit for getters in IRS Plans list page and IRS Plan Jurisdiction Selection page
- [ ] Migrate Jurisdiction getters / helpers out of IRS Plan Jurisdiction Selection

cKellyDesign:
This should be a part of #416

cKellyDesign:
Let's not do this, if anything the amount of helpers in this module should be reduced as much as possible.

PR #280, opened 2019-08-01
Streamline IRS jurisdiction reference logic

cKellyDesign:
@moshthepitt @kahama94 @p-netm - this is a small refactor of how the IRS plan page references Jurisdictions. By not iterating over and filtering 1000s of jurisdictions loaded with geojson actually seems to give quite a boost in performance! 

This PR should follow #273; changes affect the utility and not the business logic of the IRS plan page and is an incremental step towards #252 and #277


src/containers/pages/InterventionPlan/IRS/plan/index.tsx
```diff
@@ -110,7 +112,7 @@ export interface IrsPlanProps {
   isDraftPlan?: boolean;
   isFinalizedPlan?: boolean;
   isNewPlan?: boolean;
-  jurisdictionsArray: Jurisdiction[];
+  jurisdictionsById: { [key: string]: Jurisdiction };
```
moshthepitt: Do we want to necessarily replace `jurisdictionsArray `like this, or just add `jurisdictionsById` alongside it?

cKellyDesign: Yep want to completely remove it, there should be no need tor it at all at this point!

PR #282, opened 2019-08-02
Refactor link to new plans

### LINKED ISSUE #266: Refactor link to new plans
To make the code more DRY, we should move this component to one place where we can import it from when re-using:

```ts
            <Link
                to={NEW_PLAN_URL}
                className="focus-investigation btn btn-primary float-right mt-5"
              >
                Add Focus Investigation
            </Link>
```

FrankApiyo:
Fixes #266


FrankApiyo:
I see why functional components are better in this case. The constructor function is useless here

PR #294, opened 2019-08-05
make caseNum required if intervention type is FI

### LINKED ISSUE #250: FI Plan: Yup validations don't work on conditional fields
Yup validations dont yet work on conditional e.g. validate that caseNum is required when type is FI and not required when IRS

p-netm:
add Yup validation to check that `plans.caseNum` is required if and when `interventionType` is `FI`

closes #250


.circleci/config.yml
```diff
@@ -58,4 +58,4 @@ jobs:
       # run linter
       - run: yarn lint-ts
       # run tests!
-      - run: yarn test --verbose --coverage=true --forceExit --detectOpenHandles
\ No newline at end of file
+      - run: yarn test --runInBand --verbose --coverage=true --forceExit --detectOpenHandles
```
moshthepitt: What is the rationale for doing this?

src/containers/forms/PlanForm/tests/helpers.test.ts
```diff
@@ -0,0 +1,54 @@
+import { InterventionType } from '../../../../store/ducks/plans';
+import { PlanSchema } from '../helpers';
+
+/** Answers : among the validation errors raised, is there an error obj
+ * associated with the caseNum property resulting from caseNum being required
+ */
+const isCaseNumErrorPresent = (errors: any): boolean => {
+  const caseNumAssociatedErrors = errors.inner.filter(
+    (err: any) => err.path === 'caseNum' && err.message === 'Required'
```
moshthepitt: Do we have to use `any` here?

p-netm: will investigate and refactor accordingly

.circleci/config.yml
p-netm: Well, a rudimentary explanation goes like this:
Apparently, jest hits some performance issues on CD/CI server environments, Why?, I don't know yet. This causes circle ci builds to timeout with no output. This flag seems to fix the problem by "Running the tests sequentially ".

This is not much to ride on, and I will have a full explanation by the time am done resolving this issue

p-netm:
> This PR does not close #250 adequately. There are other conditional fields that have not been handled e.g. fiReason, fiStatus (list not conclusive)

Will readd this nice WIP label here, and proceed hacking at this.

.circleci/config.yml
moshthepitt: Nice work on finding this.  Can you move this commit to a different PR that we can merge ASAP so that we unblock the other PRs stuck with broken tests?

p-netm:
Caveats around the Yup schema validations:
- This tests are more helpful as regression tests rather than helping pickout current errors. This is because i wrote the tests to fit the current schema validation code. 
- To better be able to refine this tests , i would need to have context on what exactly makes each of the properties in the schema valid, this is knowledge that  i currently  do not have.

src/containers/forms/PlanForm/tests/helpers.test.ts
p-netm: Haven't found any useful leads on this yet, might need some help on this and yup schema type declaration.

moshthepitt:
@p-netm let's schedule some time next week to pair on this.  For now, let's put it on the back burner.

p-netm:
> @p-netm let's schedule some time next week to pair on this. For now, let's put it on the back burner.

Sounds good.

moshthepitt:
@p-netm what is the status of this.  Do you think you'll be able to get this closed, or should we close this PR?

p-netm:
I feel like this is already covered, there are already better tests that check for planFormVallidation, i thnk this can be closed.

Issue #299, opened 2019-08-06
FI Planning: Add list page


moshthepitt:
It seems to me that when creating new plans, we need a page that lists plans that come directly from the OpenSRP server and not the Superset tables - these have to go through an entire ETL process which may last too long for a person who is simply creating new plans.

Additionally, we should keep in mind that the Superset tables are meant more for analysis than for planning.

### Tasks

- [x] Create OpenSRP plan Redux module
- [x] Create OpenSRP plan list page

PR #305, opened 2019-08-07
Add Plans list CRUD page

moshthepitt:
Fix: #299

This PR adds a simple page that list plans gotten directly from the OpenSRP server.  The intention is for this to be used when creating or updating plans.

### Why is this necessary?

1. For the case of FI plans we cannot rely on the existing page that lists plans.  This page is more suited for reprting AND it takes several minutes before a plan shows up there.  We should also separate planning and analysis/reporting.
2. From a technical point of view the data source for the proposed plan list page is a list of plan definition objects.  In the case of FI, the data source of the data presented is the plans materialized view which has a different data structure.  To me it makes most to use one data source and one data structure when doing planning activities.  This makes things easier to reason about, and eliminates the need to translate between the two data structures.


moshthepitt:
### manage plans page

manage plans

### planning menu

planning

### reporting menu

reporting

moshthepitt:
> Just a small visual thing, let's make the `Add Focus Investigation` button vertically aligned with the header:
> 
> Screen Shot 2019-08-08 at 9 18 56 AM
> 
> Otherwise this is a great PR for incrementally unifying the Planning process between FI and IRS!

Done!

done

PR #314, opened 2019-08-08
New payloads from actions should be added to existing records in store

### LINKED ISSUE #289: Investigate why Redux store is being reloaded
When you first log in to the web app, and you navigate to the active FI page, the web app requests for all those plans from Superset.

When you go to an individual plan and then go back to the list page, you will notice that it re-loads the list from scratch instead of using what it should already have had in the Redux store.

p-netm:
Tests to check if payloads from dispatched actions replace the existing records. The expected behavior is that new dispatched records should be added to existing records if the record was not in the store. 

closes #289


p-netm:
> Looks like you prefer "REMOVE" to reset. Can you make that change in the reducer module introduced in this PR #305 - for consistency?

done

src/store/ducks/jurisdictions.ts
```diff
@@ -96,6 +104,11 @@ export default function reducer(
         ...state,
         allJurisdictionIds: action.allJurisdictionIds,
```
moshthepitt: Looks like jurisdictions are still replaced instead of merged.  Is this the case?  Do we have a test for this?

Issue #318, opened 2019-08-08
IRS Planning - Jurisdiction Selection Map Styling


cKellyDesign:
A few more interactions need to be flushed out:

- [x] Loading an existing draft should have selection `fill-color` stops generated on load
- [x] Toggling the `Toggle All` checkbox in the drilldown table should update the selection `fill-color` stops
- [x] When drilled down, `Toggle All` checkbox should only descendant Jurisdiction selections

moshthepitt:
Fixed by #319

PR #319, opened 2019-08-08
Set IRS Jurisdiction Selection Map Styling

cKellyDesign:
For #318


moshthepitt:
@cKellyDesign it is not clear how this affects a plan such as `c28bfee9-5af4-4e2b-a848-dcb8be39816c` (`Berg IRS 2 Namibia`), for example.  I am really not sure what clicking on the map does or is supposed to do

cKellyDesign:
>  I am really not sure what clicking on the map does or is supposed to do.

@moshthepitt You're right that this interaction is not the most intuitive and needs some clarity. @mberg suggested having like a note in the tooltip hinting at how to use the map.

> it is not clear how this affects a plan such as c28bfee9-5af4-4e2b-a848-dcb8be39816c (Berg IRS 2 Namibia)

The biggest visual difference here is that the Plan will actually load with its selection styling (where deselected jurisdictions have less opacity) instead of being triggered the first time a user updates the selection.

Then there is a functional update here too, wherein the Toggle All checkbox will toggle the selection of all descendant jurisdictions to match the checkbox (turn all off/on at once (before each jurisdiction would simply toggle individually)).

cKellyDesign:
>  I am really not sure what clicking on the map does or is supposed to do.

@moshthepitt You're right that this interaction is not the most intuitive and needs some clarity. @mberg suggested having like a note in the tooltip hinting at how to use the map.

> it is not clear how this affects a plan such as c28bfee9-5af4-4e2b-a848-dcb8be39816c (Berg IRS 2 Namibia)

The biggest visual difference here is that the Plan will actually load with its selection styling (where deselected jurisdictions have less opacity) instead of being triggered the first time a user updates the selection.

Then there is a functional update here too, wherein the Toggle All checkbox will toggle the selection of all descendant jurisdictions to match the checkbox (turn all off/on at once (before each jurisdiction would simply toggle individually)).

cKellyDesign:
Adding `WIP` label back due to performance issue which seems to originate from changes in this PR, continued from https://github.com/onaio/reveal-frontend/pull/379#issuecomment-522776549

I have done some perf testing and it appears call happens only after the final step of initializing the Gisida Map, with everything else happening in around ~15 seconds.

Screen Shot 2019-08-19 at 2 30 09 PM

**Things which are NOT causing this issue:**
1. The loading of GeoJSON #379  is not loading any Namibia GeoJSON at all
2. The lack of GeoJSON (issue occurs in this branch which is loading GeoJSON)
3. The new `diimagery` API key (issue occurs with fallback Mapbox v7 satellite layers)
4. The new Namibia Tilesets (issue occurs in this branch and #379 but not #378)

**Things which COULD be causing this issue:**
1. Double Initializing the Gisida map (the first stalls ~7sec, the second stalls ~1.3min), causing an initialization rate throttling?
2. Mapbox Account tileset throttling? (seems unlikely because we are not close to our limits)

src/containers/pages/InterventionPlan/IRS/plan/index.tsx
```diff
@@ -1775,14 +1914,18 @@ class IrsPlan extends React.Component<
...
-        Header: () => (
-          <Input
-            checked={planJurisdictionIds.length === filteredJurisdictions.length}
-            className="plan-jurisdiction-select-all-checkbox"
-            onChange={onToggleAllCheckboxChange}
-            type="checkbox"
-          />
-        ),
+        Header: () => {
+          const isChecked = this.getIsJurisdictionPartiallySelected(this.state.focusJurisdictionId);
```
cKellyDesign: This line right here is consuming tons of browser resources, simplifying it to use what we already have in state instead of re-calculating if the administrative jurisdiction has any selected descendant operational jurisdictions.

cKellyDesign:
Removing WIP flag b/c I believe the cause of the issue described in https://github.com/onaio/reveal-frontend/pull/319#issuecomment-522777613 has been resolved via 9d4a0f3.

@moshthepitt - I think this is ready for review, the description of what this PR at https://github.com/onaio/reveal-frontend/pull/319#issuecomment-522586793 is still accurate.

Issue #293, opened 2019-08-05
FI Plan: Test PlanForm helpers


moshthepitt:
This is part of the work to fix #260.

FrankApiyo:
- [ ] test extractActivityForForm

- [ ] test getFormActivities

- [ ] test extractActivitiesFromPlanForm

- [ ] test getNameTitle

- [ ] test doesFieldHaveErrors

- [ ] test generatePlanDefinition

PR #324, opened 2019-08-09
Add tests to planform helpers

FrankApiyo:
Fixes: #293


src/containers/forms/PlanForm/tests/helpers.test.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  it('check extractActivitiesFromPlanForm returns the correct value', () => {
+    MockDate.set('1/30/2000', 0);
+    expect(extractActivitiesFromPlanForm(activities)).toEqual(
+      expectedExtractActivityFromPlanformResult
+    );
+    MockDate.reset();
+  });
+
+  it('check getNameTitle returns the correct value when Focus Investigation(FI) is selected', () => {
```
moshthepitt: Can we redo this test to include the jurisdiction?

src/containers/forms/PlanForm/tests/helpers.test.ts
```diff
@@ -0,0 +1,81 @@
...
+  expectedPlanDefinition,
+  extractedActivitiesFromForms,
+  planActivities,
+  values,
+  values2,
+} from './fixtures';
+
+describe('PlanForm/helpers', () => {
+  it('check extractActivityForForm returns the correct value', () => {
+    expect(extractActivityForForm(bccTestPlanActivity.BCC)).toEqual(expectedActivity);
```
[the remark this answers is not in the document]
FrankApiyo: I'll add tests exactly like this one for the following activities: 

- [ ] Bednet Distribution

- [ ] Blood Screening

- [ ] Case Confirmation

- [ ] Larval Dipping

- [ ] Family Registration

- [ ] Mosquito Collection

src/containers/forms/PlanForm/tests/helpers.test.ts
```diff
@@ -0,0 +1,144 @@
...
+  it('check extractActivitiesFromPlanForm returns the correct value', () => {
+    MockDate.set('1/30/2000', 0);
+    expect(extractActivitiesFromPlanForm(activities)).toEqual(
+      expectedExtractActivityFromPlanformResult
+    );
+    MockDate.reset();
+  });
+
+  it('check getNameTitle returns the correct value when Focus Investigation(FI) is selected', () => {
+    expect(getNameTitle(event, values)).toEqual(['A1--2019-08-09', 'A1  2019-08-09']);
```
moshthepitt: Two things:

1. The generated value is missing the jurisdiction name.
2. Where does the date `2019-08-09` come from?

FrankApiyo: When the getNameTitle function is called when a location is not selected; then a jurisdiction name will be missing as you see here. But on the following line, you see that the function is called when a location is selected.

The date comes from here: https://github.com/onaio/reveal-frontend/blob/master/src/containers/forms/PlanForm/helpers.ts#L324 and/or here: https://github.com/onaio/reveal-frontend/blob/master/src/containers/forms/PlanForm/helpers.ts#L315. As you can see, it gets the current date

moshthepitt: > When the getNameTitle function is called when a location is not selected; then a jurisdiction name will be missing as you see here. But on the following line, you see that the function is called when a location is selected.

Thanks.

> The date comes from here: https://github.com/onaio/reveal-frontend/blob/master/src/containers/forms/PlanForm/helpers.ts#L324 and/or here: https://github.com/onaio/reveal-frontend/blob/master/src/containers/forms/PlanForm/helpers.ts#L315. As you can see, it gets the current date

Thanks.  However the value there is not the current date.  Where does **that** value come from?

FrankApiyo: It comes from the fixtures

Issue #251, opened 2019-07-31
Capture all tasks that should render symbol layers


kahama94:
Current implementation only tests Mosquito collection symbol leaving larval dipping and case confirmation untested. Inclusion of this will ensure tests cover all symbol layer instances

FrankApiyo:
I'll change task6 here: https://github.com/onaio/reveal-frontend/blob/master/src/store/ducks/tests/fixtures.ts#L658 to be a larval_dipping task and then task7 here: https://github.com/onaio/reveal-frontend/blob/master/src/store/ducks/tests/fixtures.ts#L699 to be a case confirmation task. Both these tasks are at the moment bed net distribution tasks

kahama94:
Fine by me since we have bednet distribution tested in other cases

cKellyDesign:
@kahama94 @FrankApiyo what's the status of this ticket?

kahama94:
I think we can move the the list that holds goals to a settings file.  After that We can test the list by matching snapshots for the map layers
 ``` store.getState()['map-1'].layers ```. This will mean when someone adds or removes a new goal on the list the tests will fail requiring snapshot update

moshthepitt:
This is no longer necessary as we'll be [getting rid of GisidaWrapper](https://github.com/onaio/reveal-frontend/issues/1150) instead.

PR #327, opened 2019-08-13
Add test that ensures that all the symbol layers are rendered

FrankApiyo:
Fixes: #251


FrankApiyo:
Will add better description to this PR

src/configs/settings.ts
```diff
@@ -205,6 +205,12 @@ export const PlanActivityTitles = [
   'BCC',
   'IRS',
 ] as const;
+
+export const goalsWithSymbols = [
```
moshthepitt: You need to add the values from the `planActivities` object e.g. `Larval_Dipping` should also be added to this list

FrankApiyo:
@moshthepitt do you think replacing a very large snapshot with smaller snapshots that will not necessarily usually fail at the same time is a good solution to reducing the sizes of snapshots. The snapshot file itself will still remain large. But individual snapshots will reduce in size significantly.

moshthepitt:
> @moshthepitt do you think replacing a very large snapshot with smaller snapshots that will not necessarily usually fail at the same time is a good solution to reducing the sizes of snapshots. The snapshot file itself will still remain large. But individual snapshots will reduce in size significantly.
> 
> That is what I just did in this last commit

Yes I think this is better because it is more readable and requires less resources to validate that the spanshot is valid.

src/configs/settings.ts
moshthepitt: @FrankApiyo did you address this?

FrankApiyo: Not yet. I'm on it.

moshthepitt: Is there a reason why you have not added `Mosquito_Collection`?

moshthepitt:
@FrankApiyo  what is the status of this?  Do you think you'll be able to get this closed, or should we close this PR?

FrankApiyo:
> @FrankApiyo what is the status of this? Do you think you'll be able to get this closed, or should we close this PR?

@moshthepitt I don't think I'll be able to get this closed this week

moshthepitt:
> @moshthepitt I don't think I'll be able to get this closed this week

@FrankApiyo could you document what you think the next steps for this are?

FrankApiyo:
> > @moshthepitt I don't think I'll be able to get this closed this week
> 
> @FrankApiyo could you document what you think the next steps for this are?

@kahama94 has volunteered to take it from here

moshthepitt:
@FrankApiyo that's fine but can you still document the next steps for this please?

FrankApiyo:
> @FrankApiyo that's fine but can you still document the next steps for this please?

The following are the next steps for this PR IMHO:

- add the correct goals with symbols to the goals_with_symbols array in settings.ts

- attempt to reduce the sizes of the snapshots

moshthepitt:
Looks like a stale PR and branch.

Issue #311, opened 2019-08-08
Make pill on FI Plan smaller


mberg:
<img width="1183" alt="Screen Shot 2019-08-07 at 8 45 06 PM" src="https://user-images.githubusercontent.com/57921/62667163-756d4e80-b954-11e9-9201-9e3e7d964fd7.png">

Not high priority

PR #328, opened 2019-08-13
Reduce font size of pill

FrankApiyo:
Fixes: #311


src/containers/pages/FocusInvestigation/map/active/style.css
```diff
@@ -47,3 +47,7 @@ div.css-1pahdxg-control {
 div.mapSidebar button.btn.btn-primary.btn-lg.btn-block {
   margin-top: 1em;
 }
+
+.badge {
```
moshthepitt: Use `btn-sm` or `btn-xs`

FrankApiyo:
I should note that the pill will not be smaller in other places it's used in the code base. Should I do something about this?

Issue #316, opened 2019-08-08
Add test for LinkToNewPlans


moshthepitt:
This component is not tested

PR #330, opened 2019-08-14
Add tests to LinkToNewPlans component

FrankApiyo:
Fixes: #316


src/components/LinkToNewPlans/test/__snapshots__/index.test.tsx.snap
```diff
@@ -0,0 +1,44 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`components/LinkToNewPlans Matches snapshot 1`] = `
+<Router
```
moshthepitt: Can you limit the output of the snapshot to just `LinkToNewPlans`?

Something like this in the test should work: `expect(wrapper.find('Router').length).toEqual(1)` and `expect(toJson(wrapper.find('LinkToNewPlans')).toMatchSnapshot();`

Issue #238, opened 2019-07-29
Move magic strings from IRS plans to Constants


cKellyDesign:
There are currently a number of display text elements which should ideally be set in the constants file for consistency and easier implementation.

cKellyDesign:
- [ ] Audit for magical strings 🌈

PR #331, opened 2019-08-14
Add IRS_PLANS constant and remove corresponding magic string

FrankApiyo:
Fixes: #238


FrankApiyo:
I only found the one string that I have removed here. I don't think there are any others

Issue #338, opened 2019-08-14
Disable certain FI PlanForm Fields


cKellyDesign:
As per slack convo starting at https://onaio.slack.com/archives/CBSU8L2SC/p1565793929024800

Let's disable certain inputs to lock down their default values:
- [x] FI Reason
  - [x] case triggered = investigation
  - [x] routine = routine
- [x] priority - medium
- [x] activity start and end dates (for IRS activities)

cKellyDesign:
This still needs some work, it looks like the plan definition payload isn't save the correct timeStart/timeEnd on the actions (timingPeriod.end should be `2019-09-14` to match the plan end date):

<img width="472" alt="Screen Shot 2019-08-14 at 12 51 11 PM" src="https://user-images.githubusercontent.com/4141527/63051795-ba344080-be92-11e9-9a26-7c5d333997a1.png">

Issue #288, opened 2019-08-02
Show correct FI details on map


moshthepitt:
![correct fi details](https://user-images.githubusercontent.com/372073/62385507-f47e1380-b55d-11e9-91e7-1a48ecebce04.png)

cKellyDesign:
@moshthepitt what is the expected result here?

moshthepitt:
> what is the expected result here?

@cKellyDesign This was not scoped.  @craigappl @mberg can you provide guidance?

mberg:
Sorry I think it’s just to show the right details about the Fi Eg The classification status isn’t right.  I think that’s all it is

moshthepitt:
@mberg but what details do we want to show?  Are we showing only the classification status?

cKellyDesign:
Let's start by just showing the classification status. (Unless there is more required @mberg?)

cKellyDesign:
> Let's start by just showing the classification status.

We're already showing start and end dates, so the task for this ticket should be:
- [ ] Add classification status above start and end dates

cKellyDesign:
closing this issue because non-routine plans are already dynamically displaying the plan's Classification Status from the plan definition.

moshthepitt:
@cKellyDesign but currently saying Case classification === "FI" is not accurate.  We need to at least change that - no?

cKellyDesign:
What are the values to be displayed?

moshthepitt:
@cKellyDesign idk which of the values to display

cKellyDesign:
Closing this as it seems to be a non-issue.

PR #342, opened 2019-08-14
[DNM] 338 disabling plan form fields

cKellyDesign:
For #338


moshthepitt:
@cKellyDesign 

The `ActionReasonType` only allows 'Investigation', and 'Routine' while `FIReasonType` allows only 'Routine', and 'Case Triggered'.  Can we get confirmation that these two fields do indeed hold exactly the same kind of data?

cKellyDesign:
> The ActionReasonType only allows 'Investigation', and 'Routine' while FIReasonType allows only 'Routine', and 'Case Triggered'. 

@moshthepitt The relation between Plans, Activities/Tasks, and Goals is where I usually get very confused.

> Can we get confirmation that these two fields do indeed hold exactly the same kind of data?

cc @craigappl @githengi @mberg

cKellyDesign:
> The ActionReasonType only allows 'Investigation', and 'Routine' while FIReasonType allows only 'Routine', and 'Case Triggered'.

@moshthepitt is this the same type of content being discussed here (👉  https://github.com/onaio/reveal-frontend/issues/288#issuecomment-521534421) or is something else?

cKellyDesign:
@mberg with the recent changes from adding the Edit FI PlanForm page, the tests which are failing here are indicating that my method of disabling certain fields isn't really congruent anymore. Since this was a pretty aesthetic for-the-demo request, I'm going ask that we don't do this and close the PR instead of trying to merge it (just to undo it later).

Issue #337, opened 2019-08-14
Fix linting errors in Planform helpers test folder


FrankApiyo:
There are several linting errors in src/containers/forms/PlanForm/tests/helpers.test.ts

Issue #335, opened 2019-08-14
Fix Deprecation warning in helpers.test.ts


moshthepitt:
```sh
console.warn node_modules/moment/moment.js:293
      Deprecation warning: value provided is not in a recognized RFC2822 or ISO format. moment construction falls back to js Date(), which is not reliable across all browsers and versions. Non RFC2822/ISO date formats are discouraged and will be removed in an upcoming major release. Please refer to http://momentjs.com/guides/#/warnings/js-date$ for more info.
      Arguments: 
      [0] _isAMomentObject: true, _isUTC: false, _useUTC: false, _l: undefined, _i: Thu Jul 20 2017 22:31:00 GMT+0300 (East Africa Time), _f: undefined, _strict: undefined, _locale: [object Object]
      Error: 
          at Function.createFromInputFallback (/home/mosh/ona/reveal-frontend/node_modules/moment/moment.js:320:98)
          at configFromString (/home/mosh/ona/reveal-frontend/node_modules/moment/moment.js:2385:15)
          at configFromInput (/home/mosh/ona/reveal-frontend/node_modules/moment/moment.js:2611:13)
          at prepareConfig (/home/mosh/ona/reveal-frontend/node_modules/moment/moment.js:2594:13)
          at createFromConfig (/home/mosh/ona/reveal-frontend/node_modules/moment/moment.js:2561:44)
          at createLocalOrUTC (/home/mosh/ona/reveal-frontend/node_modules/moment/moment.js:2648:16)
          at createLocal (/home/mosh/ona/reveal-frontend/node_modules/moment/moment.js:2652:16)
          at hooks (/home/mosh/ona/reveal-frontend/node_modules/moment/moment.js:12:29)
          at extractActivityForForm (/home/mosh/ona/reveal-frontend/src/containers/forms/PlanForm/helpers.ts:148:11)
          at Object.<anonymous> (/home/mosh/ona/reveal-frontend/src/containers/forms/PlanForm/tests/helpers.test.ts:29:12)
          at Object.asyncJestTest (/home/mosh/ona/reveal-frontend/node_modules/jest-jasmine2/build/jasmine_async.js:108:37)
          at /home/mosh/ona/reveal-frontend/node_modules/jest-jasmine2/build/queue_runner.js:56:12
          at new Promise (<anonymous>)
          at mapper (/home/mosh/ona/reveal-frontend/node_modules/jest-jasmine2/build/queue_runner.js:43:19)
[...]

PR #354, opened 2019-08-16
Fix linting errors

FrankApiyo:
Fixes: #337 and also Fixes: #335


FrankApiyo:
This also fixes #335 because `moment` usage is replaced with `parseISO`

PR #363, opened 2019-08-16
Only show plans with Active/Complete plan status in ActiveFI

### LINKED ISSUE #359: Remove draft/retired plans from FI plan reporting list
We don't need to show these plans on pages meant for reporting, and should filter them out.

p-netm:
refactor fiilter arguments passed to the plans selectors

closes #359 .


p-netm:
I dont think that this requires any new tests. The test in this case would be essentially exercising the selectors in use, which have already been tested in the dux module. for the component, there are already tests that show the component will render whatever the selector gives it, so i "think" its all good.

Issue #298, opened 2019-08-06
PlanForm: Dynamic plan activities


moshthepitt:
We need to make it possible to alter the activities in the PlanForm.

This means:

- ability to add new activities
- ability to remove existing activities

### Constraints

- IRS plans can only have the IRS activity
- FI plans cannot have the IRS activity

cKellyDesign:
@mberg @craigappl how soon is this feature required?

moshthepitt:
It's not a lot of work actually!  The biggest issue is the design but I'll implement by Thursday next week and then if we need to redesign we'll do so

cKellyDesign:
@moshthepitt is this adequately scoped?

moshthepitt:
Nope.  But I have enough to work with

PR #364, opened 2019-08-16
Add support for dynamic plan activities

moshthepitt:
Fix: #298


src/containers/forms/PlanForm/style.css
```diff
@@ -0,0 +1,11 @@
+.close {
+  color: #6c757d !important;
+  font-size: 30px;
+}
+
```
cKellyDesign: @moshthepitt can we add some margin-bottom to these buttons?

Screen Shot 2019-08-16 at 10 13 48 AM

moshthepitt: Yes! Will do

moshthepitt: activities

 done

Issue #373, opened 2019-08-19
Add tests for NewPlan


moshthepitt:
We need tests for the NewPlan component.

PR #375, opened 2019-08-19
Add tests to NewPlan component

FrankApiyo:
Fixes: #373


src/containers/pages/InterventionPlan/NewPlan/test/__snapshots__/index.test.tsx.snap
```diff
@@ -0,0 +1,238 @@
...
+<h3
+  className="mb-3 page-title"
+>
+  Create New Plan
+</h3>
+`;
+
+exports[`containers/pages/NewPlan renders correctly: plan form props 1`] = `
+Object {
+  "allFormActivities": Array [
```
moshthepitt: Can we make this comparison against a fixture as opposed to have it in the snapshot?

Something like:

```ts
expect(wrapper.find('PlanForm').props()).toEqual(fixtures.theExpoectedProps);
```

FrankApiyo:
the snapshot is only 10 lines now

src/containers/pages/InterventionPlan/NewPlan/test/index.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+      <Router history={history}>
+        <NewPlan />
+      </Router>
+    );
+
+    // check that page title is displayed
+    expect(toJson(wrapper.find('h3.mb-3.page-title'))).toMatchSnapshot('page title');
+
+    // check that planform receives the correct props
+    expect(JSON.stringify(wrapper.find('PlanForm').props())).toEqual(JSON.stringify(someObject));
```
moshthepitt: 1. Why do we have to call `JSON.stringify` on both sides?
2. Can we name it something better than `someObject`?

p-netm: I think am missing something; how/at what point is  `someObject` passed in into PlanFrom as a prop?

moshthepitt: We are comparing the value of `wrapper.find('PlanForm').props()` to `someObject`

PR #378, opened 2019-08-19
377 Fix IRS List Page

### LINKED ISSUE #377: Fix IRS Plans List Page
Currently the IRS page is not successfully loading lists of IRS plans because the it needs to dynamically find which `useContext` list item should be used for inferring `interventionType`.

cKellyDesign:
Fixes #377


cKellyDesign:
@moshthepitt unfortunately i think this will constitute a release `v0.1.9`

src/constants.tsx
```diff
@@ -208,6 +208,8 @@ export const ROUTINE_PLAN = 'Routine';
 export type ROUTINE_PLAN = typeof ROUTINE_PLAN;
 export const CASE_TRIGGERED_PLAN = 'Case-triggered';
 export type CASE_TRIGGERED_PLAN = typeof CASE_TRIGGERED_PLAN;
+export const INTERVENTION_TYPE = 'interventionType';
```
moshthepitt: Do you think that we can use the value from `useContextCodes` in settings.ts instead of adding a constant?

src/constants.tsx
```diff
@@ -208,6 +208,8 @@ export const ROUTINE_PLAN = 'Routine';
 export type ROUTINE_PLAN = typeof ROUTINE_PLAN;
 export const CASE_TRIGGERED_PLAN = 'Case-triggered';
```
moshthepitt: Do you think that we can use the value from `FIReasons` in settings.ts instead of a new constant? 

Also, I think that this string should not be 'Case-triggered' but 'Case Triggered' (which is what is in the settings module, coming from this spec: https://docs.google.com/spreadsheets/d/1kGPgyF0eJdag19FFL1Oog2gdphdPtKobY4EQ0gMgtgw/edit#gid=1805975071).

cKellyDesign: There's a few conditionals using this `CASE_TRIGGERED_PLAN`, am switching them to use `FIReasons[1]` instead, not sure if this will fix things that were silently broken before. I have confirmed that plans are using `Case Triggered` in the useContext array.

PR #379, opened 2019-08-19
IRS Jurisdiction Selection Sans-Geojson

cKellyDesign:
For #8 (to get Namibia running on tilesets alone), should follow ~~#378~~ and #319


cKellyDesign:
So far we have successfully rendered and modified a Plans jurisdiction selection via only tileset geometries. Since making changes (today) I am experiencing stalled Mapbox GL JS requests consitently blocking rendering for 1.3-1.4 **minutes** every time we load a draft IRS plan in the Jurisdiction Selection page.

I have done some perf testing and it appears call happens only after the final step of initializing the Gisida Map, with everything else happening in around ~15 seconds.

Screen Shot 2019-08-19 at 2 30 09 PM

This issue is not presenting itself on master or #378, but is in #319 and this branch, since this branch is branched off of #319 the issue must be introduced in #319.

**Things which are NOT causing this issue:**
1. The loading of GeoJSON this branch is not loading any Namibia GeoJSON at all
2. The lack of GeoJSON (issue occurs in #319 which is loading GeoJSON)
3. The new `diimagery` API key (issue occurs with fallback Mapbox v7 satellite layers)
4. The new Namibia Tilesets (issue occurs in this branch and #319 but not #378)

**Things which COULD be causing this issue:**
1. Double Initializing the Gisida map (the first stalls ~7sec, the second stalls ~1.3min), causing an initialization rate throttling?
2. Mapbox Account tileset throttling? (seems unlikely because we are not close to our limits)

mberg:
It’s not mapbox.

cKellyDesign:
on `Things which COULD be causing this issue: #1`:
I've debugged this a bit and have fixed the double initialization issue and now the stall is still present but ~half the length.

Screen Shot 2019-08-20 at 7 31 03 AM

cKellyDesign:
@moshthepitt this PR is also ready for review, biggest changes [between this branch and #319](https://github.com/onaio/reveal-frontend/compare/318-jurisdiction-selection-styling...8-irs-namibia-tilesets) are: 

1. Supporting the IRS Jurisdiction Selection page with only Tilesets by distinguishing between administrative and operational tilesets
2. Updating the tilesets being used to ones which contain data matching OpenSRP
3. Using `opensrp_api/location/findByProperties` instead of `opensrp_api/location/<ID>` to avoid ever loading Admin level 0 geojson

PR #389, opened 2019-08-21
Refactor handlers for stopPropagation & PreventDefault

### LINKED ISSUE #258: Add docstrings
These functions need docstrings:

- [ ] stopPropagation
- [ ] preventDefault
- [ ] stopPropagationAndPreventDefault

p-netm:
closes #258

### DIFF
--- a/src/containers/pages/InterventionPlan/IRS/plan/index.tsx
+++ b/src/containers/pages/InterventionPlan/IRS/plan/index.tsx
@@ -817,7 +817,7 @@ class IrsPlan extends React.Component<
   }
   /** onSaveEditNameButtonClick - handler disabling inline-editing */
   private onSaveEditNameButtonClick(e: MouseEvent) {
-    e.preventDefault();
+    preventDefault(e);
     this.setState({
       isEditingPlanName: false,
       previousPlanName: '',
--- a/src/helpers/utils.tsx
+++ b/src/helpers/utils.tsx
@@ -398,17 +398,29 @@ export function roundToPrecision(n: number, precision: number = 0): number {
   return Math.round(n * factor) / factor;
 }
 
-export function stopPropagation(e: Event | MouseEvent | any) {
-  e.stopPropagation();
-}
-
-export function preventDefault(e: Event | MouseEvent | any) {
+/** click handler that cancels an event's default behavior
+ *
+ * @param {Event | MouseEvent} e - a synthetic event wrapper around native dom events
+ */
+export function preventDefault(e: MouseEvent) {
   e.preventDefault();
 }
 
-export function stopPropagationAndPreventDefault(e: Event | MouseEvent | any) {
-  preventDefault(e);
-  stopPropagation(e);
+/** click handler that cancels an event's default behavior
+ *
+ * @param {MouseEvent} e - a synthetic event wrapper around native dom events
+ */
+export function stopPropagation(e: MouseEvent) {
+  e.stopPropagation();
+}
+
+/** click handler that cancels both an event propagation and its default behavior
+ *
+ * @param {MouseEvent} e - a synthetic event wrapper around native dom events
+ */
+export function stopPropagationAndPreventDefault(e: MouseEvent) {
+  e.preventDefault();
+  e.stopPropagation();
 }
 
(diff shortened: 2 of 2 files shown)


src/helpers/utils.tsx
```diff
@@ -398,17 +398,21 @@ export function roundToPrecision(n: number, precision: number = 0): number {
...
 
-export function stopPropagation(e: Event | MouseEvent | any) {
-  e.stopPropagation();
-}
-
+/** handler that cancels an event's default behavior
+ *
+ * @param {Event | MouseEvent} e - a synthetic event wrapper around native dom events
+ */
 export function preventDefault(e: Event | MouseEvent | any) {
```
moshthepitt: It seems that this one is not used anywhere?

src/containers/pages/InterventionPlan/IRS/plan/index.tsx
```diff
@@ -48,7 +45,6 @@ import {
 } from '../../../../../constants';
 import {
   FlexObject,
-  preventDefault,
   RouteParams,
   stopPropagation,
```
moshthepitt: Is this one still needed?

src/containers/pages/InterventionPlan/IRS/plan/index.tsx
```diff
@@ -741,7 +737,7 @@ class IrsPlan extends React.Component<
   // Jurisdiction Hierarchy Control
   /** onTableBreadCrumbClick - handler for drilldown table breadcrumb clicks to reset the table hierarchy */
   private onTableBreadCrumbClick = (e: MouseEvent) => {
-    preventDefault(e);
+    e.preventDefault();
```
cKellyDesign: This is pretty low hanging fruit, if this is too much hassle then let's just not have util functions for these.

src/helpers/utils.tsx
p-netm: yeah the idea was to only leave the one that has them both as a standalone function.
Dont think i was sober enough during this pr

src/containers/pages/InterventionPlan/IRS/plan/index.tsx
p-netm: yes,

src/containers/pages/InterventionPlan/IRS/plan/index.tsx
p-netm: `stopPropagation` and `stopPropagationAndPreventDefault` have to be click handler functions for the syntax to work:
```
```

cKellyDesign: In the second example, I think we either replace `e.preventDefault()` with `preventDefault(e)` or we can just leave it as is.

p-netm: ok.

src/containers/pages/InterventionPlan/IRS/plan/index.tsx
```diff
@@ -817,7 +817,7 @@ class IrsPlan extends React.Component<
   }
   /** onSaveEditNameButtonClick - handler disabling inline-editing */
   private onSaveEditNameButtonClick(e: MouseEvent) {
-    e.preventDefault();
+    preventDefault();
```
cKellyDesign: We'll need to pass `e` into `preventDefault` here

p-netm: oh yeah. :+1:

cKellyDesign:
I think the main/original use case for having these functions is feeding them directly into component handler attributes:

```
<Button onClick={preventDefault} />
```
which was intended to be a bit DRYer than:
```
function handleSpecificButtonClick(e) { e.preventDefault() }
<Button onClick={handleButtonClick} />
```
and to avoid the linting error `Lambdas are forbidden in JSX attributes due to their rendering performance impact` when doing:
```
<Button onClick={(e) => { e.preventDefault() } />
```

cKellyDesign:
@moshthepitt thoughts on this? If you feel it's just excess overhead then let's scrap it and close #258

Issue #127, opened 2019-06-20
cleanup Handler helper


kahama94:
The handler helper hasn't used typescript to it's full potential.

cKellyDesign:
@kahama94 what is the status of this?

PR #415, opened 2019-08-26
Add more specific types on handlers component

kahama94:
fix: #127

### DIFF
--- a/src/helpers/handlers/index.ts
+++ b/src/helpers/handlers/index.ts
@@ -1,3 +1,4 @@
+import { Feature, Point } from 'geojson';
 import { GisidaMap } from 'gisida';
 import { LngLat, Map } from 'mapbox-gl';
 import { MAP_ID } from '../../constants';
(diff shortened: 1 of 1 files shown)


src/helpers/handlers/index.ts
```diff
@@ -11,14 +12,22 @@ declare global {
   }
   const mapboxgl: typeof mapboxgl;
 }
-
+export interface Feature {
```
moshthepitt: Do you think we should use the built-in `Feature` interface? i.e. `import { Feature } from 'geojson';`

kahama94: That would most preferred. I tried it out and found Feature interface does not have layer which we require in this case so I decided to use this one instead

moshthepitt: Should we then extend that built in feature interface, instead of creating a new one?

kahama94: yeah! most definitely

src/helpers/handlers/index.ts
```diff
@@ -12,22 +13,20 @@ declare global {
...
+ */
+export interface FeatureWithLayer extends Feature {
   layer: FlexObject;
-  properties: FlexObject;
-  source: string;
-  state: FlexObject;
-  type: string;
-  geometry?: FlexObject;
-  vectorTileFeature?: FlexObject;
+  geometry: Point;
```
moshthepitt: Doesn't the default implementation of `Feature` include geometry, which accepts points among other things?  What is the reason for changing it so that it is limited to just `Point`?

kahama94: True but It also accepts GeometryCollection which has no coordinates. I've included some docstrings to inform on this. I am overriding the geometry prop and strictly using Point since that is what we are dealing with in this scenario. What would be the best approach to this scenario ?

moshthepitt: Could you explain the problem in a little more detail?

src/helpers/handlers/index.ts
```diff
@@ -13,24 +21,31 @@ declare global {
...
+     * Geometry type is a union of seven types.
+     * Unfortunately, not all of those types include the coordinates property
+     * Let's exclude GeometryCollection which has no coordinates for our case
+     */
     if (
       feature &&
       feature.geometry &&
-      feature.geometry.coordinates &&
+      (feature.geometry as
+        | Point
```
moshthepitt: Could you please define a type to hold this and then use it here? For clarity?

moshthepitt:
@kahama94  what is the status of this?  Do you think you'll be able to get this closed, or should we close this PR?

kahama94:
The pr is ready fro review I'd say

src/helpers/handlers/index.ts
```diff
@@ -11,14 +21,25 @@ declare global {
...
+ * Let's exclude GeometryCollection which has no coordinates for our case when extending Feature
+ * FeatureWithLayer Interface extends Feature, Adds layer prop which is missing on Feature
+ */
+export interface FeatureWithLayer
+  extends Feature<Point | MultiPoint | LineString | MultiLineString | Polygon | MultiPolygon> {
+  layer: FlexObject;
+}
+export interface CustomGeometry {
+  geometry: Point | MultiPoint | LineString | MultiLineString | Polygon | MultiPolygon;
+}
```
moshthepitt: What is this used for?

kahama94: My bad. we don't need it

Issue #416, opened 2019-08-26
Cleanup IrsPlan module


moshthepitt:
After #402, the `IrsPlan` module and component needs to be audited for unused code and then cleaned.

cKellyDesign:
- [ ] Removes New Plan functionality
- [ ] Removes Edit Plan Name functionality
- [ ] Adds missing docstrings
- [ ] Removes isNewPlan state / functionality
- [ ] Removes unused imports and variables

Issue #237, opened 2019-07-29
Replace componentWillReceiveProps


cKellyDesign:
This is a ticket to track the work around replacing Reacts depricated `componentWillReceiveProps` lifecycle function.

cKellyDesign:
Seems like this would go hand in hand with #286, especially if we are loading Jurisdiction geojson as the user Drills down through the Jurisdiction hierarchy.

cKellyDesign:
@moshthepitt are there any good resources you can share on why and how?

moshthepitt:
### Why

[Update on Async Rendering (explains why this was deprecated)](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html)

### How

1.[You should move all the code from the componentWillMount to the constructor or componentDidMount](https://stackoverflow.com/a/49213753)
2. [Replacing ‘componentWillReceiveProps’ with ‘getDerivedStateFromProps’](https://hackernoon.com/replacing-componentwillreceiveprops-with-getderivedstatefromprops-c3956f7ce607)
3. [Simply use React hooks](https://reactjs.org/docs/hooks-intro.html)

cKellyDesign:
@moshthepitt am going to incorporate this approach in the IRS Report List and IRS Report Map views :)

moshthepitt:
This is no longer necessary as we'll be [getting rid of GisidaWrapper](https://github.com/onaio/reveal-frontend/issues/1150) instead.  `componentWillReceiveProps` was only used here.

PR #430, opened 2019-08-29
416 IRS plan cleanup

cKellyDesign:
For #416 - This PR is a starting point for IRS cleanup and adding tests

- Removes New Plan functionality
- Removes Edit Plan Name functionality
- Adds missing docstrings
- Removes `isNewPlan` state / functionality
- Removes unused imports and variables
- Fixes typos (`jurisidction` and `decendant`)
- Updates all method docstrings to include @params with typings
- Consolidates fill color settings
- Replaces some `any`s with more specific typings


cKellyDesign:
This PR is ready for review, it's just incremental progress towards #416 and is largely subtractive.

src/containers/pages/InterventionPlan/IRS/plan/tests/index.test.tsx
```diff
@@ -62,4 +64,31 @@ describe('containers/pages/IRS/plan', () => {
     expect(toJson(wrapper)).toMatchSnapshot();
     wrapper.unmount();
   });
+
+  // it('loads planById from Plans store', () => {});
```
moshthepitt: Could we remove this commented out code?

cKellyDesign:
> should we refactor to not use componentWillReceiveProps now instead of waiting for the (near) future?

Now might be a good time, since this is more refactoring than cleanup I would think this PR isn't the place.

cKellyDesign:
@moshthepitt - there are still some `any`s in this module, mostly around the following areas which I think may warrant their own tickets/PRs:

1. Gisida / Mapbox specific objects which don't have explicit types (gisida) or are missing properties (mapbox)
2. MouseEvents, for some reason the default MouseEvents and React synthetic events aren't playing nicely, I found this when trying to type them a while ago and don't remember the specific issue

moshthepitt:
> > should we refactor to not use componentWillReceiveProps now instead of waiting for the (near) future?
> 
> Now might be a good time, since this is more refactoring than cleanup I would think this PR isn't the place.

Do you think we should create an issue for this?

cKellyDesign:
>>> should we refactor to not use componentWillReceiveProps now instead of waiting for the (near) future?
>>
>> Now might be a good time, since this is more refactoring than cleanup I would think this PR isn't the place.
>
> Do you think we should create an issue for this?

@moshthepitt ya we have one here https://github.com/onaio/reveal-frontend/issues/237

Issue #382, opened 2019-08-20
IRS Reporting pages web UI


moshthepitt:
There are some dashboard mockups here https://smartregister.atlassian.net/wiki/spaces/REV/pages/1152516110/IRS+Forms+and+Web+UI+Dashboards+Namibia

We need to scope how exactly to come up with these dashboards in Reveal.

moshthepitt:
This is part of #192

cKellyDesign:
It is not at all clear to me what we are supposed to build for IRS reporting based on the link in the description.

@mberg @craigappl - What is the timeline on this and please provide mockups before scoping.

cc @rowo @moshthepitt

mberg:
@rowo please see this issue

https://smartregister.atlassian.net/wiki/spaces/REV/pages/1152516110/IRS+Forms+and+Web+UI+Dashboards+Namibia

It has what the table should look like.  this is in reveal powered by superset.  It's supposed to work very similar to how mspray dashboards currently work (minus for now some of the higher level map aggregation views)

rowo:
Navigation is a bit unrefined. Need new IRS reporting item. Maybe just use FI one and see what's different. Need to resolve conflict between old nav organized by FI and IRS vs. planning and reporting.

So new views similar to FI and apply to IRS. We only had planning before and that was pusehd to mSpray. Now it's going to live here in Reveal.

cKellyDesign:
Marking this as in progress b/c Roger is working on Mockups

rowo:
@mberg can you review?

Few questions:
- Are these all housed within a single IRS plan? 
- Is the user supposed to be able to navigate to other IRS plans from previous years or see any of that information?
- Is there an individual village page or is that just the map since we're only looking at the current IRS plan and don't need to see past IRS in that area?

### All Districts i.e. National

![IRS - all districts](https://user-images.githubusercontent.com/1584163/64225563-e9463c80-cea9-11e9-8b33-768b947cb470.png)

### Single District

![IRS - single district](https://user-images.githubusercontent.com/1584163/64225682-6ec9ec80-ceaa-11e9-8fcc-017117f3b088.png)

### Village Map

![IRS - village map](https://user-images.githubusercontent.com/1584163/64225856-2f4fd000-ceab-11e9-8626-2d87e666c398.png)

### District Map

![IRS - district map](https://user-images.githubusercontent.com/1584163/64225861-31b22a00-ceab-11e9-861c-fbcf48cb0a97.png)

[...]

PR #435, opened 2019-09-04
IRS Reporting Table

cKellyDesign:
For #382, should be merged after #430 

This branch adds the following env variables:
* `REACT_APP_SUPERSET_IRS_REPORTING_JURISDICTIONS_DATA_SLICE`
* `REACT_APP_SUPERSET_IRS_REPORTING_JURISDICTIONS_DATA_SLICE_NA`

[values in slack thread](https://onaio.slack.com/archives/CBSU8L2SC/p1568751455160000?thread_ts=1568749754.159500&cid=CBSU8L2SC)


cKellyDesign:
## Current Table Progress
- [x] dynamically add columns
- [ ] build coloring logic into `columnGetter` configs

Screen Shot 2019-09-10 at 7 55 14 PM

src/configs/env.ts
```diff
@@ -50,6 +50,10 @@ export const SUPERSET_JURISDICTIONS_DATA_SLICE =
   process.env.REACT_APP_SUPERSET_JURISDICTIONS_DATA_SLICE || '0';
 export type SUPERSET_JURISDICTIONS_DATA_SLICE = typeof SUPERSET_JURISDICTIONS_DATA_SLICE;
 
+export const SUPERSET_IRS_REPORTING_JURISDICTIONS_DATA_SLICE =
+  process.env.REACT_APP_SUPERSET_IRS_REPORTING_JURISDICTIONS_DATA_SLICE || '0';
```
cKellyDesign: New env variable to specify the superset slice @moshthepitt put together

cKellyDesign: New env variable to specify the superset slice @moshthepitt put together

src/configs/settings.ts
```diff
@@ -702,9 +704,140 @@ export const symbolLayerConfig = {
...
+  drilldownColumnGetters: {
+    [key: string]: () => any;
+  };
+  juridictionTyper: (j: any) => JurisidictionTypes;
+  sliceProps: string[];
+}
+/* tslint:disable:object-literal-sort-keys */
+export const irsReportingCongif: { [key: string]: IrsReportingCongif } = {
+  '550': {
+    drilldownColumnGetters: {
```
cKellyDesign: `drilldownColumnGetters` are functions returning columns for drilldown table props, this is where we can build in contextually-specific columns (like background-colors)

cKellyDesign: @moshthepitt @p-netm @kahama94 - would love your thoughts / critique of this approach of making the IRS Reporting pages configurable 🙏

src/helpers/hierarchy.ts
```diff
@@ -0,0 +1,89 @@
+import { ChildrenByParentId, Jurisdiction } from '../store/ducks/jurisdictions';
+
+/** getDescendantJurisdictionIds - recursive hierarchy util to get all ancestors of certain Jurisdictions
```
cKellyDesign: Shifting helper methods here which can be shared between IRS Planning and IRS Reporting

src/store/ducks/jurisdictions.ts
```diff
@@ -33,11 +36,29 @@ export interface AllJurisdictionIds {
...
 
+/** Interface descriing a list of all child ids of a given jurisdiction id */
+export interface ChildrenByParentId {
+  [key: string]: string[];
+}
+
+/** Interface descriing a list of all jurisdictions ids of a plan id */
+export interface JurisdictionIdsByPlanId {
+  [key: string]: string[];
+}
```
cKellyDesign: Storing strings to avoid re-computing hierarchical relations

src/containers/pages/InterventionPlan/IRS/report/index.tsx
```diff
@@ -0,0 +1,610 @@
...
+      jurisdictionsById: JurisdictionsById,
+      planById,
+      planId,
+      supersetService,
+    } = this.props;
+
+    // get Jurisdicitons from Store or Superset
+    const jurisdictionsArray = Object.keys(JurisdictionsById).length
+      ? Object.keys(JurisdictionsById).map(j => JurisdictionsById[j])
+      : await supersetService(SUPERSET_IRS_REPORTING_JURISDICTIONS_DATA_SLICE, {
```
cKellyDesign: Only load data if needed, might need to add some sort of checker helper to dynamically check if the jurisdictions need to be loaded based on IRS-type-specific conditions.

src/containers/pages/InterventionPlan/IRS/report/index.tsx
```diff
@@ -0,0 +1,610 @@
...
+            ),
+            id: 'name',
+          },
+        ],
+      },
+    ];
+
+    // define configuration for dynamic column generation
+    const config: IrsReportingCongif | undefined =
+      irsReportingCongif[SUPERSET_IRS_REPORTING_JURISDICTIONS_DATA_SLICE];
```
cKellyDesign: IRS Specific configs are called to build columns based on sliceId

src/configs/env.ts
moshthepitt: @cKellyDesign could we also add this to the sample file?  Also, please share the current value on #reveal or somewhere?

src/store/ducks/jurisdictions.ts
```diff
@@ -64,18 +97,36 @@ interface RemoveAllJurisdictionIdsAction extends AnyAction {
...
+  | RemoveJurisdictionIdsByPlanId
   | AnyAction;
 
 /** interface to describe jurisdiction state */
 interface JurisdictionState {
   allJurisdictionIds: AllJurisdictionIds;
-  jurisdictionsById: { [key: string]: Jurisdiction };
+  childrenByParentId: ChildrenByParentId;
+  jurisdictionIdsByPlanId: JurisdictionIdsByPlanId;
+  jurisdictionsById: { [key: string]: JurisdictionTypes };
```
moshthepitt: Why did we change from `Jurisdiction` to `JurisdictionTypes`?  I believe `JurisdictionTypes` is just a list of strings like ['administrative', 'operational']?

src/App/App.tsx
```diff
@@ -56,94 +62,134 @@ class App extends Component {
...
+                component={IrsReport}
+              />
+
+              {/* IRS Reporting Map view */}
               <ConnectedPrivateRoute
                 disableLoginProtection={DISABLE_LOGIN_PROTECTION}
                 exact={true}
-                path={`${INTERVENTION_IRS_URL}/plan/:id`}
+                path={`${ACTIVE_IRS_PLAN_URL}/:id/${MAP}/:jurisdictionId`}
                 component={IrsPlan}
```
moshthepitt: Are we planning to reuse IrsPlan for IRS Reporting?

src/configs/settings.ts
```diff
@@ -702,9 +704,140 @@ export const symbolLayerConfig = {
...
+  lockedfirst: number;
+  lockedmopup: number;
+  refusalsfirst: number;
+  refusalsmopup: number;
+  sprayeffectiveness: number;
+  structuresfound: number;
+  structuressprayed: number;
+  targetcoverage: number;
+}
+export interface IrsReportingCongif {
```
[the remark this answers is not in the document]
cKellyDesign: @kahama94 was trying to make it more specific to the context of IRS Reports, this config structure can be copied for other types of tables but this one is being specifically referenced and utilized by the IRS Reporting page.

src/App/App.tsx
cKellyDesign: @moshthepitt Nope, IRS Reporting Table page has its own container. If we open an `active` IRS plan in the IrsPlan module it should render a table for assigning teams to plan-jurisdictions. (once we build it)

src/store/ducks/jurisdictions.ts
cKellyDesign: Ah, poor naming convention on my part, I can rename but essentially `JurisdictionTypes` is being brought in from settings config file to allow `Jurisdiction`s with additional properties to be added to the same reducer.

cKellyDesign:
@moshthepitt thanks for the feedback, few questions:

> I would prefer that we use the URL to "know" which jurisdiction we are looking

1. agreed, do you think we should simply add single url param (`irs/report/<plan_id>/table/<jurisdiction_id>`) or nest them as the user drills down (`irs/report/<plan_id>/table/<country_id>/<region_id>/<district_id>`)
2. however, we don't know what the top-level jurisdiction_id is for a plan until we load the page, b/c of this we can probably just update the URL async as the page loads the data

moshthepitt:
> 1a. agreed, do you think we should simply add single url param (irs/report/<plan_id>/table/<jurisdiction_id>) or nest them as the user drills down (irs/report/<plan_id>/table/<country_id>/<region_id>/<district_id>)

Leaning towards `irs/report/<plan_id>/table/<jurisdiction_id>` - you?

> however, we don't know what the top-level jurisdiction_id is for a plan until we load the page, b/c of this we can probably just update the URL async as the page loads the data

How can I change the data to make this work easier?

cKellyDesign:
> Leaning towards irs/report/<plan_id>/table/<jurisdiction_id>

same here, that's enough to make the URLs unique :)

> How can I change the data to make this work easier?

I'm not really sure, since we would need to know the top-level jurisdiction before we click into the table view page.

moshthepitt:
> I'm not really sure, since we would need to know the top-level jurisdiction before we click into the table view page.

@cKellyDesign I'll investigate returning the top level jurisdiction(s) from the plans Superset slice

cKellyDesign:
> it seems that we are re-using the Jurisdiction reducer but with objects that have different fields

@moshthepitt yeah i was wondering how we would feel about this, the intent here is that the `jurisdictions` reducer can accept entities which are extending `Jurisdiction` so that we can build out `N` different types of `Jurisdiction` in settings for different contexts.

Right now they are compatible in the actual store, just need to add a standard way to check if a `Jurisdiction` in the store is a standard `Jurisdiction` or one of the custom ones. In other words, if a user loads IRS Plan, then `JurisdictionsById` will be populated with standard `Jurisdiction`s, when then navigating to IRS Report the page needs to evaluate what's in the store to see if needs to re-request the data.

cKellyDesign:
> I'll investigate returning the top level jurisdiction(s) from the plans Superset slice

@moshthepitt ah ok, currently the IRS Reporting plans list view is using the OpenSRP endpoint, but that can be updated if we get that top-level Jurisdiction id coming in from Superset :)

src/configs/env.ts
cKellyDesign: added

cKellyDesign:
## Dynamic Indicator Cells and Table Styling Status
Screen Shot 2019-09-11 at 12 49 15 PM

PR #438, opened 2019-09-09
Check for goal-target on plan definiton in planform

### LINKED ISSUE #437: Update Plan Form - check for goals
Currently we are assuming all plans have goals/targets and are not protecting our code for scenarios when business logic breaks down and we need to load a plan with no goals/targets without crashing. This should be a pretty simple fix.

Screen Shot 2019-09-09 at 1 55 51 PM

**Reproduction Steps**
1. Login
2. Navigate to `Manage Plans`
3. Click `IRS Namibia Training` to edit the plan

cKellyDesign:
Fixes #437

### DIFF
--- a/src/containers/forms/PlanForm/helpers.ts
+++ b/src/containers/forms/PlanForm/helpers.ts
@@ -26,6 +26,7 @@ import {
   PlanActionCodesType,
   planActivities,
   PlanActivity,
+  PlanActivityTitlesType,
   PlanDefinition,
   PlanGoal,
   PlanGoalDetail,
@@ -36,7 +37,7 @@ import {
   UseContext,
 } from '../../../configs/settings';
 import { DATE, IS, NAME, REQUIRED } from '../../../constants';
-import { FlexObject, generateNameSpacedUUID } from '../../../helpers/utils';
+import { generateNameSpacedUUID } from '../../../helpers/utils';
 import { InterventionType, PlanStatus } from '../../../store/ducks/plans';
 
 /** separate FI and IRS activities */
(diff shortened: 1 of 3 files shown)


src/containers/forms/PlanForm/helpers.ts
```diff
@@ -150,13 +150,16 @@ export function extractActivityForForm(activityObj: PlanActivity): PlanActivityF
...
+      activityObj.goal.target[0].due &&
+      activityObj.goal.target[0].due !== ''
         ? parseISO(`${activityObj.goal.target[0].due}${DEFAULT_TIME}`)
         : moment()
             .add(DEFAULT_ACTIVITY_DURATION_DAYS, 'days')
             .toDate(),
     goalPriority: activityObj.goal.priority || goalPriorities[1],
-    goalValue: activityObj.goal.target[0].detail.detailQuantity.value || 0,
+    goalValue:
+      (activityObj.goal.target && activityObj.goal.target[0].detail.detailQuantity.value) || 0,
```
moshthepitt: Could we use the default value of IRS plans from settings, instead of 0?

cKellyDesign: @moshthepitt not sure what you mean by `the default value of IRS plans from settings`, do you mean the variable `ZERO`?

moshthepitt: I mean this: https://github.com/onaio/reveal-frontend/blob/master/src/configs/settings.ts#L280  - 0 is actually invalid for that field :(

cKellyDesign: oooh i see, sorry i thought you meant the `[0]` not the `|| 0`, I will replace the later with the default value. :)

cKellyDesign:
> Could we add a test for this please?

@moshthepitt - tests added to check plan definitions without targets against `extractActivityForForm` at 40e8de2

PR #439, opened 2019-09-10
237 update lifecycle methods irs plan

cKellyDesign:
For #237 in the IRS Plan module, this is a quick PR updating the IRS plan module to use `getDerivedStateFromProps` instead of `componentWillReceiveProps`

### DIFF
--- a/src/containers/pages/InterventionPlan/IRS/plan/index.tsx
+++ b/src/containers/pages/InterventionPlan/IRS/plan/index.tsx
@@ -1117,6 +1117,8 @@ class IrsPlan extends React.Component<
   }
 
   /** getGisidaWrapperProps - GisidaWrapper prop builder building out layers and handlers for Gisida
+   * @param {IrsPlanProps | null} props the most recent props the component has received
+   * @param {IrsPlanState} state the mostly recently dereived state of the component
    * @returns {GisidaProps|null} props object for the GisidaWrapper or null
    */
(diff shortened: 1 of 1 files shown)


src/containers/pages/InterventionPlan/IRS/plan/index.tsx
```diff
@@ -156,14 +156,67 @@ interface IrsPlanState {
...
+    if (newPlan && childlessChildrenIds && country && isLoadingGeoms) {
+      const filteredJurisdictions = childlessChildrenIds.map(j => jurisdictionsById[j]);
+      const loadedJurisdictions = filteredJurisdictions.filter((j: Jurisdiction) => j.geojson);
+
+      if (loadedJurisdictions.length === filteredJurisdictions.length) {
+        const nextState: IrsPlanState = {
+          ...state,
+          isLoadingGeoms: false,
+        };
+        const gisidaWrapperProps = selfIrsPlan.getGisidaWrapperProps(props, nextState);
```
cKellyDesign: It seems that within a static function like this we can't call other methods on the class, whether public or private. I've tried calling `this.getGisidaWrapperProps` and `IrsPlan.getGisidaWrapperProps`, and using an static external variable to reference the class seems a bit hacky.

@moshthepitt - any suggested approaches or do you think this is fine?

moshthepitt:
Superseded by #977

Issue #432, opened 2019-08-29
Update Header Navigation


cKellyDesign:
The structuring of the navigation needs to be updates:

- [x] Turn off `Reporting > Historic FI` (nest it in a new `env` variable)
- [x] Add link `Reporting > IRS` 
- [x] Rename `Planning > Manage Plans` to `All Plans`
- [x] Rename `Planning > IRS` to `IRS Plans`
- [x] Add link `Planning > Add New Plan`

cc @rowo @moshthepitt

rowo:
@cKellyDesign Can you explain a bit on why the navigation changed from intervention focused to planning and reporting, or alternatively is there a log of the decision making I can look at?

moshthepitt:
@rowo this is where the change was introduced https://github.com/onaio/reveal-frontend/pull/305

Basically I felt that the previous pages (like the focus investigation list that we had and still have) are not suitable when you are doing plans because they get data from Superset - and this data goes through an ETL process before it arrives there.  So it was common for you to create a plan but only see it 15min later or something.

The new "manage plans" page in contrast pulls directly from OpenSRP which is a much easier source of data when you are creating/updating plans.

cKellyDesign:
@rowo I don't recall this being explicitly scoped, however Planning vs Reporting seems like a natural delineation to me as they are different activities / work flows / stages of the Plan Lifecycle. I suppose it does work both ways if we structure the Planning/Reporting the same under each intervention type, ideally we should get feedback from the client/users on this.

@mberg @craigappl - any input here?

moshthepitt:
Fixed by #513 and #476

PR #440, opened 2019-09-11
Update Header Navigation

FrankApiyo:
Fixes issue: #432


FrankApiyo:
The snapshot sizes have reduced because they now concentrate on ensuring that specific elements of the header have been rendered correctly instead of the whole header

src/configs/env.ts
```diff
@@ -16,6 +16,10 @@ export type ENABLE_IRS = typeof ENABLE_IRS;
 export const ENABLE_FI = process.env.REACT_APP_ENABLE_FI === 'true';
 export type ENABLE_FI = typeof ENABLE_FI;
 
+/** Do you want to enable the HISTORIC_FI? */
+export const ENABLE_HISTORIC_FI = process.env.REACT_APP_ENABLE_HISTORIC_FI === 'false';
+export type ENABLE_HISTORIC_FI = typeof ENABLE_HISTORIC_FI;
```
moshthepitt: Please add this to sample.env

moshthepitt:
Please also fix the conflict

moshthepitt:
Seems that this is both superseded and replaced by https://github.com/onaio/reveal-frontend/pull/546

Issue #417, opened 2019-08-26
IRS Update Plan View


cKellyDesign:
<img width="919" alt="Screen Shot 2019-08-26 at 5 58 57 AM" src="https://user-images.githubusercontent.com/4141527/63706515-79b6ba00-c7e4-11e9-911d-eda3d066afaa.png">

Currently when opening an IRS plan in the `UpdatePlan` page all of the selected jurisdiction IDs are being rendered (by the thousands) and needs to be optimized (which needs scoping).

Example `Home => Manage Plans => Testaru NA 1`

### Possible Solutions
1. Move the selected Jurisdiction list/information below the other form fields
2. If only one Jurisdiction is selected, show relevant information for that Jurisdiction
3. If multiple Jurisdictions are selected, show the number of jurisdictions (and relevant aggregations?)
4. Add a button to edit the Jurisdiction Selection, which navigates to `/intervention/irs/draft/<plan_id>
5. If changes are made without saving a draft before navigating away from page, prompt user to save

I don't think we can do all of these before [User Acceptance](https://github.com/onaio/reveal-frontend/milestone/6), but maybe 1, 2, and/or 3.

@moshthepitt @mberg - Thoughts?

moshthepitt:
Yeah this is a real nasty problem.  Maybe we should loop @rowo in to help think through this?

Also, note that it is closely related to https://github.com/onaio/reveal-frontend/issues/372

moshthepitt:
> There can easily be thousands of selected Jurisdictions in an IRS plan, so if we do allow users to edit IRS Jurisdiction selections in this form then we should also add a search field.

> Focus investigations should update their single selected Jurisdiction via dropdown (not done), 

I'd like to note that in the initial spec, we should **not** allow editing of jurisdictions for any plan (FI or otherwise) via the PlanForm.

moshthepitt:
> I don't exactly know what this page is (it's not in the original designed scope, right?).

@rowo This is the page that kind of morphed from your mockups here https://github.com/onaio/reveal-frontend/issues/199

cKellyDesign:
This is no longer blocked, please:
- [ ] add `overflow-y: scroll` and `max-height: 200px` around this list

Once the page pulls data from OpenSRP and has Jurisdiction Names:
- [ ] replace id strings with names

moshthepitt:
Finally closed by #443

PR #443, opened 2019-09-13
Display jurisdiction Names instead of jurisdiction ids in updatePlan 

p-netm:
part of #417 

- Adds new JurisdictionIdsToNames component that takes in a string array of ids and lazy loads the names either from store or from the the opensrp service.


src/components/JurisdictionIdToName/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+JurisdictionIdToName.defaultProps = defaultProps;
+
+export { JurisdictionIdToName };
+
+// connect component
+
+/** interface for ampState returned obj */
+interface DispatchedProps {
+  idsToNamesLookup: FlexObject<string>;
```
moshthepitt: I think it is probably possible to not use FlexObject here - we already know what we are dealing with based on line 108, no?

src/containers/pages/InterventionPlan/UpdatePlan/index.tsx
```diff
@@ -5,6 +5,9 @@ import { connect } from 'react-redux';
 import { RouteComponentProps } from 'react-router';
 import { Col, Row } from 'reactstrap';
 import { Store } from 'redux';
+import ConnectedJurisdictionIdToName, {
+  JurisdictionIdsToNames,
```
moshthepitt: I think that this should be `JurisdictionIdToName`.  If that is the case, why is a test not failing?  What test do we need to add?

src/containers/forms/PlanForm/index.tsx
```diff
@@ -74,6 +75,7 @@ export interface PlanFormProps {
   initialValues: PlanFormFields /** initial values for fields on the form */;
   jurisdictionLabel: string /** the label used for the jurisdiction selection */;
   redirectAfterAction: string /** the url to redirect to after form submission */;
+  JurisdictionIdToName?: typeof ConnectedJurisdictionIdToName | null;
```
moshthepitt: Can we add a docstring here to match all the other props?  For consistency :)

src/containers/pages/InterventionPlan/UpdatePlan/index.tsx
moshthepitt: Actually I think that this is an unused import.

src/containers/pages/InterventionPlan/UpdatePlan/tests/__snapshots__/index.test.tsx.snap
```diff
@@ -82,6 +82,7 @@ exports[`components/InterventionPlan/UpdatePlan renders plan definition list cor
 
 exports[`components/InterventionPlan/UpdatePlan renders plan definition list correctly: PlanForm 1`] = `
 Object {
+  "JurisdictionIdToName": null,
```
moshthepitt: Do we need another test where this is not null?

src/containers/pages/InterventionPlan/UpdatePlan/tests/index.test.tsx
```diff
@@ -22,6 +22,7 @@ describe('components/InterventionPlan/UpdatePlan', () => {
     const mock: any = jest.fn();
     const serviceMock: any = jest.fn(async () => []);
     return {
+      JurisdictionIdToName: null,
```
moshthepitt: Should we test when this is not null?

src/helpers/utils.tsx
```diff
@@ -36,8 +36,8 @@ import { Plan } from '../store/ducks/plans';
 import { InitialTask } from '../store/ducks/tasks';
 import { colorMaps, ColorMapsTypes } from './structureColorMaps';
 /** Interface for an object that is allowed to have any property */
-export interface FlexObject {
-  [key: string]: any;
+export interface FlexObject<T = any> {
```
moshthepitt: Nice

src/components/JurisdictionIdToName/index.tsx
```diff
@@ -0,0 +1,115 @@
+/** will be responsible for lazy loading a jurisdiction id
```
moshthepitt: This should me moved to `containers`

src/components/JurisdictionIdToName/index.tsx
p-netm: Then i think, i should remove the interface totally.
Ok, so if the types of the obj returned by `mapStateToProps` are implicitly well defined, then we dont need an interface describing the output, in this case the `DispatchedProps` interface?

p-netm:
> Do you think that instead of editing the top level jurisdiction redux module we should create a new one inside the OpenSRP directory inside ducks? One reason for this may be ease of portability i.e. we may need to move this to OpenSRP web at some point.

I dont think i have arguments against this, so i will do this.

src/containers/pages/InterventionPlan/UpdatePlan/index.tsx
p-netm: There isn't yet an integration test for JurisdictionIdToName in updatePlan. will add one.

src/containers/forms/PlanForm/index.tsx
p-netm: :+1:

src/components/JurisdictionIdToName/index.tsx
moshthepitt: I had not considered not using it completely, I just noticed that we dont need to use FlexObject because we already know the shape of what we are dealing with

src/store/ducks/jurisdictions.ts
```diff
@@ -85,10 +85,12 @@ export type ImmutableJurisdictionState = JurisdictionState &
 /** initial state */
 const initialState: ImmutableJurisdictionState = SeamlessImmutable({
   allJurisdictionIds: {} as AllJurisdictionIds,
+  geoLessJurisdictionsById: {},
```
moshthepitt: Do we still need this here?

p-netm: nope

p-netm:
Just realized a problem in this, apparently the ids are chunked correctly but the component still makes an api request for each id,

src/store/ducks/opensrp/jurisdictions/index.ts
```diff
@@ -0,0 +1,149 @@
+import { get, keyBy, values } from 'lodash';
```
cKellyDesign: @p-netm what is the reasoning for not simply using the existing `src/store/ducks/jurisdictions.ts`?

src/containers/forms/PlanForm/index.tsx
```diff
@@ -257,7 +262,16 @@ const PlanForm = (props: PlanFormProps) => {
                       <ul id="selected-jurisdiction-list">
                         {values.jurisdictions.map((jurisdiction, index) => (
```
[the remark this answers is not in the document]
p-netm: @cKellyDesign There was this piece of planform-specific code that's coupled with each of those ids,

```typescript
<fieldset>
                              <Field
                                type="hidden"
                                readOnly={true}
                                name={`jurisdictions[${index}].id`}
                                id={`jurisdictions-${index}-id`}
                                value={jurisdiction.id}
                              />
                              <Field
                                type="hidden"
                                readOnly={true}
                                name={`jurisdictions[${index}].name`}
                                id={`jurisdictions-${index}-name`}
                                value={jurisdiction.name}
                              />
</fieldset>
```

I was not sure if moving this code into JurisdictionIdToName component would have been ideal. I suppose i could move the fieldset into the `JurisdictionidToName` component. should i make this change?

src/store/ducks/opensrp/jurisdictions/index.ts
p-netm: Mosh thought this should go here, i think he mentioned something to do with portability of this ducks module to other FE apps that call to opensrp,.. i am not sure though if this was the reason, @moshthepitt, please help me answer this :point_up:

cKellyDesign: > portability of this ducks module to other FE apps that call to opensrp

Ok yeah that makes sense 👍

src/containers/forms/PlanForm/index.tsx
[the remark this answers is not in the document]
cKellyDesign: Oh interesting, maybe you could pass it in as the mapping function which is used in the `JurisdictionidToName` render function. Maybe something like:

```
<JurisdictionidToName
  id={values.jurisdictions.map(jur => jur.id)} 
  someMapperProp={(d, i) => (<li key={i}><fieldset>...</fieldset></li>)}
/>
```
```
// in JurisdictionidToName
const jurisdictionsArray = [...jurisdictions for this plan which have just been fetched]
const listItems = jurisdictionsArray.map(this.props.someMapperProp)
return <ul>{listItems}</ul>;
```

@moshthepitt what do you think about all this?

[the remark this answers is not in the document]
moshthepitt: @cKellyDesign I think that is a nice solution

moshthepitt:
We probably need to use this one https://github.com/OpenSRP/opensrp-server-web/issues/139

src/constants.tsx
```diff
@@ -367,3 +369,11 @@ export const FI_PLAN_TYPE = 'FI';
 export type FI_PLAN_TYPE = typeof FI_PLAN_TYPE;
 export const IRS_PLAN_TYPE = 'IRS';
 export type IRS_PLAN_TYPE = typeof IRS_PLAN_TYPE;
+
+//
```
cKellyDesign: Some cleanup

src/constants.tsx
```diff
@@ -367,3 +369,11 @@ export const FI_PLAN_TYPE = 'FI';
...
+
+/** dictates how many jurisdiction ids to pass as filter parameter to the
+ * findJurisdictionById endpoint.
+ */
+export const JURISDICTION_NAMES_CHUNK_SIZE = 17;
+export type JURISDICTION_NAMES_CHUNK_SIZE = typeof JURISDICTION_NAMES_CHUNK_SIZE;
```
cKellyDesign: Is this something that could/should be configured differently for different implementations? If so then should go in `env.ts`, otherwise looks good here.

src/helpers/utils.tsx
cKellyDesign: Yeah nice touch

src/constants.tsx
p-netm: Yeah , its configurable

cKellyDesign:
> It seems that our solution is failing when the number of jurisdictions is too many.

@moshthepitt it looks like this has been resolved 🎉

moshthepitt:
@cKellyDesign @p-netm I think we still need to refactor slightly to use https://github.com/OpenSRP/opensrp-server-web/issues/139

cKellyDesign:
@moshthepitt is OpenSRP/opensrp-server-web#139 ready for implementation?

moshthepitt:
@cKellyDesign I think it is okay to wait for that to be ready.  I feel that it will give us a much better way to achieve the goals of this PR without the chunked repeated requests to the server, and I feel that there is not real pressure to have this issue fixed right now so we can afford to wait until the PR on the OpenSRP server side is approved and merged.  Do you agree?

cKellyDesign:
@p-netm what's the status on this one? Do you need anything to complete?

p-netm:
> @p-netm what's the status on this one? Do you need anything to complete?




@cKellyDesign This is done, ready for rereview and merge.

src/components/forms/PlanForm/index.tsx
```diff
@@ -127,6 +126,14 @@ export interface PlanFormProps {
...
   jurisdictionLabel: string /** the label used for the jurisdiction selection */;
   redirectAfterAction: string /** the url to redirect to after form submission */;
+  renderLocationNames?: (
+    child?: (
+      locationName: string,
+      locationId: string,
+      index: number
+    ) => JSX.Element /** nested render content for each location name */
+  ) => JSX.Element;
+  /** a render prop that renders the plan's location names */
```
cKellyDesign: @p-netm why the nested functions? Is it possible to flatten to a single func that just accepts those three params?

p-netm:
@moshthepitt @cKellyDesign Ready for rereview

src/containers/pages/InterventionPlan/UpdatePlan/PlanLocationNames/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+    serve
+      .read(planId)
+      .then((response: Location[]) => {
+        store.dispatch(actionCreator(response, planId));
+      })
+      .catch((err: Error) => {
+        displayError(err);
+      });
+  }
```
cKellyDesign: @p-netm would it be possible to move this out of the component and export it as its own const? (for testability?)

src/containers/pages/InterventionPlan/UpdatePlan/PlanLocationNames/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+  useEffect(() => {
+    if (plan) {
+      const planId = plan.identifier;
+      loadLocations(serviceClass, planId, fetchLocationsAction).catch(err => displayError(err));
+    }
+  }, []);
```
cKellyDesign: Is there ever a situation where plan is not in the store when this first mounts? Is there ever a scenario where the plan might change?

Seems like no, but just trying to think of edge cases :)

src/containers/pages/InterventionPlan/UpdatePlan/index.tsx
```diff
@@ -93,6 +94,14 @@ const UpdatePlan = (props: RouteComponentProps<RouteParams> & UpdatePlanProps) =
...
     ...propsForUpdatingPlans,
     ...(plan && { initialValues: getPlanFormValues(plan) }),
+    /** a renderProp prop. this tells the planForm; I will give you a component that knows of the plan you are displaying,
+     * the component will get jurisdictions associated with that plan and render them as links, what you(planForm)
+     * will do, is provide a child prop(a renderProp) that tells the mentioned component what other stuff you would wish
+     * displayed alongside the jurisdiction names links.
+     */
+    renderLocationNames: (
+      child?: (locationName: string, locationId: string, index: number) => JSX.Element
+    ) => <ConnectedPlanLocationNames child={child} plan={plan} />,
```
cKellyDesign: Should we move this typing to it's own line so it can be exported and used everywhere it needs to be defined?

src/store/ducks/opensrp/locations/index.ts
```diff
@@ -0,0 +1,147 @@
...
+/** creates actions that remove locations assigned to a plan from store
+ * @param {string} planId -  id of plans with the above assigned locations
+ */
+export const removeLocations = (planId: string): RemoveLocationsAction => {
+  return {
+    locationsByPlanId: { [planId]: [] },
+    planId,
+    type: REMOVE_LOCATIONS,
+  };
+};
```
cKellyDesign: Nice 😎

src/containers/pages/InterventionPlan/UpdatePlan/PlanLocationNames/index.tsx
p-netm: done

src/containers/pages/InterventionPlan/UpdatePlan/PlanLocationNames/index.tsx
p-netm: no, this component is  only shown when you are editing plan, not creating it.

src/containers/pages/InterventionPlan/UpdatePlan/index.tsx
p-netm: yeap.

src/components/forms/PlanForm/index.tsx
p-netm: Am afraid the other alternative would be to get rid of renderLocationNames props entirely and use the `PlanLocationNames` component from within `PlanForm`. having a nested renderProp is a consequence of passing the component responsible for rendering the locationNames as a prop.
@cKellyDesign @moshthepitt

Issue #441, opened 2019-09-12
Update Reveal UI Colors


cKellyDesign:
This ticket tracks work needed to update the color scheme to match the marvel mockups: https://marvelapp.com/cc83f0e/screen/61277177

PR #444, opened 2019-09-13
Added fonts, updated layout to full width, updated home cards

OluwatobiMayowa:
Mockup: https://marvelapp.com/cc83f0e/screen/61277177/handoff

Closes: #441 
Closes: #445


src/App/App.tsx
```diff
@@ -53,140 +53,142 @@ import './App.css';
...
-                path="/"
-                component={Home}
-              />
-              <ConnectedPrivateRoute
-                disableLoginProtection={DISABLE_LOGIN_PROTECTION}
-                exact={true}
-                path={INTERVENTION_IRS_URL}
-                component={IrsPlans}
-              />
+        <Container>
```
cKellyDesign: Is this nested `<Container>` intentional? Does it accomplish something in particular?

src/containers/pages/Home/Home.tsx
```diff
@@ -31,7 +31,7 @@ class Home extends React.Component<{}, {}> {
               <Link to={INTERVENTION_IRS_URL} className="intervention-link">
                 <Button
                   color="outline"
-                  className="btn-intervention btn-lg btn-block btn-outline-dark"
+                  className="btn-intervention btn-lg btn-block btn-outline-green"
```
cKellyDesign: 👍

src/App/App.tsx
```diff
@@ -53,140 +53,142 @@ import './App.css';
 class App extends Component {
   public render() {
     return (
-      <Container>
+      <Container fluid={true}>
```
cKellyDesign: @mberg will be so happy, I'm glad it was this simple to go fluid!

src/App/App.tsx
OluwatobiMayowa: It's to be able to style the top nav differently from the body. Since changing to fluid means that all the body contents now went 100%. The nested is to keep them as they were.

PR #451, opened 2019-09-19
Mark IRS Plan as Complete

cKellyDesign:
This PR: 

* adds a button to the IRS reporting table page which navigates the user to the Plan Completion form page (which is used to mark FI plans as complete as well)
* allows the Plan Completion form page to work with PlanRecord types of plans
* handles plans endpoint arrays response as well as single object


moshthepitt:
I'm guessing the intention of for tests to come later... should we create an issue?

moshthepitt:
@cKellyDesign lets add tests for this one, yeah?

moshthepitt:
Merging this into the `382-IRS-reporting-map` branch.   Not sure if we'll use this long term, but want to close this open PR>

Issue #388, opened 2019-08-21
Nambia IRS Sprayed Room logic v2


mberg:
IRS for Namibia is different then the standard IRS to date where tasks are based on known structures per operational area.

With Namibia, IRS is measured in sprayed rooms not structures sprays. Namibia will provide for each operational a value of estimated rooms (based on estimated number of structures). The rooms won't be calculated against enumerated structures in the operational area.  Rhis is different then traditional IRS that uses the number of enumerated structures for that area.

Planning

They will provide a CSV that will have the operational area name and the number of estimated rooms.

We will append a target_rooms field to the [jurisdiction table](https://superset.reveal-stage.smartregister.org/superset/explore/?form_data=%7B%22datasource%22%3A%2250__table%22%2C%22viz_type%22%3A%22table%22%2C%22slice_id%22%3A471%2C%22url_params%22%3A%7B%7D%2C%22granularity_sqla%22%3Anull%2C%22time_grain_sqla%22%3A%22P1D%22%2C%22time_range%22%3A%22100+years+ago+%3A+%22%2C%22groupby%22%3A%5B%5D%2C%22metrics%22%3A%5B%5D%2C%22percent_metrics%22%3A%5B%5D%2C%22timeseries_limit_metric%22%3Anull%2C%22row_limit%22%3A10000%2C%22include_time%22%3Afalse%2C%22order_desc%22%3Atrue%2C%22all_columns%22%3A%5B%22id%22%2C%22geographic_level%22%2C%22parent_id%22%2C%22name%22%5D%2C%22order_by_cols%22%3A%5B%5D%2C%22adhoc_filters%22%3A%5B%5D%2C%22table_timestamp_format%22%3A%22%25Y-%25m-%25d+%25H%3A%25M%3A%25S%22%2C%22page_length%22%3A0%2C%22include_search%22%3Afalse%2C%22table_filter%22%3Afalse%2C%22align_pn%22%3Afalse%2C%22color_pn%22%3Atrue%7D).   This needs to be added to both the hierarchy table which the android client downloads and also the superset slice.

To set the target for the plan we will introduce a **dynamic** property to set the target goal.  This will allow us to apply a single activity in a single plan across all jurisdictions.  To do this we will set the target quantity in the plan to equal **jurisdiction.target_rooms*.

_Planning Interface_

Will show in the jurisdiction table the number of rooms.

**Mobile Client**

[...]

Issue #381, opened 2019-08-20
IRS Indicators Scoping


moshthepitt:
It seems that we need the following IRS indicators:

- structuresTargeted
- structuresFound
- structuresSprayed
- targetCoverage
- foundCoverage
- householdsNotAccessible
- refusalsFirst
- refusalsMopup
- lockedFirst
- lockedMopup

Each of these needs to be scoped i.e. in Reveal, where do you get the numbers for these indicators, and how do you calculate them?

Helpful doc: https://docs.google.com/spreadsheets/d/1-F9exjuVo6qZ6YuQW8NLKLuysjTyIb4IPKEWOQL4RjQ/edit#gid=744478914

moshthepitt:
This is part of #192

moshthepitt:
- **structuresTargeted** ='total structures targeted' as per shapefile or csv
- **structuresFound** =sum(nSprayableTotal)
- **structuresSprayed** =sum(ifnull(nSprayedTotalFirst,0)) + ifnull(nSprayedTotalMop,0))
- **targetCoverage** =structuresSprayed/structuresTargeted
- **foundCoverage** =structuresSprayed/structuresFound
- **householdsNotAccessible** =sum(case when householdAccessible = 'No' then 1 else 0) 
- **refusalsFirst** =sum(case when unsprayedReasonFirst = 'Refused' then 1 else 0) /structuresFound
- **refusalsMopup** =sum(case when unsprayedReasonFirst = 'Locked' then 1 else 0) /structuresFound
- **lockedFirst** =sum(case when unsprayedReasonMop = 'Refused' then 1 else 0) /structuresFound
- **lockedMopup** =sum(case when unsprayedReasonMop = 'Locked' then 1 else 0) /structuresFound

This is from: https://docs.google.com/spreadsheets/d/1-F9exjuVo6qZ6YuQW8NLKLuysjTyIb4IPKEWOQL4RjQ/edit#gid=744478914

moshthepitt:
For comparison, this is how the indicators were calculated in mSpray: https://github.com/onaio/mspray/blob/docs/docs/irs.md

moshthepitt:
### Differences with mSpray

- **New Structures**: this was included in mSpray calculations.  Could we confirm that it is now not required?
- **Sprayed duplicate**: this was included in mSpray calculations.  Could we confirm that it is now not required?
[...]

Issue #192, opened 2019-07-12
IRS Scoping 


armageddon:
For IRS Planning MVP, by end of July, CHAI is expecting the following:

- Ona will manually upload shapefiles (note that Akros has a copy for both Namibia and Botswana)
- CHAI to select operational areas (OAs) in Reveal interface (worst case scenario if CHAI provides OA IDs, Ona can do it)
- CHAI or MOH team create users and teams in OpenMRS interface (eventually this will be built into OpenSRP) 
- CHAI to assign team leads in Reveal interface
- For Namibia, CHAI to provide the denominator targets as a CSV and Ona will load it (rather than have people individually click through and edit each target)
- Everything that Matt showed in IRS planning mock-ups on June 27, except risk and resource estimate
@moshthepitt @mberg @cKellyDesign I am adding info from Confluence here

armageddon:
Notes on UI behavior
- By default all districts will be selected - user will have to deselect them to remove them from the plan, can also unselect OAs under the selected districts
- Once team assignment done - there would be a "Save Plan" or "Publish Plan" that would generate the tasks and assign to team members
- Teams are created in the OpenMRS interface - will need to provide clear documentation on how to do this, until the OpenSRP teams UI is ready


Namibia Denominator 
- Denominator will be a manually entered target for all sprayable rooms, which has already been determined by Namibia NMCP in May during microplanning, and will be provided by CHAI in CSV and uploaded by Ona
- Numerator should be the sum of all sprayable rooms, as collected in a field within the form submitted per structure, or from all forms submitted for a given structure, if multiple forms are submitted (e.g., 1 structure may have 5 rooms; the numerator for the structure should then be 5 – progress is measured against the # of rooms)


Other key considerations
- Data will be siloed per county - will not be able to see team members from other countries


Outstanding Questions
- Editing a plan that has already been published is complex - need to figure this out (e.g., how to remove tasks, reassign tasks)
- Another complicating factor: mop ups - different team members might be deployed on different days to the same op area - do we add new teams?
- Program might add additional teams if they realize add'l resources are needed


Post-MVP Requests:
- Hierarchies are currently districts and OAs. Note that we will eventually want this to aggregate up into regions, etc.
- Would want a mouse over or label on the map.
[...]

Issue #179, opened 2019-07-08
Investigate IRS Reporting Needs


moshthepitt:
We need to do a gap analysis of sorts to figure out what the Reveal web app may be lacking that is required for IRS reporting.

One way of doing this may be setting up a Reveal frontend instance that is set up to use mSpray data from the last IRS season (this data may need to be manipulated a little to fit the current data structures i.e, OpenSRP tasks, structures and jurisdictions).

mlberg:
This is post July 31. We'll do this in August

cKellyDesign:
Given the latest requirements this conversation won't happen until September.

moshthepitt:
@craigappl can we get accurate dates for when IRS reporting needs to be in the web app?  Once we have those dates, we need a few weeks to tackle this issue.

PR #459, opened 2019-09-25
IRS Reporting Improvements

moshthepitt:
This PR adds the following

- Support for Zambia IRS reports
- Refactors IRS reporting tables to use a simpler component that leverages more on the DrillDownTable component features
- Add new redux modules that are specifically for IRS reports
- Adds initial tests for IRS reporting work

Fix: #179
Fix: #192
Fix: #381 
Fix: #382
Fix: #388 
Closes: #435 
Closes: #436 
Closes: #430 
Closes: #454


moshthepitt:
### Next steps

The long term aim is to write code that allows us to have the same IRS reporting views (the jurisdictions table and the map view) in a way that they are re-usable for different countries/plans via configuration.

To move us towards that goal we need to:

1. Create a re-usable IRS structures reducer, similar to the re-usable IRS jurisdictions reducer introduced in https://github.com/onaio/reveal-frontend/pull/459/commits/f971901f89afa91a03a1eabb4486062bc9dba56c
2. Refactor the IRS map component to use the re-usable structures and jurisdictions reducers
3. Improve the way we create IRS configs in `settings.ts` so that we can have a way to deploy both Zambia and Namibia using one branch (based on this branch)

package.json
```diff
@@ -11,7 +11,7 @@
     "@mapbox/geojson-extent": "^0.3.2",
     "@onaio/connected-private-route": "^0.0.8",
     "@onaio/connected-reducer-registry": "^0.0.1",
-    "@onaio/drill-down-table": "^0.0.20",
+    "@onaio/drill-down-table": "^0.0.21",
```
moshthepitt: Had to upgrade this component to support our usage, as seen in the `IRSTableCell` component

src/components/IRSTableCell/index.tsx
```diff
@@ -0,0 +1,34 @@
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
```
moshthepitt: This component is responsible for rendering the table cell that holds the jurisdiction name (and that does the drilling down).

src/containers/pages/IRS/JurisdictionsReport/helpers.ts
```diff
@@ -0,0 +1,308 @@
...
+        Cell: (cell: CellInfo) => getIRSThresholdAdherenceIndicator(cell),
+        Header: '',
+        accessor: 'roomcov',
+        className: 'indicator centered',
+      },
+    ],
+  },
+];
+
+export const IRSTableColumns: { [key: string]: Column[] } = {
```
moshthepitt: I think having the column definitions in the settings module is tricky because the columns are actual code and may have logic in them and since settings.ts should have as little "dynamic code" as possible, I think having something like `IRSTableColumns` which holds all the defined table columns might be useful.  When we have a clearer idea of how to do dynamic settings, this may be generated somehow by a deployment script.  The intention is to have an object that holds the column defs.  Env vars would then be used to pick particular column defs based on the IRSTableColumns object properties.

src/containers/pages/IRS/JurisdictionsReport/helpers.ts
```diff
@@ -0,0 +1,308 @@
...
+        Header: '',
+        accessor: 'roomcov',
+        className: 'indicator centered',
+      },
+    ],
+  },
+];
+
+/** columns for Namibia IRS focus (spray) areas */
+export const ZambiaFocusAreasColumns = [
```
moshthepitt: We probably need to clean these up or make them better

src/containers/pages/IRS/JurisdictionsReport/index.tsx
```diff
@@ -0,0 +1,290 @@
...
+  fetchJurisdictions: typeof fetchGenericJurisdictions;
+  fetchPlans: typeof fetchIRSPlans;
+  hasChildren: typeof hasChildrenFunc;
+  jurisdictions: GenericJurisdiction[] | null;
+  plan: IRSPlan | null;
+  service: typeof supersetFetch;
+}
+
+/** Renders IRS Jurisdictions reports */
+const JurisdictionReport = (props: GenericJurisdictionProps & RouteComponentProps<RouteParams>) => {
```
moshthepitt: This is the new IRS reporting table component

src/containers/pages/IRS/JurisdictionsReport/index.tsx
```diff
@@ -0,0 +1,290 @@
...
+    },
+    identifierField: 'jurisdiction_id',
+    linkerField: 'jurisdiction_name',
+    minRows: 0,
+    parentIdentifierField: 'jurisdiction_parent_id',
+    resizable: true,
+    rootParentId: jurisdictionId || '',
+    shouldUseEffect: false,
+    showPagination: false,
+    useDrillDownTrProps: false,
```
moshthepitt: We want to supply our own so we dont use the default

src/containers/pages/IRS/JurisdictionsReport/index.tsx
```diff
@@ -0,0 +1,290 @@
...
+          }
+        },
+      };
+    },
+    identifierField: 'jurisdiction_id',
+    linkerField: 'jurisdiction_name',
+    minRows: 0,
+    parentIdentifierField: 'jurisdiction_parent_id',
+    resizable: true,
+    rootParentId: jurisdictionId || '',
```
moshthepitt: DrillDownTable uses this to determine what "level" of the drill down to display

src/containers/pages/IRS/JurisdictionsReport/index.tsx
```diff
@@ -0,0 +1,290 @@
...
+    }
+  }
+
+  const tableProps = {
+    CellComponent: IRSTableCell,
+    columns: columnsToUse,
+    data,
+    defaultPageSize: data.length,
+    extraCellProps: { urlPath: baseURL },
+    getTdProps: (state: any, rowInfo: any, column: any, instance: any) => {
```
moshthepitt: We want to set the clicked jurisdiction id as the jurisdictionId state variable

src/containers/pages/IRS/Map/helpers.ts
```diff
@@ -0,0 +1,221 @@
...
+  return gisidaWrapperProps;
+};
+
+/** Get breadcrumbs for a jurisdiction object
+ * This uses the jurisdiction_name_path and jurisdiction_path fields to get
+ * the breadcrumbs of the parents of a given jurisdiction object
+ * @param {GenericJurisdiction} jurisdiction - the jurisdiction in question
+ * @param {string} urlPath - the url path that we append the jurisdiction id to
+ */
+export const getJurisdictionBreadcrumbs = (
```
moshthepitt: We are using the `jurisdiction_name_path` and `jurisdiction_path` properties to get breadcrumbs.

src/containers/pages/IRS/Map/index.tsx
```diff
@@ -0,0 +1,332 @@
...
+  fetchStructures: fetchGenericStructures,
+  focusArea: null,
+  jurisdiction: null,
+  plan: null,
+  service: supersetFetch,
+  structures: null,
+};
+
+/** The IRS Reporting Map component */
+const IRSReportingMap = (props: IRSReportingMapProps & RouteComponentProps<RouteParams>) => {
```
moshthepitt: the new IRS reporting map component

src/containers/pages/IRS/plans/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+/** interface for PlanList props */
+interface PlanListProps {
+  fetchPlans: typeof fetchIRSPlans;
+  plans: IRSPlan[];
+  service: typeof supersetFetch;
+}
+
+/** Simple component that loads the new plan form and allows you to create a new plan */
+const IRSPlansList = (props: PlanListProps) => {
```
moshthepitt: The new IRS plans list component

src/helpers/ProgressBar.tsx
```diff
@@ -15,13 +24,14 @@ interface ProgressBarProps {
 const defaultProgressBarProps: Partial<ProgressBarProps> = {
   decimalPoints: 0,
   height: '10px',
+  indicatorThresholds: null,
   max: 100,
   min: 0,
 };
 
 /** Displays configurable progress bar */
 const ProgressBar = (props: ProgressBarProps) => {
```
moshthepitt: @cKellyDesign there was a considerable amount of work done on the ProgressBar component courtesy of @RayceeM  here https://github.com/onaio/js-tools/pull/95 - I am not sure whether our changes here play nicely with that, and if not how to handle them.  Ideas?

src/helpers/hierarchy.ts
```diff
@@ -0,0 +1,89 @@
+import { ChildrenByParentId, Jurisdiction } from '../store/ducks/jurisdictions';
```
moshthepitt: @cKellyDesign after refactoring the IRS reporting table component I am not sure if we need these anymore.  What are your thoughts on this?

src/store/ducks/generic/jurisdictions.ts
```diff
@@ -0,0 +1,290 @@
+import { get, keyBy, values } from 'lodash';
```
moshthepitt: This is the generic jurisdictions reducer module.

src/store/ducks/generic/jurisdictions.ts
```diff
@@ -0,0 +1,290 @@
...
+  jurisdiction_depth: number;
+  jurisdiction_id: string;
+  jurisdiction_name: string;
+  jurisdiction_name_path: string[];
+  jurisdiction_parent_id: string;
+  jurisdiction_path: string[];
+  plan_id: string;
+}
+
+export interface NamibiaJurisdiction extends GenericJurisdiction {
```
moshthepitt: Not sure if we should have this defined here.

src/store/ducks/generic/jurisdictions.ts
```diff
@@ -0,0 +1,290 @@
...
+  refusalsfirst: number;
+  refusalsmopup: number;
+  sprayeffectiveness: number;
+  structuresfound: number;
+  structuressprayed: number;
+  targetcoverage: number;
+  target_2019: number;
+}
+
+export interface ZambiaJurisdiction extends GenericJurisdiction {
```
moshthepitt: Not sure if we should have this defined here.

src/store/ducks/generic/jurisdictions.ts
```diff
@@ -0,0 +1,290 @@
...
+  sprayedstruct: number;
+  totareas: number;
+  targareas: number;
+  targstruct: number;
+  perctvisareaseffect: number;
+  spraycovtarg: number;
+  roomcov: number;
+}
+
+export interface ZambiaIRSFocusArea extends GenericJurisdiction {
```
moshthepitt: Not sure if we should have this defined here.

src/store/ducks/generic/plans.ts
```diff
@@ -0,0 +1,174 @@
+import { get, keyBy, values } from 'lodash';
```
moshthepitt: This is just an IRS plans reducer module.  It is not "generic"

src/store/ducks/generic/structures.ts
```diff
@@ -0,0 +1,210 @@
+import { Feature, Geometry } from 'geojson';
```
moshthepitt: This is the generic structures reducer module

src/store/ducks/jurisdictions.ts
```diff
@@ -1,6 +1,7 @@
 import { get, keyBy, keys, pickBy, values } from 'lodash';
```
moshthepitt: @cKellyDesign I of the opinion that we do not need any changes made to this module.  In my view it is meant to serve just plain jurisdictions with geojson.  The generic jurisdictions module can handle the rest of jurisdictions data needs.  Thoughts?

src/store/ducks/plans.ts
```diff
@@ -64,6 +64,7 @@ export interface Plan extends PlanRecord {
   jurisdiction_name_path: string[];
```
moshthepitt: @cKellyDesign I think we do not need the changes in this reducer module.

src/store/ducks/structures.ts
```diff
@@ -1,6 +1,11 @@
 import { keyBy, values } from 'lodash';
```
moshthepitt: @cKellyDesign I think we do not need the changes made here.

moshthepitt:
@cKellyDesign @p-netm 

## Major changes in this PR

- **New IRS reporting plans component "ConnectedIRSPlansList"** - no big change here apart from using ListView over ReactTable.  Has tests.
- **New IRS reporting table component "ConnectedJurisdictionReport"** - changed approach, to more deeply leverage the DrillDownTable component features in drilling down and displaying content in the cell that drills down.  Has tests.
- **New IRS reporting map component "ConnectedIRSReportingMap"** - no big change, really.  Tried to make it simpler, mainly by changing to a function-based component and moving some things to a helpers module.  Has tests.
- **helpers for ConnectedJurisdictionReport and ConnectedIRSReportingMap** - holds code that was pulled out of the original versions of these components.  It is an open question where these will eventually live.  Don't have unit tests but are tested indirectly via the tests for ConnectedIRSReportingMap and ConnectedJurisdictionReport.
- **New generic reducer modules for jurisdictions and structures**.  These work by allowing us to store slightly different objects using the same reducer.  This is very much needed if we are eventually going to support having, for example, different plans with slightly different jurisdiction objects for reporting on the same site
- **a slightly different approach to IRS configuration on settings.ts**.  I felt that the first version of our settings while being very good, was not going to scale well as we would need to add different settings variables for different countries and plans.  Right now we are back to having configs that work for one country at a time in aims of satisfying the [YAGNI principle](https://deviq.com/yagni/) i.e. I dont think that we should double-down on any one path until we have a clearer idea of how settings are going to work.  This should be clearer via https://github.com/onaio/reveal-frontend/issues/464 and #156

moshthepitt:
@cKellyDesign as noted above, I added new components and moved things around, but I have not removed any of the things that were there before.  Would like you to take a look so that we can come up with a plan for that.

You'll need these env vars:

```sh
REACT_APP_SUPERSET_IRS_REPORTING_JURISDICTIONS_DATA_SLICES=556,557
REACT_APP_SUPERSET_IRS_REPORTING_JURISDICTIONS_FOCUS_AREA_LEVEL=2
REACT_APP_SUPERSET_IRS_REPORTING_STRUCTURES_DATA_SLICE=551
REACT_APP_SUPERSET_IRS_REPORTING_PLANS_SLICE=555
REACT_APP_SUPERSET_IRS_REPORTING_JURISDICTIONS_COLUMNS=zambiaJurisdictions2
```

.env.sample
```diff
@@ -14,6 +14,12 @@ REACT_APP_SUPERSET_PLANS_TABLE_SLICE=0
 REACT_APP_SUPERSET_GOALS_SLICE=0
 REACT_APP_SUPERSET_JURISDICTIONS_SLICE=0
 REACT_APP_SUPERSET_JURISDICTIONS_DATA_SLICE=0
+REACT_APP_SUPERSET_IRS_REPORTING_PLANS_SLICE=0
+REACT_APP_SUPERSET_IRS_REPORTING_JURISDICTIONS_DATA_SLICES=0,1,2,3
```
moshthepitt: With this we are able to load jurisdictions from as many slices as we want.  The last one in the list should be the lowest level (i.e. focus area level)

src/helpers/ProgressBar.tsx
cKellyDesign: @moshthepitt looks like once onaio/js-tools#95 is merged we should install/use it and update the IRS Report Map to use it correctly by passing in `lineColorThresholds` as specified in the package.

src/store/ducks/generic/plans.ts
cKellyDesign: Are the new jurisdiction reducer module "generic"? Was there a particular reason behind naming the directory `generic`?

src/store/ducks/jurisdictions.ts
cKellyDesign: Agreed, if we are specifying secondary reducer modules allowing us to store different types of jurisdictions/plans/etc then we should remove the customization changes we made to the existing reducer modules.

src/store/ducks/generic/plans.ts
moshthepitt: @cKellyDesign actually none whatsoever, I had named it "IRS" at first but then that didnt seem appropriate

.env.sample
cKellyDesign: Cool, does this assume that there is one jurisdiction level per slice? Or are we able to contain multiple levels of hierarchy per slice?

In general this makes sense if / because we have different requirements at different hierarchical levels.

src/containers/pages/IRS/JurisdictionsReport/helpers.ts
cKellyDesign: > settings.ts should have as little "dynamic code" as possible, I think having something like `IRSTableColumns`...

This makes a lot of sense, I am wondering if we will eventually want to abstract it one step further to be something like `ReportTableColumns` to cover different intervention types, but that's probably not yet necessary.

.env.sample
moshthepitt: In Namibia we have all the levels in one slice.  While in Zambia we have two slices: one with all the levels apart from the lowest level; and one with just the lowest level.  This should work in both cases

moshthepitt: I guess an underlying assumption is that we'll usually have two slices as in Zambia's case

src/containers/pages/IRS/JurisdictionsReport/helpers.ts
moshthepitt: > This makes a lot of sense, I am wondering if we will eventually want to abstract it one step further to be something like ReportTableColumns to cover different intervention types, but that's probably not yet necessary.

Yes!  I almost did this but then pulled back so that we focus on IRS for now.  But I think we should do this for the existing FI tables ASAP.

moshthepitt:
> You'll need these env vars:

@cKellyDesign sorry these are the actual env vars that you need:

```sh
REACT_APP_SUPERSET_IRS_REPORTING_JURISDICTIONS_DATA_SLICES=556,557
REACT_APP_SUPERSET_IRS_REPORTING_JURISDICTIONS_FOCUS_AREA_LEVEL=2
REACT_APP_SUPERSET_IRS_REPORTING_STRUCTURES_DATA_SLICE=551
REACT_APP_SUPERSET_IRS_REPORTING_PLANS_SLICE=555
REACT_APP_SUPERSET_IRS_REPORTING_JURISDICTIONS_COLUMNS=zambiaJurisdictions2019
REACT_APP_SUPERSET_IRS_REPORTING_FOCUS_AREAS_COLUMNS=zambiaFocusArea2019
```

src/configs/settings.ts
```diff
@@ -1,6 +1,25 @@
-/** This is the main configuration module */
+/** This is the main configuration module
```
moshthepitt: @cKellyDesign curious if you think this docstring has good guidelines for the settings module

cKellyDesign: @moshthepitt I think these are great rules to instate, for me it's a bit of a paradigm shift in moving dynamic functionalities (functions and other imports) out of this file and into helpers in more relevant directories.

Should we also allow Typings to be imported? Should Typings ever be defined here?

moshthepitt: @cKellyDesign I dont know what to do about typings :( - maybe we should just leave them here until we are clearer on what to do about them?  Of course any type defs that naturally belong in other places should be defined in those places.

src/containers/pages/IRS/Map/index.tsx
cKellyDesign: @moshthepitt this is gorgeous, good work man

src/helpers/hierarchy.ts
cKellyDesign: Agreed, this is now unused and the first iterations of these are contained in the IRS Jurisdiction Selection `Plan` page, which needs to be refactored, at which point we will migrate such functions into helpers.

src/store/ducks/generic/plans.ts
cKellyDesign: @moshthepitt how about we change `generic` to `custom`? This should type of vernacular should be considered in #464.

.env.sample
cKellyDesign: > I guess an underlying assumption is that we'll usually have two slices as in Zambia's case

@moshthepitt this is a very insightful distinction to make, at the very least the multiple slices delineate between aggregated and non-aggregated data.

Another assumption this implementation has is that a single slice may contain multiple hierarchical levels and it contain a single level, and that we deduce the complete location tree the same way in both cases. (which is awesome)

src/store/ducks/tasks.ts
```diff
@@ -210,6 +210,21 @@ export function getTasksByPlanId(state: Partial<Store>, planId: string): Task[]
   return values((state as any)[reducerName].tasksById).filter((e: Task) => e.plan_id === planId);
 }
 
+/** get tasks filtered by plan and jurisdiction ids
+ * @param {Partial<Store>} state - the redux store
+ * @param {string} planId - the plan id
+ * @param {string} jurisdictionId - the jurisdiction id
+ * @returns {Task[]} an array of tasks
+ */
+export function getTasksByPlanJurisdictionIds(
```
cKellyDesign: Nice!

src/store/ducks/generic/plans.ts
moshthepitt: @cKellyDesign agreed

src/helpers/ProgressBar.tsx
cKellyDesign: Issue created at https://github.com/onaio/reveal-frontend/issues/467

src/store/ducks/generic/jurisdictions.ts
cKellyDesign: I think we either define them here as custom versions of the `GenericJurisdiction` interface, or import them from helper or typing files.

src/configs/settings.ts
cKellyDesign: @moshthepitt What if we added `types.ts` files adjacent to `helpers.ts` files?

src/store/ducks/generic/plans.ts
cKellyDesign: Change of heart here, it seems the generic reducer/typings/methods should remain `generic`, while we shift the custom typings out of the generic code (and ideally dynamically generate them from static configs in `settings.ts`).

I don't think this needs to happen in this PR, issue created at https://github.com/onaio/reveal-frontend/issues/468

cKellyDesign:
> a slightly different approach to IRS configuration on settings.ts

@moshthepitt - Yeah I definitely agree with your approach, the last of the first iteration is still coupled to the ProgressBar which will be replaced by https://github.com/onaio/reveal-frontend/issues/467.

I think this PR is ready to be re-reviewed and merged.

src/configs/settings.ts
moshthepitt: > @moshthepitt What if we added types.ts files adjacent to helpers.ts files?

I think we should do this @cKellyDesign - I created an issue for this https://github.com/onaio/reveal-frontend/issues/469

.env.sample
pld: Is this explained elsewhere? If not, can we add a comment line above this constant that makes the importance of the order of the slice list clear?

src/containers/pages/IRS/JurisdictionsReport/helpers.ts
```diff
@@ -0,0 +1,308 @@
...
+): FlexObject[] {
+  if (item.jurisdiction_parent_id === null && !result.includes(item)) {
+    result.push(item);
+  } else {
+    const parentItems = tree.filter(el => el.jurisdiction_id === item.jurisdiction_parent_id);
+    if (parentItems.length > 0) {
+      if (!result.includes(parentItems[0])) {
+        result.push(parentItems[0]);
+      }
+      return getTree(tree, parentItems[0], result);
```
pld: this is good for now, and this function looks tail recursive which is good, but if the javascript interpreter isn't supporting tail call optimization, for large hierarchies this could cause memory issues by filling the stack as it's doing the recursion so we may want to rewrite as explicitly iterative in the future to avoid this

src/containers/pages/IRS/JurisdictionsReport/index.tsx
```diff
@@ -0,0 +1,290 @@
...
+  hasChildren: typeof hasChildrenFunc;
+  jurisdictions: GenericJurisdiction[] | null;
+  plan: IRSPlan | null;
+  service: typeof supersetFetch;
+}
+
+/** Renders IRS Jurisdictions reports */
+const JurisdictionReport = (props: GenericJurisdictionProps & RouteComponentProps<RouteParams>) => {
+  const [jurisdictionId, setJurisdictionId] = useState<string | null>(
+    props.match && props.match.params && props.match.params.jurisdictionId
```
pld: instead of the ternary, you could wrap the multiple conjunctive clause in parentheses and after closing paren suffix with `|| null`

src/containers/pages/IRS/Map/index.tsx
```diff
@@ -0,0 +1,336 @@
...
+          { comparator: planId, operator: '==', subject: 'plan_id' },
+        ]);
+      }
+
+      // get the plan
+      await service(SUPERSET_IRS_REPORTING_PLANS_SLICE, fetchPlansParams).then(
+        (result: IRSPlan[]) => fetchPlans(result)
+      );
+    } catch (e) {
+      // do something with the error?
```
pld: what about creating an issue for this? notifying the user or retry seem like things we could do, but issue can just refer to this and say we need to discuss, decide, then implement

pld: this can also be a `TODO: clean up in [pasted link to issue]`

src/containers/pages/IRS/plans/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+  /** async function to load the data */
+  async function loadData() {
+    try {
+      setLoading(plans.length < 1); // only set loading when there are no plans
+      await service(SUPERSET_IRS_REPORTING_PLANS_SLICE).then((result: IRSPlan[]) =>
+        fetchPlans(result)
+      );
+    } catch (e) {
+      // do something with the error?
```
pld: how about that issue I mentioned creating above is a more generic issue that links to this line of the code to, and here we have a todo referring to the same issue

src/containers/pages/IRS/Map/index.tsx
cKellyDesign: Ultimately I think this requires a few design decisions on how we handle errors in general, a while back we had talked about Growl-like notifications (which would include but not be limited to error handling), but this needs scoping.

https://github.com/onaio/reveal-frontend/issues/300

src/helpers/ProgressBar.tsx
```diff
@@ -31,19 +41,23 @@ const ProgressBar = (props: ProgressBarProps) => {
   const decimalValue = value / range;
   const percentValue = decimalValue * 100;
   const percentValueString = percentValue.toFixed(decimalPoints);
+
+  const backgroundColor = indicatorThresholds
+    ? getThresholdColor({ value } as CellInfo, indicatorThresholds, true)
+    : decimalValue >= GREEN_THRESHOLD
```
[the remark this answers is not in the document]
cKellyDesign: I think this will be handled when we replace this component with the shared package from js-tools, via #467

cKellyDesign:
@pld responding to this comment: https://github.com/onaio/reveal-frontend/pull/459#discussion_r330238940

Agreed, we've been discussing how best to do this #464, #471, #469, #468

PR #461, opened 2019-09-27
Issue # progress indicators visualizations

p-netm:
adds a marker on the progress bar that shows the target value.
modifies the goal reporting to be completed tasks of total_tasks

progressIndicators


p-netm:
withData

p-netm:
> modifies the goal reporting to be completed tasks of total_tasks

@moshthepitt @cKellyDesign Is this still the correct logic?, In master we have completed_tasks of target_value which would then mean we do not really need the marker, right?

moshthepitt:
@p-netm lets shelve this until @samkanga and @craigappl define the requirement better.

moshthepitt:
@p-netm closing this as it doesn't seem like we'll be able to merge it any time soon.

PR #463, opened 2019-09-29
Team assignment create edit organization

### LINKED ISSUE #409: Teams Assignment - Add new Team view
__overview__
- page with form to add new teams

### LINKED ISSUE #410: Team Management - update Team view
__Overview__
Add a view to modify Team data.

p-netm:
Breaking down And cleaning TeamAssignment modules.
Creating And editing an organization
- Adds form to enable the above,
- Adds view that implements the form.

closes #409, closes #410


src/containers/forms/OrganizationForm/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+  SAVING,
+  TEAM_LIST_URL,
+} from '../../../constants';
+import { generateNameSpacedUUID } from '../../../helpers/utils';
+import { OpenSRPService } from '../../../services/opensrp';
+
+const OrgFormNameSpace = '9a4c8cb0-df70-11e9-b38b-57f114a50538';
+
+/** default type value for organizations */
+const defaultOrganizationType = {
```
pld: not if it should be move or not, but have we considered that, is there another file to centralize this type of constant in? If we don't think it should move, that's fine, just wondering

src/containers/forms/OrganizationForm/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+                disabled={disabledFields.includes('name')}
+                className={errors.name ? `form-control is-invalid` : `form-control`}
+              />
+              <ErrorMessage
+                name="name"
+                component="small"
+                className="form-text text-danger name-error"
+              />
+            </FormGroup>
+            <FormGroup>
```
moshthepitt: Is there a reason why you are using radio buttons instead of one checkbox?

src/containers/forms/OrganizationForm/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+  identifier: Yup.string(),
+  name: Yup.string().required(REQUIRED),
+});
+
+/** interface for data fields for team's form */
+interface OrganizationFormFields {
+  active: boolean;
+  identifier: string;
+  name: string;
+  type?: any;
```
moshthepitt: It seems like we know what kind of object goes here, no?  If so let us not use `any` as the type

p-netm: will cahnge this

src/containers/forms/OrganizationForm/index.tsx
p-netm: Moved this to the organization Ducks module, seems like it more naturally belongs there over a constants file.

src/containers/forms/OrganizationForm/index.tsx
p-netm: Seemed like a good idea at the time

src/containers/forms/OrganizationForm/index.tsx
```diff
@@ -0,0 +1,195 @@
...
+          if (editMode) {
+            // 2 calls for each for updating team information and updating practitioner_role table
+            organizationService
+              .update(values)
+              .then(() => {
+                setSubmitting(false);
+                setIfDoneHere(true);
+              })
+              .catch((e: Error) => {
+                setGlobalError(e.message);
```
cKellyDesign: 🎉

src/containers/forms/OrganizationForm/index.tsx
```diff
@@ -0,0 +1,195 @@
...
+          const organizationService = new OpenSRPService(OPENSRP_ORGANIZATION_ENDPOINT);
+          setFieldValue(
+            'identifier',
+            generateNameSpacedUUID(`${moment().toString()}`, OrgFormNameSpace)
+          );
+
+          if (editMode) {
+            // 2 calls for each for updating team information and updating practitioner_role table
+            organizationService
+              .update(values)
```
cKellyDesign: I know these are the values being generated from the form, but it's not clear to me what Type of data this is, maybe  we could add a typing here if it's easy?

src/containers/forms/OrganizationForm/index.tsx
cKellyDesign: Could / should there be other status we would want to demarcate here? (eg `Pending`, `Closed`, etc?)

If not then let's make this a checkbox.

src/containers/forms/OrganizationForm/index.tsx
```diff
@@ -0,0 +1,195 @@
...
+      <Formik
+        initialValues={initialValues}
+        validationSchema={OrgSchema}
+        // tslint:disable-next-line: jsx-no-lambda
+        onSubmit={(values, { setSubmitting, setFieldValue }) => {
+          const organizationService = new OpenSRPService(OPENSRP_ORGANIZATION_ENDPOINT);
+          setFieldValue(
+            'identifier',
+            generateNameSpacedUUID(`${moment().toString()}`, OrgFormNameSpace)
+          );
```
cKellyDesign: @p-netm does this happen synchronously? When submitting I'm getting an error from OpenSRP because the `identifier` was not provided (it's an empty string in the request).

Is it better to spread the values and set the identifier manually here instead?

cKellyDesign: This is resolved by 7a1635a

cKellyDesign:
@moshthepitt this PR is ready for re-review, tests are passing and organizations are saving.

src/containers/forms/OrganizationForm/index.tsx
cKellyDesign: I don't think this should be a blocker to merging.

pld:
merging with the understanding that @moshthepitt may request changes that don't get into this PR, those can go into a new issue

src/containers/forms/OrganizationForm/index.tsx
```diff
@@ -0,0 +1,197 @@
...
+  identifier: Yup.string(),
+  name: Yup.string().required(REQUIRED),
+});
+
+/** interface for data fields for team's form */
+export interface OrganizationFormFields {
+  active: boolean;
+  identifier: string;
+  name: string;
+  type: typeof defaultOrganizationType;
```
moshthepitt: @p-netm can we create an issue to create an interface for this?  It is not usual, and imho not recommended, to use `typeof` with an object that you already have, and thus should be able to create an interface for.  it also sets us up for more DRY code in the future when we have other objects of this type/shape.

src/containers/forms/OrganizationForm/index.tsx
moshthepitt: @p-netm @cKellyDesign can we create an issue to at least discuss this?  If we can use a checkbox idk what valid reasons we would have not to.

Issue #482, opened 2019-10-02
Namibia target coverage bug on map


moshthepitt:
![bugz](https://user-images.githubusercontent.com/372073/66056091-41bb3900-e53f-11e9-8432-4f00d72c1a87.png)

moshthepitt:
The progress bar is also not colored like it should be

cKellyDesign:
@moshthepitt in the screenshot it looks like the denominator is zero for the first row so, it's unclear what the value actually should be

moshthepitt:
@cKellyDesign I think the thing is looking for a wrong value hence getting zero

cKellyDesign:
This is the endpoint being called and you're right, `structurestargeted` is not one of the properties: https://superset.reveal-stage.smartregister.org/superset/slice_json/554?form_data={"adhoc_filters":[{"clause":"WHERE","expressionType":"SIMPLE","comparator":"9f9210ef-9ad3-4cb6-953e-2c9466febeac","operator":"==","subject":"jurisdiction_id"},{"clause":"WHERE","expressionType":"SIMPLE","comparator":"6807a4a9-52bf-4d0d-b4d0-50fdd86590f8","operator":"==","subject":"plan_id"}],"row_limit":1}

moshthepitt:
oh goodness, it should be `target_2019`

PR #484, opened 2019-10-02
[Bug Fix] IRS Map Indicator Row values

cKellyDesign:
Fix: #482 
Also adds Jurisdiction name to page title

### DIFF
--- a/src/containers/pages/IRS/Map/helpers.ts
+++ b/src/containers/pages/IRS/Map/helpers.ts
@@ -85,6 +85,13 @@ export const IRSIndicatorRows: { [key: string]: IndicatorRows } = {
       title: 'Spray Success Rate (PMI SC)',
       value: 'spraysuccess',
     },
+    {
+      denominator: 'sprayed_rooms_eligible',
+      description: 'Percent of rooms sprayed over eligible',
+      numerator: 'sprayed_rooms_sprayed',
+      title: 'Spray Coverage (Rooms)',
+      value: 'roomcov',
+    },
   ],
 };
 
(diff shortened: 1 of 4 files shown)


src/containers/pages/IRS/Map/helpers.ts
```diff
@@ -31,13 +31,16 @@ export const IRSIndicatorStops: { [key: string]: string[][] } = {
...
+export type IndicatorPropIngestor = (obj: any) => string | number;
+
 /** interface to describe and indicator row item */
 export interface IndicatorRowItem {
   denominator: string | number;
   description: string;
   numerator: string | number;
   title: string;
-  value: string | number;
+  value: string | number | IndicatorPropIngestor;
```
cKellyDesign: @moshthepitt how do you feel about this approach of adding custom ingestion methods? Alternatively we can simplify it and just check if `focusArea[row.value] < 1` to handle this use case.

moshthepitt: @cKellyDesign I like this

cKellyDesign:
Oh good point, though I would only expect the snapshot only to fail if the indicator row fixture changed from a key-string to a ingest-method. Will add test.

src/containers/pages/IRS/Map/helpers.ts
cKellyDesign: @moshthepitt I actually removed this for the sake of simplicity, i think it's safe to assume that:

1. All progress bar `value`s are represented as percentages
2. Values which are floats need to be multiplied by 100 and rounded
3. Values which are integers should be used as is
4. We never want to render decimals in the indicator row

Do you think these are safe assumptions? If so then we don't need to add the same ingestion method (with different strings) to all the `value` properties.

moshthepitt: @cKellyDesign I think these sound sound :)

Issue #491, opened 2019-10-03
IRS reporting table view breaks when there is no data from Superset


moshthepitt:
As in the site completely breaks with no recourse apart from a refresh.  We should handle this more gracefully.  Currently happening in Zambia production.

cKellyDesign:
My suspicion is that our actionCreators that we're calling should be wrapped in a conditional to only save results to stores when there are results to save.

https://github.com/onaio/reveal-frontend/blob/master/src/containers/pages/IRS/JurisdictionsReport/index.tsx#L83

moshthepitt:
@cKellyDesign I think we should do that regardless, but I feel that it is not the core problem - whether something goes into the Redux store or not shouldn't matter - ama?

moshthepitt:
![ummm](https://user-images.githubusercontent.com/372073/66141674-bc9e5580-e60c-11e9-899c-b9c922ffcf3d.png)

So Zambia is currently doing this

moshthepitt:
@cKellyDesign ^^

cKellyDesign:
@moshthepitt Is there a level 0 Jurisdiction for `Zambia`?

moshthepitt:
yes there is

moshthepitt:
right now Zambia loads actually (coz we got some data synced) @cKellyDesign

cKellyDesign:
@moshthepitt looks like it's working now in production
<img width="1186" alt="Screen Shot 2019-10-03 at 9 01 43 AM" src="https://user-images.githubusercontent.com/4141527/66143764-8b963480-e5bc-11e9-93a6-4ac2e21bb34f.png">

moshthepitt:
yeah we got data in.  The screenshot above yours happens when there is no data for jurisdictions

moshthepitt:
@cKellyDesign also, currently the map page is not showing on my end in Zambia

cKellyDesign:
> currently the map page is not showing on my end in Zambia

@moshthepitt Seems there are no Jurisdicitons in the superset slice with this planId: https://superset.reveal-zm.smartregister.org/superset/slice_json/160?form_data={%22adhoc_filters%22:[{%22clause%22:%22WHERE%22,%22expressionType%22:%22SIMPLE%22,%22comparator%22:%229f1e0cfa-5313-49ff-af2c-f7dbf4fbdb9d%22,%22operator%22:%22==%22,%22subject%22:%22plan_id%22}],%22row_limit%22:1}

[...]

Issue #350, opened 2019-08-15
Handle superset requests which return no records


cKellyDesign:
Currently if a query to a superset slice returns no data the request errors out with `400 - Bad Request` and should be handled correctly.

moshthepitt:
Seems this is fixed.  Please re-open if not.

PR #492, opened 2019-10-03
[Bugfix] Only save jurisdictions to store if the slice returns data

cKellyDesign:
For #491

### DIFF
--- a/src/containers/pages/IRS/JurisdictionsReport/index.tsx
+++ b/src/containers/pages/IRS/JurisdictionsReport/index.tsx
@@ -92,9 +92,11 @@ const JurisdictionReport = (props: GenericJurisdictionProps & RouteComponentProp
             { jurisdiction_depth: true }
           );
         }
-        await service(slice, fetchJurisdictionsParams).then((result: GenericJurisdiction[]) =>
-          fetchJurisdictions(slice, result)
-        );
+        await service(slice, fetchJurisdictionsParams).then((result: GenericJurisdiction[]) => {
+          if (result) {
+            return fetchJurisdictions(slice, result);
+          }
+        });
       });
     } catch (e) {
       // todo - handle error https://github.com/onaio/reveal-frontend/issues/300
--- a/src/containers/pages/IRS/JurisdictionsReport/index.tsx
+++ b/src/containers/pages/IRS/JurisdictionsReport/index.tsx
@@ -92,11 +92,9 @@ const JurisdictionReport = (props: GenericJurisdictionProps & RouteComponentProp
             { jurisdiction_depth: true }
           );
         }
-        await service(slice, fetchJurisdictionsParams).then((result: GenericJurisdiction[]) => {
-          if (result) {
-            return fetchJurisdictions(slice, result);
-          }
-        });
+        await service(slice, fetchJurisdictionsParams).then((result: GenericJurisdiction[]) =>
+          fetchJurisdictions(slice, result)
+        );
       });
     } catch (e) {
       // todo - handle error https://github.com/onaio/reveal-frontend/issues/300
--- a/src/containers/pages/IRS/JurisdictionsReport/index.tsx
+++ b/src/containers/pages/IRS/JurisdictionsReport/index.tsx
@@ -183,8 +183,8 @@ const JurisdictionReport = (props: GenericJurisdictionProps & RouteComponentProp
   }
 
   const tableProps = {
-    ...(columnsToUse && { columns: columnsToUse }),
     CellComponent: IRSTableCell,
+    columns: [...(columnsToUse || [])],
     data,
     defaultPageSize: data.length,
     extraCellProps: { urlPath: baseURL },


src/containers/pages/IRS/JurisdictionsReport/index.tsx
```diff
@@ -92,9 +92,11 @@ const JurisdictionReport = (props: GenericJurisdictionProps & RouteComponentProp
             { jurisdiction_depth: true }
           );
         }
-        await service(slice, fetchJurisdictionsParams).then((result: GenericJurisdiction[]) =>
-          fetchJurisdictions(slice, result)
-        );
+        await service(slice, fetchJurisdictionsParams).then((result: GenericJurisdiction[]) => {
+          if (result) {
+            return fetchJurisdictions(slice, result);
```
pld: what about making this check the responsibility of `fetchJurisdictions`?

cKellyDesign:
I am not able to reproduce the issue locally, but it seems the `(400) Bad Request` error from Superset slice API is not getting caught in the `try catch` block when compiled in production.

The service this module uses returns `undefined` when there is no data being returned so wrapping the redux action creator call in a conditional is currently my best guess.

src/containers/pages/IRS/JurisdictionsReport/index.tsx
cKellyDesign: Oh that's a good point, then it would protect it everywhere it's being called 👍

pld: cool, and future extension could be having it return an object on error than caller knows how to render for user... but not now :)

cKellyDesign: Actually, if the results are being passed in as `undefined`, then [this line](https://github.com/onaio/reveal-frontend/blob/master/src/store/ducks/generic/jurisdictions.ts#L113) should actually be preventing `undefined` [from being iterated upon](https://github.com/onaio/reveal-frontend/blob/master/src/store/ducks/generic/jurisdictions.ts#L116).

moshthepitt:
> I am not able to reproduce the issue locally, but it seems the (400) Bad Request error from Superset slice API is not getting caught in the try catch block when compiled in production.

@cKellyDesign so one hacky way to reproduce is to point to a slice with filters set up such that you dont get any data back

cKellyDesign:
@moshthepitt yeah that was what I was trying to do, but it seems to not tripping up in the same way when we get the same server response / no data. Is there a way we can easily compile prod code and run it locally?

moshthepitt:
I guess we could run yarn build locally and then load the web app from the build folder?

moshthepitt:
@cKellyDesign the stacktrace on the Zambia site seems to indicate that it breaks here https://github.com/onaio/js-tools/blob/master/packages/DrillDownTable/src/WithHeaders/index.tsx#L12

moshthepitt:
not sure how it is getting there

moshthepitt:
So the Zambia settings are currently like so:

```sh
REACT_APP_SUPERSET_IRS_REPORTING_JURISDICTIONS_COLUMNS: zambia2019
REACT_APP_SUPERSET_IRS_REPORTING_FOCUS_AREAS_COLUMNS: zambia2019
```

Instead of (going to redeploy asap)

```sh
  REACT_APP_SUPERSET_IRS_REPORTING_JURISDICTIONS_COLUMNS: zambiaJurisdictions2019
  REACT_APP_SUPERSET_IRS_REPORTING_FOCUS_AREAS_COLUMNS: zambiaFocusArea2019
```

When it cant find the columns it goes into `withHeaders` in the link above, only this time it goes in with no data.  Maybe we need an issue on DrillDownTable to not process non-objects in withHeaders

cKellyDesign:
Ah, yeah not clear how it's getting there, but this line will break if `item` is `undefined` https://github.com/onaio/js-tools/blob/master/packages/DrillDownTable/src/helpers/utils.ts#L8

Should be `Object.keys(Object(item))` instead.

cKellyDesign:
Oh so the error is actually happening because the other Jurisdiciton api calls are passing and saving, but then hitting snag later trying to build tables off of configs that are being incorrectly referenced? (will test this momentarily)

src/containers/pages/IRS/JurisdictionsReport/index.tsx
```diff
@@ -183,8 +183,8 @@ const JurisdictionReport = (props: GenericJurisdictionProps & RouteComponentProp
   }
 
   const tableProps = {
-    ...(columnsToUse && { columns: columnsToUse }),
     CellComponent: IRSTableCell,
+    columns: [...(columnsToUse || [])],
```
moshthepitt: Can we have a test where this happens and we check that the resulting columns have actually been generated as expected by the drill down component?  I am not sure that that will happen, but it might

cKellyDesign: Am trying to get a test to crash with the old code now, so we can test this solution

cKellyDesign: Seems like we would need to move this to the helpers file so we can test the actual props object instead of mounting everything and attempting to get the props off the mocked component.

cKellyDesign:
Closing this PR because it's no longer an emergency and while we need to handle this scenario of no-data from slices, that's part of a broader discussion in #350

PR #500, opened 2019-10-07
Fix parent jurisdiction id filter

moshthepitt:
PR ensures that filtering plans by parent jurisdiction id works.

### DIFF
--- a/src/store/ducks/plans.ts
+++ b/src/store/ducks/plans.ts
@@ -425,13 +425,13 @@ export const fetchPlanRecords = (planList: PlanRecordResponse[] = []): FetchPlan
 /** getPlansById - get plansById by intervention type
  * @param {Partial<Store>} state - the redux store
  * @param {InterventionType} intervention - the intervention type
- * @param {string[]} status - the plan status
+ * @param {string[]} statusList - the plan statuses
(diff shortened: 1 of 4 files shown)


src/store/ducks/plans.ts
```diff
@@ -425,64 +425,67 @@ export const fetchPlanRecords = (planList: PlanRecordResponse[] = []): FetchPlan
...
 ): Plan[] {
   return values((state as any)[reducerName].plansById).filter(
     (plan: Plan) =>
       plan.plan_intervention_type === intervention &&
-      (status.length ? status.includes(plan.plan_status) : true) &&
+      (statusList.length ? statusList.includes(plan.plan_status) : true) &&
       (reason ? plan.plan_fi_reason === reason : true) &&
       (jurisdictions.length ? jurisdictions.includes(plan.jurisdiction_id) : true) &&
-      (parentJurisdictionId ? plan.jurisdiction_path.includes(parentJurisdictionId) : true)
+      (parentJurisdictionId && !plan.jurisdiction_path ? false : true) &&
```
moshthepitt: This is the main fix in the PR.  We are ensuring that we take into account the possibility of `plan.jurisdiction_path` being null

Issue #501, opened 2019-10-07
Set IRS Reporting Map to Full Page Width


cKellyDesign:
As per requested by @mberg at https://onaio.slack.com/archives/CBSU8L2SC/p1570456351058900

cKellyDesign:
It looks like it will actually be easier to make the whole site full-width by simply removing the nested `Container` in `App.tsx` to allow all pages to have a fluid layout.

<img width="1402" alt="Screen Shot 2019-10-07 at 7 52 40 AM" src="https://user-images.githubusercontent.com/4141527/66323177-3d44a680-e8d8-11e9-8fd9-2f14f0edf788.png">
<img width="1409" alt="Screen Shot 2019-10-07 at 7 35 10 AM" src="https://user-images.githubusercontent.com/4141527/66322389-e7bbca00-e8d6-11e9-8d78-33606ee98129.png">
<img width="1399" alt="Screen Shot 2019-10-07 at 7 43 10 AM" src="https://user-images.githubusercontent.com/4141527/66322392-e7bbca00-e8d6-11e9-9f51-86355ebc662b.png">
<img width="1395" alt="Screen Shot 2019-10-07 at 7 43 28 AM" src="https://user-images.githubusercontent.com/4141527/66322393-e7bbca00-e8d6-11e9-9b70-a40addd2cb8b.png">
<img width="1397" alt="Screen Shot 2019-10-07 at 7 43 44 AM" src="https://user-images.githubusercontent.com/4141527/66322394-e8546080-e8d6-11e9-9ad7-77e9db74e67e.png">
<img width="1396" alt="Screen Shot 2019-10-07 at 7 43 57 AM" src="https://user-images.githubusercontent.com/4141527/66322395-e8546080-e8d6-11e9-9e0a-5d1e849fc2de.png">
<img width="1398" alt="Screen Shot 2019-10-07 at 7 44 21 AM" src="https://user-images.githubusercontent.com/4141527/66322396-e8546080-e8d6-11e9-9c53-54e812888eee.png">
<img width="1399" alt="Screen Shot 2019-10-07 at 7 44 38 AM" src="https://user-images.githubusercontent.com/4141527/66322397-e8546080-e8d6-11e9-9b74-ba7690c0e3a9.png">
[...]

PR #502, opened 2019-10-07
Full Width Site

cKellyDesign:
This PR is for #501 and currently makes the entire Reveal Webapp full-width.


src/App/App.tsx
```diff
@@ -65,197 +65,195 @@ class App extends Component {
       <Container fluid={true}>
         <Helmet titleTemplate={`%s | ` + WEBSITE_NAME} defaultTitle="" />
         <ConnectedHeader />
-        <Container>
```
cKellyDesign: This line was removed

src/App/App.tsx
```diff
@@ -65,197 +65,195 @@ class App extends Component {
...
-                <ConnectedPrivateRoute
-                  disableLoginProtection={DISABLE_LOGIN_PROTECTION}
-                  exact={true}
-                  path={LOGOUT_URL}
-                  component={ConnectedLogout}
-                />
-              </Switch>
-            </Col>
-          </Row>
-        </Container>
```
cKellyDesign: and this line was removed, everything else is just indentation.

moshthepitt:
weird

I think it looks a little weird.

@rowo @OluwatobiMayowa what are your thoughts?

src/App/App.tsx
moshthepitt: Instead of removing it, would it be possible to define a width for it other than the default (I think it might be 1140px or something like that)?

rowo:
@moshthepitt @cKellyDesign 
I agree with mosh, the nav should be updated to be full width too (to line up with the red lines of the content). The boostrap site shows this in their nav: https://getbootstrap.com/docs/4.0/utilities/sizing/

src/App/App.tsx
cKellyDesign: @moshthepitt what is the value of keeping it and re-defining it as fluid?

src/App/App.tsx
cKellyDesign: https://github.com/onaio/reveal-frontend/pull/502#issuecomment-539452200

Oh i see, it's to match the header's indentation. Sawa then just changing this to fluid makes sense.

cKellyDesign:
@rowo in this case the header was wrapped in more `Container` elements than the rest of the page, compounding more padding. The body of the page is now wrapped in the same number of Containers as the header, so everything has the same amount of padding no matter the page width.

@moshthepitt @pld I think this one is ready to merge!

Screen Shot 2019-10-08 at 7 57 35 AM
Screen Shot 2019-10-08 at 7 58 11 AM

src/App/App.tsx
cKellyDesign: Good catch @moshthepitt thanks

PR #506, opened 2019-10-08
Move type definitions out of settings

### LINKED ISSUE #469: Move type definitions out of settings.ts
We need to move all the type definitions out of settings.ts and into something like types.ts.  This is based on the conversation here https://github.com/onaio/reveal-frontend/pull/459#discussion_r329840979

DavisRayM:
Moves type definitions out of settings.ts into a separate types.ts file adjacent to helpers.ts.

Fix #469

### DIFF
--- a/src/configs/settings.ts
+++ b/src/configs/settings.ts
@@ -35,6 +35,7 @@ import {
   OPENSRP_OAUTH_STATE,
   OPENSRP_USER_URL,
 } from './env';
+import { IndicatorRows, JurisdictionTypes } from './types';
 
 /** Interfaces */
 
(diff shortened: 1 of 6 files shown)


src/configs/settings.ts
```diff
@@ -13,6 +13,13 @@
  */
 import { Providers } from '@onaio/gatekeeper';
 import { Expression, LngLatBoundsLike } from 'mapbox-gl';
+import {
+  ActionReasonType,
+  GoalPriorityType,
+  PlanActionCodesType,
+  PlanActivities,
+  UseContextCodesType,
+} from '../containers/forms/PlanForm/types';
```
cKellyDesign: Seems like this is a little circular, but for this type of thing (pun intended) I think it makes sense!

src/configs/types.ts
```diff
@@ -0,0 +1,36 @@
+import {
+  adminLayerColors,
+  GREEN_THRESHOLD,
+  IndicatorRowItem,
+  JurisdictionLevels,
+  ORANGE_THRESHOLD,
+  YELLOW_THRESHOLD,
+} from './settings';
```
cKellyDesign: Wonderful

moshthepitt:
@DavisRayM please update the note at the top of the settings module and then we can merge

src/configs/settings.ts
moshthepitt: yeah, lol, :)

PR #509, opened 2019-10-09
Team assignment assigning practitioner workflow

p-netm:
- rebased from the TeamAssignment-SingleOrganization;
-Adds workflow for assigning Practitioner to Teams


p-netm:
practitioner
practitioner1

src/containers/pages/OrganizationViews/AssignPractitioners/index.tsx
```diff
@@ -0,0 +1,279 @@
...
+    const allPractitioners = await serve.list();
+    return formatOptions(allPractitioners);
+  };
+
+  // TODO - Hack in this: typings for the changeHandler function.
+  /** This sets the state selectedOptions
+   * @param {ValueType<SelectOption>} -  the so far selected options
+   * @param {ActionMeta} - information on the change event; custom react-select event
+   */
+  const changeHandler = (chosenOptions: any, { action, removedValue, option }: any) => {
```
cKellyDesign: Why the `any` hack and not the imported Typings?

src/containers/pages/OrganizationViews/AssignPractitioners/index.tsx
```diff
@@ -0,0 +1,279 @@
...
+  const promiseOptions = async (typedChars: string) => {
+    // merging practitioners that are assigned to the organization and all practitioners
+    const orgPractitioners = formatOptions(assignedPractitioners, true);
+    const allPractitioners = await loadAllPractitioners();
+    const mergedOptions = {
+      ...keyBy(orgPractitioners, option => option.value),
+      ...keyBy(allPractitioners, option => option.value),
+    };
+    return filterOptions(typedChars, values(mergedOptions));
+  };
```
cKellyDesign: Could we instead make the API call ~1sec after the user stops typing?

src/containers/pages/OrganizationViews/AssignPractitioners/index.tsx
```diff
@@ -0,0 +1,279 @@
...
+    // selected options
+    const practitionerIds = selectedOptions.map(option => option.value);
+    const jsonArrayPayload = practitionerIds.map((practitionerId, index) => ({
+      active: true,
+      code: PRACTITIONER_CODE,
+      identifier: generateNameSpacedUUID(`${Date.now()} ${index}`, PRACTITIONER_ROLE_NAMESPACE),
+      organization: organization.identifier,
+      practitioner: practitionerId,
+    }));
+    const serve = new serviceClass(`${OPENSRP_PRACTITIONER_ROLE_ENDPOINT}/add`);
```
cKellyDesign: I think we have `add` as a constant

src/containers/pages/OrganizationViews/AssignPractitioners/index.tsx
```diff
@@ -0,0 +1,279 @@
...
+    serve.create(jsonArrayPayload).then(() => {
+      loadOrgPractitioners(organization.identifier, serviceClass, fetchPractitionerRolesCreator);
+      // TODO - possible candidate for setting state on unmounted component
+      setSelectedOptions([]);
+    });
```
cKellyDesign: Should we add a `.catch` here, even if it's empty?

src/containers/pages/OrganizationViews/AssignPractitioners/index.tsx
```diff
@@ -0,0 +1,279 @@
...
+              readOnly={true}
+              name={option.name}
+              id={option.identifier}
+              value={option.identifier}
+            />
+          </section>
+        ))
+      ) : (
+        <section>
+          <p className="text-info"> No Practitioners Added yet</p>
```
cKellyDesign: Let's move this to a constant

src/containers/pages/OrganizationViews/AssignPractitioners/index.tsx
p-netm: I believe the typings for this changeHandler are not in sync with the library code
the allowed type i.e [ActionMeta](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/643edf5b1c8a4e32d2f5a1e61cfa64dac6ba098d/types/react-select/src/types.d.ts#L69) for the second argument  to the changeHandler only accounts for the `action`, but doesn't for the removedValue, or option

src/containers/pages/OrganizationViews/AssignPractitioners/index.tsx
p-netm: yeah, i will see how i can factor this in

src/containers/pages/OrganizationViews/AssignPractitioners/index.tsx
p-netm: just worried that this is an error that should be avoided instead of handled.

src/containers/pages/OrganizationViews/AssignPractitioners/index.tsx
```diff
@@ -0,0 +1,279 @@
...
+  const loadAllPractitioners = async (service: typeof OpenSRPService = OpenSRPService) => {
+    const serve = new service(OPENSRP_PRACTITIONER_ENDPOINT);
+    const allPractitioners = await serve.list();
+    return formatOptions(allPractitioners);
+  };
+
+  // TODO - Hack in this: typings for the changeHandler function.
+  /** This sets the state selectedOptions
+   * @param {ValueType<SelectOption>} -  the so far selected options
+   * @param {ActionMeta} - information on the change event; custom react-select event
```
moshthepitt: You havent named the params in this docstring

src/containers/pages/OrganizationViews/AssignPractitioners/index.tsx
moshthepitt: What do you think needs to be done to resolve the "hack" - should we create an issue with those details?

src/containers/pages/OrganizationViews/AssignPractitioners/index.tsx
```diff
@@ -0,0 +1,279 @@
...
+        break;
+      case 'select-option':
+        setSelectedOptions([...selectedOptions, option]);
+        break;
+      case 'clear':
+        chosenOptions = chosenOptions.filter((v: any) => !v.isFixed);
+    }
+  };
+
+  // TODO - This will initiate an api request for the same exact data each time someone types
```
moshthepitt: What do you think needs to be done to resolve this TODO - should we create an issue with those details?

src/containers/pages/OrganizationViews/AssignPractitioners/index.tsx
moshthepitt: @p-netm but leaving it this way will result in an exception being raised - now?  Much better to somehow get a nice error message to the user than to have exceptions happen - now?  (I think we have an issue to deal with errors somewhere)

src/containers/pages/OrganizationViews/AssignPractitioners/tests/index.test.tsx
```diff
@@ -0,0 +1,157 @@
+import reducerRegistry from '@onaio/redux-reducer-registry';
```
moshthepitt: What can we do to remove the Typescript warnings in this module?

src/containers/pages/OrganizationViews/AssignPractitioners/utils.ts
```diff
@@ -0,0 +1,54 @@
...
+/** interface of an option in the component's state */
+export interface SelectOption {
+  readonly label: string;
+  readonly value: string;
+  readonly isFixed: boolean;
+}
+
+/** custom styling for fixed options */
+export const styles = {
+  multiValue: (base: any, state: any) => {
```
moshthepitt: why so much `any` usage here?

src/components/LinkAsButton/index.tsx
```diff
@@ -21,7 +21,7 @@ const LinkAsButton = (props: Props) => {
 
 /** default props for LinkAsButton component */
 const defaultProps: Props = {
-  classNameProp: 'focus-investigation btn btn-primary float-right mt-5',
+  classNameProp: 'focus-investigation btn btn-primary float-right mt-5 mr-3',
```
moshthepitt: What effect does this have to everyone who uses this component?  Is it viable to only add the `mr-3` to the particular button that needs the right margin?

src/constants.tsx
```diff
@@ -1,5 +1,3 @@
-import { Marker } from 'mapbox-gl';
```
moshthepitt: Nice catch

src/containers/pages/OrganizationViews/AssignPractitioners/index.tsx
```diff
@@ -0,0 +1,279 @@
...
+const defaultAssignPractitionerProps: AssignPractitionerProps = {
+  assignedPractitioners: [],
+  fetchOrganizationsCreator: fetchOrganizations,
+  fetchPractitionerRolesCreator: fetchPractitionerRoles,
+  organization: null,
+  serviceClass: OpenSRPService,
+};
+
+/** type intersection for all types that pertain to the props */
+export type PropsTypes = AssignPractitionerProps & RouteComponentProps<RouteParams>;
```
moshthepitt: Nice

src/containers/pages/OrganizationViews/AssignPractitioners/index.tsx
p-netm: will add the error boundary

src/containers/pages/OrganizationViews/AssignPractitioners/index.tsx
```diff
@@ -0,0 +1,279 @@
...
+  // TODO - scenario where this request doesn't return all practitioners in a single query
+  /** load all practitioners at least all of those returned
+   * in a single call to Practitioners endpoint
+   * @param {typeof OpenSRPService} service - the OpenSRPService
+   */
+  const loadAllPractitioners = async (service: typeof OpenSRPService = OpenSRPService) => {
+    const serve = new service(OPENSRP_PRACTITIONER_ENDPOINT);
+    const allPractitioners = await serve.list();
+    return formatOptions(allPractitioners);
+  };
```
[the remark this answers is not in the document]
p-netm: Yes i think it can, I was hesitant due to the fact that there is no way to tell opensrp for this endpoint how many records should be returned, so i have to trust that the default max page size for this endpoint does not change.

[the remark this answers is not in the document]
p-netm: No actually on second thought we would need some sought of pagination on this endpoint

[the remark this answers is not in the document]
p-netm: will create a ticket for this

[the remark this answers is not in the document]
p-netm: #535

src/containers/pages/OrganizationViews/AssignPractitioners/utils.ts
p-netm: Will create issue to investigate this

moshthepitt: @p-netm have you created the issue?

p-netm:
Ticket for tests #526

moshthepitt:
Over to you @cKellyDesign

moshthepitt:
@cKellyDesign I'll merge this so that we can get a release ready, with the understanding that you may have further change requests.

PR #510, opened 2019-10-09
Remove indicatorRowItem Reference

cKellyDesign:
@DavisRayM it looks like a type was added to the types.ts file referencing `IndicatorRowItem` which is no longer in settings.ts. It doesn't look like there are any references to these types, am I missing something or are these okay to remove?

### DIFF
--- a/src/configs/types.ts
+++ b/src/configs/types.ts
@@ -1,7 +1,6 @@
 import {
   adminLayerColors,
   GREEN_THRESHOLD,
-  IndicatorRowItem,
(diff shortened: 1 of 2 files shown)


src/configs/types.ts
```diff
@@ -1,7 +1,6 @@
 import {
   adminLayerColors,
   GREEN_THRESHOLD,
-  IndicatorRowItem,
```
cKellyDesign: The import throwing compiling errors.

Issue #515, opened 2019-10-11
Allow active plans to be edited in the UI


cKellyDesign:
We should be able to see and click on Active plans in the IRS Planning list view.

moshthepitt:
- Does this need to be for every plan?
- Should we just trust that things can and change in a plan, and it should be okay to edit it whenever?
- Does it apply to all plan fields, or are there some tha should not be editable?

cKellyDesign:
@mberg @craigappl @pld @githengi - please consider Mosh's questions above ☝️

cKellyDesign:
> Does this need to be for every plan?

For now I'm making changes which should only effect IRS plan Jurisdiction selection (and eventually assignment).

moshthepitt:
For context, we intentionally disabled [editing many fields](https://github.com/onaio/reveal-frontend/blob/master/src/containers/forms/PlanForm/index.tsx#L884) on plans.  Looks like we need to review this.

moshthepitt:
> For now I'm making changes which should only effect IRS plan Jurisdiction selection (and eventually assignment).

But is it absolutely agreed upon that we should do this, or is what is happening in Zambia a rare thing?  I guess I am just worried that freely editing plans probably has a lot of consequences - though we are doing it clunkily anyway right now.

PR #516, opened 2019-10-11
515 enabling editing on finalized IRS plans

cKellyDesign:
This resolves: #515 
Closes: #542 
This PR:
* Adds a list page and header nav for `Assigning > IRS` plans
* Renders organization names instead of assignment counts in Assignment Cell
* Filters `Planning > IRS` to only show draft plans
* Renders `Save Assignments` button when assigning teams to finalized plan(-jurisdiction)s
* Does not update the PlanDefinition of Finalized plans (when saving assignments)
* Renders only Locations selected in the Finalized plan in assignment drilldown table
* Does not render checkboxes in the assignment drilldown table
* Redirects the user to the Plan Assignment view when saving a Finalized Plan
* Renames header nav items and nests Teams, Practitioners, and Users under Admin dropdown


cKellyDesign:
Setting this to DNM because we need to fully understand the implications of editing Finalized plans.

src/containers/pages/IRS/assignments/index.tsx
```diff
@@ -0,0 +1,181 @@
...
+  async function loadData() {
+    try {
+      setLoading(plans.length < 1); // only set loading when there are no plans
+      OpenSrpPlanService.list()
+        .then(planResults => {
+          // filter for IRS plans
+          const irsPlans = planResults.filter((p: PlanPayload) => {
+            for (const u of p.useContext) {
+              if (u.code === useContextCodes[0]) {
+                if (u.valueCodableConcept === IRS_PLAN_TYPE) {
```
pld: can we replace 77-81 with

```javascript
return (u.valueCodableConcept === IRS_PLAN_TYPE);
```

src/containers/pages/IRS/assignments/index.tsx
```diff
@@ -0,0 +1,181 @@
...
+                  return true;
+                } else {
+                  return false;
+                }
+              }
+            }
+            return false;
```
pld: we only get here if `p.useContext` is empty, is that correct?

src/containers/pages/InterventionPlan/IRS/plan/index.tsx
```diff
@@ -446,7 +449,11 @@ class IrsPlan extends React.Component<
     }
 
     // update state after fetching plan from OpenSRP
-    if (!newPlan && planById && planById.plan_jurisdictions_ids) {
+    // update state after saving plan to store
+    if (
+      (!newPlan && planById && planById.plan_jurisdictions_ids) ||
```
pld: you can simplify this to mention `planById` only once

src/containers/pages/InterventionPlan/IRS/plan/index.tsx
```diff
@@ -1794,15 +1801,21 @@ class IrsPlan extends React.Component<
...
-    const isFocusJurisdictionTopLevel = tableCrumbs[0] && focusJurisdictionId === tableCrumbs[0].id;
+    const { newPlan, focusJurisdictionId, tableCrumbs } = state;
+    const { assignmentsArray, isDraftPlan, jurisdictionsById, planId } = this.props;
 
     if (!newPlan || !newPlan.plan_jurisdictions_ids) {
       return null;
     }
 
+    // finalized plans should only show rows for jurisdictions selected in the plan
+    const filteredJurisdictionIds = isDraftPlan
```
pld: this might be simpler as

```javascript
!isDraftPlan && state.newPlan!.plan_jurisdictions_ids || state.filteredJurisdictionIds
```

src/containers/pages/IRS/assignments/index.tsx
```diff
@@ -0,0 +1,181 @@
...
+          );
+          return fetchPlans(irsPlanRecords);
+        })
+        .catch(err => {
+          // console.log('ERR', err)
+        });
+    } catch (e) {
+      // do something with the error?
+    } finally {
+      setLoading(false);
```
p-netm: might result in a `setState on unmounted component error`, possible solution would be to have `loading` as a computed variable of the plans length. i.e. `loading = plans.length < 1`

src/containers/pages/IRS/assignments/index.tsx
cKellyDesign: Yeah we only get there if for some reason the plan doesn't have the useContext for intervention type. Not sure why/how this would happen though.

src/containers/pages/InterventionPlan/IRS/plan/index.tsx
cKellyDesign: Ooh nice one

src/containers/pages/IRS/assignments/index.tsx
cKellyDesign: @p-netm thanks, this was really insightful! I ended up moving this line up just after the OpenSRP call (to which I added `await`) so it should no longer be in a block that will be called asynchronously. To your point, we should NEVER be setting component state in an asynchronous block or callback. 

👍* 💯

src/containers/pages/InterventionPlan/IRS/plan/index.tsx
cKellyDesign: Done

src/containers/pages/InterventionPlan/IRS/plan/index.tsx
cKellyDesign: It is so

src/containers/pages/IRS/assignments/index.tsx
cKellyDesign: done!

PR #483, opened 2019-10-02
Team assignment practitioner form

p-netm:
Adds a form for creating and editing a practitioner record

PR #519, opened 2019-10-11
Team assignment practitioner create view

### LINKED ISSUE #458: Team-Assignment - Creating a Pracitioner
Preliminary work towards assigning practitioners to a team requires a page to actually create the practitioner.

Task:
- Adding a form and a page to create practitioners

Blocker:
- Need an userId from openMRS , so as to link the the practitioner to an openMRS user.
 Will probably need a proxy on the opensrp server that helps get this data from openMRS.

p-netm:
Create/Editing Practitioner 
closes #458 closes #483


p-netm:
Issue to add tests 
#526

src/containers/forms/PractitionerForm/UserIdSelect/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const loadOpenMRSUsers = async (service: typeof OpenSRPService = OpenSRPService) => {
+    const currentUserIndex = 0;
+    let filterParams = {
+      page_size: OPENMRS_USERS_REQUEST_PAGE_SIZE,
+      start_index: currentUserIndex,
+    };
+    const serve = new service('user');
+    let responseSize: number = 100;
+
+    while (responseSize === OPENMRS_USERS_REQUEST_PAGE_SIZE) {
```
p-netm: Questionable: Each time the component mounts, it makes iterative calls to opensrp untill it gets all users from openSRP proxied via opensrp.

src/constants.tsx
```diff
@@ -328,3 +328,7 @@ export const FI_PLAN_TYPE = 'FI';
 export type FI_PLAN_TYPE = typeof FI_PLAN_TYPE;
 export const IRS_PLAN_TYPE = 'IRS';
 export type IRS_PLAN_TYPE = typeof IRS_PLAN_TYPE;
+
+// error messages
```
cKellyDesign: Nice, good idea.

src/containers/forms/PractitionerForm/UserIdSelect/index.tsx
cKellyDesign: What happens here if there are exactly 100 openMRS users to return? Oh the second call will have zero results, which should have no effect when calling `setOpenMRSUsers` with an empty array, ama?

p-netm: oops there is a bug, each call to setState will override already stored openmrs users, ill fix that quckly. 
@cKellyDesign The logic is that each request will have results array with a max length  of a 100 objects. So if we get less that a 100 records in any request then that's how i know that the page returned in the that request is the last page, and therefore i should not make any subsequent calls.
so ``` while any request returns the maximum request size:.
assume there is more data and make another request,
else know we just got the last page and stop making further requests ```

src/containers/forms/PractitionerForm/UserIdSelect/index.tsx
```diff
@@ -0,0 +1,89 @@
+/** wrapper around react-select component that enables one
+ * to select a single user from openmrs
+ */
+import React, { useEffect, useState } from 'react';
+import Select from 'react-select';
+import { ValueType } from 'react-select/src/types';
+import { OpenSRPService } from '../../../../services/opensrp';
+
+const OPENMRS_USERS_REQUEST_PAGE_SIZE = 100;
```
moshthepitt: Can we please move this to an env variable, with a default int value?  So that we don't have to make a "code change" in the event that we want to change it.

src/containers/forms/PractitionerForm/UserIdSelect/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+export interface Props {
+  onChangeHandler?: (value: Option) => void;
+  serviceClass: typeof OpenSRPService;
+}
+
+export const defaultProps = {
+  serviceClass: OpenSRPService,
+};
+
+export interface Option {
```
moshthepitt: Lets add a docstring for this guy

src/containers/forms/PractitionerForm/UserIdSelect/tests/index.test.tsx
```diff
@@ -0,0 +1,38 @@
...
+import toJson from 'enzyme-to-json';
+import React from 'react';
+import UserIdSelect from '..';
+import { OpenSRPService } from '../../../../../services/opensrp';
+import { response } from './fixtures';
+
+// tslint:disable-next-line: no-var-requires
+const fetch = require('jest-fetch-mock');
+
+describe('src/*/forms/userIdSelect', () => {
```
moshthepitt: I feel that these tests are insufficient.  How can we include the fixtures in the tests?  Can we test that the list of users is rendered?

src/containers/forms/PractitionerForm/index.tsx
```diff
@@ -0,0 +1,246 @@
...
+  SAVE,
+  SAVING,
+  USERNAME,
+  YES,
+} from '../../../constants';
+import { generateNameSpacedUUID } from '../../../helpers/utils';
+import { OpenSRPService } from '../../../services/opensrp';
+
+/** Namespace for generating uuids */
+const PractitionerFormNameSpace = '78295ac0-df73-11e9-a54b-dbf5e5b2d668';
```
moshthepitt: Can we make this an env variable please?

src/containers/forms/PractitionerForm/index.tsx
```diff
@@ -0,0 +1,246 @@
...
+  userId: undefined,
+  username: undefined,
+};
+
+/** yup validations for practitioner data object from form */
+export const PractitionerSchema = Yup.object().shape({
+  active: Yup.boolean(),
+  identifier: Yup.string(),
+  name: Yup.string()
+    .matches(/^[a-zA-Z0-9]+([_ -]?[a-zA-Z0-9])*$/, NAME_REGEX_ERROR)
```
moshthepitt: Will this come back to bite us?  Why do we need to validate the name using a regular expression?

src/containers/forms/PractitionerForm/index.tsx
```diff
@@ -0,0 +1,246 @@
...
+              <ErrorMessage
+                name="name"
+                component="small"
+                className="form-text text-danger name-error"
+              />
+            </FormGroup>
+
+            {!editMode ? (
+              <FormGroup>
+                <Label for="userId"> openMRS username </Label>
```
moshthepitt: I am wondering whether we need to mention OpenMRS at all here?  Does the person using this case about OpenMRS at all?  Should we drop the OpenMRS reference?  cc: @cKellyDesign @craigappl @pld

src/containers/forms/PractitionerForm/tests/index.test.tsx
```diff
@@ -0,0 +1,34 @@
+import { mount, shallow } from 'enzyme';
```
moshthepitt: Do you feel that these tests offer us full coverage of this component and its usage?

src/containers/pages/PractitionerViews/CreateEditPractitioner/tests/index.test.tsx
```diff
@@ -0,0 +1,264 @@
...
+const history = createBrowserHistory();
+
+describe('src/containers/pages/CreateEditOrganization', () => {
+  beforeEach(() => {
+    jest.resetAllMocks();
+    store.dispatch(practitionerDucks.removePractitionerRolesAction);
+    store.dispatch(practitionerDucks.removePractitionersAction);
+  });
+
+  it('renders EditTeamView without crashing', () => {
```
moshthepitt: Is a practitioner the same thing as a team?

src/containers/forms/PractitionerForm/index.tsx
p-netm: will fix this

src/containers/pages/PractitionerViews/CreateEditPractitioner/tests/index.test.tsx
p-netm: nope

src/containers/forms/PractitionerForm/tests/index.test.tsx
p-netm: not currently, i was hoping to do that as a different issue, since i think it might take awhile, created #526

src/containers/forms/PractitionerForm/index.tsx
pld: good catch, let's remove all the OpenMRS references, they shouldn't be aware of OpenMRS, we use it opaquely

p-netm: @pld @moshthepitt What should the label read, should it just be username and yet they are not going to use any arbitrary name of their choosing

src/containers/forms/PractitionerForm/UserIdSelect/tests/index.test.tsx
p-netm: will work on the tests in #526

src/containers/forms/PractitionerForm/index.tsx
moshthepitt: Yeah lets just drop the "openmrs" part and leave it as just username for this and the other places we mentioned openmrs.  Also consider moving to constants module

moshthepitt:
@p-netm please fix conflicts

src/containers/pages/PractitionerViews/CreateEditPractitioner/tests/index.test.tsx
moshthepitt: @p-netm then perhaps we should not reference teams in this file?

moshthepitt:
@p-netm please fix conflicts

src/constants.tsx
```diff
@@ -356,3 +358,13 @@ export const BLOOD_SCREENING_CODE = 'Blood Screening';
...
+
+// plans
+export const ROUTINE_PLAN = 'Routine';
+export type ROUTINE_PLAN = typeof ROUTINE_PLAN;
+export const CASE_TRIGGERED_PLAN = 'Case-triggered';
+export type CASE_TRIGGERED_PLAN = typeof CASE_TRIGGERED_PLAN;
+export const FI_PLAN_TYPE = 'FI';
+export type FI_PLAN_TYPE = typeof FI_PLAN_TYPE;
+export const IRS_PLAN_TYPE = 'IRS';
+export type IRS_PLAN_TYPE = typeof IRS_PLAN_TYPE;
```
moshthepitt: These were removed and their usage refactored in https://github.com/onaio/reveal-frontend/pull/551

src/configs/env.ts
```diff
@@ -201,3 +201,9 @@ export type DEFAULT_PLAN_VERSION = typeof DEFAULT_PLAN_VERSION;
 
 export const ENABLE_PRACTITIONERS = process.env.REACT_APP_ENABLE_PRACTITIONERS === 'true';
 export type ENABLE_PRACTITIONERS = typeof ENABLE_PRACTITIONERS;
+
+export const PRACTITIONER_FORM_NAMESPACE = '78295ac0-df73-11e9-a54b-dbf5e5b2d668';
```
moshthepitt: Can we read this from an env variables, having this value as the fallback default? e.g. https://github.com/onaio/reveal-frontend/blob/master/src/configs/env.ts#L8

src/configs/env.ts
```diff
@@ -201,3 +201,9 @@ export type DEFAULT_PLAN_VERSION = typeof DEFAULT_PLAN_VERSION;
...
+export type PRACTITIONER_FORM_NAMESPACE = typeof PRACTITIONER_FORM_NAMESPACE;
+
+export const OPENMRS_USERS_REQUEST_PAGE_SIZE = 100;
```
moshthepitt: Can we read this from an env variable, having this value as the fallback default? e.g. https://github.com/onaio/reveal-frontend/blob/master/src/configs/env.ts#L44

PR #539, opened 2019-10-17
Add Gisida timeout

moshthepitt:
- Use an env var to set Gisida timeout duration
- Remove erroneous usage of `SUPERSET_MAX_RECORDS` in GisidaWrapper


src/components/GisidaWrapper/index.tsx
```diff
@@ -184,7 +180,7 @@ class GisidaWrapper extends React.Component<GisidaProps, GisidaState> {
           // Dirty work around! Arbitrary delay to allow style load before adding layers
           setTimeout(() => {
             this.initMap(null, null);
-          }, SUPERSET_MAX_RECORDS);
+          }, GISIDA_TIMEOUT);
```
cKellyDesign: 👍

PR #540, opened 2019-10-17
[hotfix] Filter out invalid Jurisdiction IDs saved in old plan definitions

cKellyDesign:
This branch adds a quick step to filter all Plan Definition `plan_jurisdictions_ids` to remove ids which are no longer a part of the location tree.


moshthepitt:
What kind of test can we write for this? @cKellyDesign

PR #553, opened 2019-10-23
547 plans reducer getter

### LINKED ISSUE #547: Add Plans Reducer getter for all Plans
Currently all the getter methods in the Plans reducer require a parameter for Intervention Type and default to `FI`, and we simply need a way to return all plans without filtering by type. Reasoning behind this is so we don't need to update pages which show all plans every time we add a new Plan type.

p-netm:
closes #547 

__previously__: 
plans selectors had `FI` as the default value for intervention type filter

__now__:
Plan selectors do not filter against any intervention type be default,


moshthepitt:
@p-netm please fix conflicts

Issue #557, opened 2019-10-25
Zambia assign page broken


moshthepitt:
When you click into a plan to do assignment, the page breaks and we get this in the console

```
Uncaught (in promise) TypeError: Cannot read property 'parent_id' of undefined
    at t.value (main.9eeaa40d.chunk.js:1)
    at t.value (main.9eeaa40d.chunk.js:1)
    at main.9eeaa40d.chunk.js:1
```

PR #559, opened 2019-10-25
557 uncaught errors from promises

cKellyDesign:
Resolves: #557 

This PR:
* adds catch blocks to Promises in plan/assign page
* adds conditional to only check properties of existing jurisdictions


moshthepitt:
@cKellyDesign can we add a test to protect against regression?

cKellyDesign:
@moshthepitt test added

PR #564, opened 2019-10-27
Add IRS indicators Legend to IRS reports

moshthepitt:
Adds a component that explains the colors in IRS reporting tables similar to what was in mSpray - see https://macepa.mspray.onalabs.org/

irs legend2

### DIFF
--- a/src/configs/settings.ts
+++ b/src/configs/settings.ts
@@ -725,8 +725,8 @@ export interface IndicatorThresholds {
   [key: string]: IndicatorThresholdItem;
 }
 
-/** Indicator Thresholds for NA (Namibia) */
-export const indicatorThresholdsNA: IndicatorThresholds = {
+/** IRS Reporting configs */
+export const indicatorThresholdsIRS: IndicatorThresholds = {
   GREEN_THRESHOLD: {
     color: '#2ECC40',
     name: 'Green',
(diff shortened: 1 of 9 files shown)


src/configs/settings.ts
```diff
@@ -711,35 +712,44 @@ export const symbolLayerConfig = {
...
-    orEquals?: boolean;
-    value: number;
-  };
+  [key: string]: IndicatorThresholdItem;
 }
 
-/** Indicator Thresholds for NA (Namibia) */
-export const indicatorThresholdsNA: IndicatorThresholds = {
+/** IRS Reporting configs */
+export const indicatorThresholdsIRS: IndicatorThresholds = {
```
cKellyDesign: @moshthepitt should this be `IRS_NA` to support different thresholds and colors for different implementations of the IRS intervention type? (Like if Zambia has different thresholds or if Thailand wants to use different colors, etc?)

moshthepitt: @cKellyDesign yes it should, although right now ZM and NA are the same values so I thought we should do the configurability when we need it.

moshthepitt: @cKellyDesign thoughts on that?

cKellyDesign: Ah yeah ok sounds good, might as well not over engineer it until it's necessary :)

cKellyDesign: :shipit:

Issue #563, opened 2019-10-27
Fix goal typing


moshthepitt:
- The goals reducer module type definition needs to be made more consistent with the type definitions for goals in settings/PlanForm
- Fix type warnings here https://github.com/onaio/reveal-frontend/blob/master/src/helpers/indicators.tsx#L82

Issue #498, opened 2019-10-06
Fix spray coverage for rooms description


moshthepitt:
![rooms](https://user-images.githubusercontent.com/372073/66266818-cacdbb00-e832-11e9-9d4e-8809db1dfe27.png)

PR #567, opened 2019-10-28
Action Codes and Goal Units

cKellyDesign:
Closes: #498 
Closes: #563


src/store/ducks/goals.ts
```diff
@@ -1,14 +1,15 @@
...
+import { PlanActionCodes } from '../../configs/settings';
 import { FlexObject } from '../../helpers/utils';
 
 /** the reducer name */
 export const reducerName = 'goals';
 
 /** interface for goal Object */
 export interface Goal {
-  action_code: string;
+  action_code: typeof PlanActionCodes[number];
```
moshthepitt: I think this one is already defined here https://github.com/onaio/reveal-frontend/blob/b0ea6c03e5479537a782aa5340b757087d3b5214/src/containers/forms/PlanForm/types.ts#L29

src/constants.tsx
```diff
@@ -96,7 +97,7 @@ export const STRUCTURE_LAYER = 'structure-layer';
 export type STRUCTURE_LAYER = typeof STRUCTURE_LAYER;
 export const PROGRESS = 'Progress';
 export type PROGRESS = typeof PROGRESS;
-export const STRUCTURES = 'structure(s)';
+export const STRUCTURES = `${STRUCTURES_UNIT_TYPE}(s)`;
```
moshthepitt: Will this switch everything to be the value of the env defined?  It seems for #498 we want just the last indicator to say rooms while the others say structures.

cKellyDesign: oh hmm, i thought one of the countries measured sprayable units as rooms, let me get clarification in the ticket

src/store/ducks/goals.ts
cKellyDesign: oh cool, will switch

src/constants.tsx
cKellyDesign: thanks for the explanation @moshthepitt changes made

moshthepitt:
Nice!  What kind of test can we add for this, @cKellyDesign ?

cKellyDesign:
Haaaaa, of the Unit variety! Adding now :)

moshthepitt:
Nice :)

PR #583, opened 2019-10-31
Clearable select

### LINKED ISSUE #582: BUG: Cannot remove selected practitioner from select in assignPractitioner
In the assign Practitioner view shown below:
newIssue

One cannot remove the option that they select first. i.e. the practitioner record `Lusaka` shown above cannot be removed from the select by actions that  should remove it e.g. clicking backspace or clicking on the `x` on the option itself. this behavior is occurring when we have just one option selected,

p-netm:
can now clear any option in the assign practitioner's page Select irrespective of when it was selected as an option; by conventionally clicking backspace or the cancel `x` on the option itself.

Also minor enhancement: adds breadCrumb link back to single organization view for the organization that's being assigned practitioners.
closes #582


src/containers/pages/OrganizationViews/AssignPractitioners/index.tsx
```diff
@@ -112,9 +113,6 @@ const AssignPractitioner = (props: PropsTypes) => {
     chosenOptions: ValueType<SelectOption>,
     { action, removedValue, option }: any
   ) => {
-    if (!chosenOptions) {
-      return;
-    }
```
cKellyDesign: @p-netm is this what was blocking the component from deselecting the final option?

p-netm: @cKellyDesign not exactly the final option but the only the first selected option.
Steps to reproduce:
- open an organizations without assigned practitioners
- select one practitioner from the select dropDwon
- try to remove it

cKellyDesign:
@p-netm is this ready to merge?

p-netm:
Yes

PR #591, opened 2019-11-13
Fix for new practitoiner view

### LINKED ISSUE #589: New Practitioner Form page fixes
- [x] Sort the usernames drop-down alphabetically
- [x] Capitalize the `n` in `New Practitioner` breadcrumb and page title

Screen Shot 2019-11-12 at 10 39 52 AM

p-netm:
closes #589 
- Adds sort to the openMrs users dropDown
- Minor Fix for text case in dropDown.


src/containers/forms/PractitionerForm/UserIdSelect/index.tsx
```diff
@@ -102,10 +102,18 @@ export const UserIdSelect: React.FC<Props> = props => {
     setOpenMRSUsers(unMatchedUsers);
   };
 
-  const options = openMRSUsers.map((user: OpenMRSUser) => ({
-    label: user.display,
-    value: user.uuid,
-  }));
+  const options = React.useMemo(() => {
```
cKellyDesign: Nice! It would be cool to measure the performance boost here, if it's substantial using React Memos could be pretty useful when dealing with the heavy IRS location hierarchies.

cKellyDesign:
@p-netm is this ready for review?

p-netm:
> @p-netm is this ready for review?

@cKellyDesign yes, but there is still this https://github.com/onaio/reveal-frontend/issues/589#issuecomment-553482279 requested feature, am not sure what i should do about it.

cKellyDesign:
@p-netm ok cool, I have responded in the ticket (https://github.com/onaio/reveal-frontend/issues/589#issuecomment-555526392) and think it should be pretty straightforward to implement since we should already have all the practitioners in the store, ama? 

If we cannot assume that we have all practitioners in the store then we can simply call OpenSRP API with the `filterByProperties` flag checking for results with the matching `practitioner.name`.

p-netm:
> If we cannot assume that we have all practitioners in the store then we can simply call OpenSRP API with the `filterByProperties` flag checking for results with the matching `practitioner.name`.

@cKellyDesign apparently `filterByProperties` is only available for strucures and locations.

cKellyDesign:
>  we should already have all the practitioners in the store

@p-netm Is there a case where this wouldn't be true?

p-netm:
> > we should already have all the practitioners in the store
> 
> @p-netm Is there a case where this wouldn't be true?

@cKellyDesign yes, once practitioners in the server are more than the default response size which is a thousand.

cKellyDesign:
> once practitioners in the server are more than the default response size which is a thousand.

Cool that makes sense, I will make sure there is a ticket for enforcing unique `practitioner.name`s

Issue #590, opened 2019-11-12
Check if jurisdiction is virtual in map link conditional


cKellyDesign:
Currently the logic determining whether or not to render a Map icon link in the IRS Report drilldown table is whether or not the Jurisdiction is childless. In the location tree, all virtual jurisdictions are childless (b/c they don't have spray areas) but should not load in the map.

- [x] Add `is_virtual_jurisdiction` to Jurisdiction reducer/typings
- [x] Disable the map icon for virtual jurisdictions
- [x] Remove the link itself from the label for virtual jurisdictions

This work is dependent on having virtual jurisdictions in staging superset:
- [x] Include virtual jurisdictions in existing Superset slices (556 for ZM, 554 for NA)
- [x] Include `is_virtual_jurisdiction` column in Superset slices

cKellyDesign:
This `accessor` function should return a normal `<span />` without the map icon if `j.is_virtual === true`: https://github.com/onaio/reveal-frontend/blob/master/src/containers/pages/InterventionPlan/IRS/plan/index.tsx#L1891-L1899

gstuder-ona:
To make this easier to dev, I've added these to staging:
zambia_irs_jurisdictions_with_virtual: https://superset.reveal-stage.smartregister.org/superset/explore/?form_data=%7B%22slice_id%22%3A%20562%7D (slice 562)
zambia_focus_area_irs_with_virtual: https://superset.reveal-stage.smartregister.org/superset/explore/?form_data=%7B%22slice_id%22%3A%20563%7D (slice 563)

You should be able to use them as a backend when testing, they include the "is_virtual_jurisdiction" field.

gstuder-ona:
Not sure if this is SOP, but for compatibility, it would be awesome if the frontend didn't always require the "is_virtual_jurisdiction" field and just defaults it to "false" - that way it'll be a lot easier to coordinate rolling out the new slices to production.

cKellyDesign:
> didn't always require the "is_virtual_jurisdiction" field and just defaults it to "false"

@gstuder-ona I think [this single change](https://github.com/onaio/reveal-frontend/pull/592/commits/3d0f1ec02ef9fc5b1e06765aab20511f828e103a#diff-54e3d3f62215c4c5c9995ba2f144899bR19) will do the trick, it's a little tricky to test even now in staging cause the data looks a bit weird, (looks like there are virtual spray areas which are children of spray areas), but it looks like it will be solid.

PR #592, opened 2019-11-13
590 No map link for virtual locations

cKellyDesign:
This PR resolves: #590


cKellyDesign:
Ah yeah will have to add tests with fixtures containing virtual jurisdictions.

cKellyDesign:
@pld tests added :)

Issue #538, opened 2019-10-16
MDA Plan Scoping


cKellyDesign:
This ticket tracks the scoping conversation around the webapp's responsibility for MDA plans ([confluence ticket](https://smartregister.atlassian.net/wiki/spaces/REV/pages/1213530119/Mass+Drug+Administration))

### Reporting
- [x] R1. Derek has provided [this sheet of dashboard definitions](https://docs.google.com/spreadsheets/d/17DEdIOMuxI-MO3aGBUJBntBgi1VqHgqRUzvKGZ__PUA/edit?usp=sharing)
- [x] R2. We need to define table columns / indicators to display in the report table
- [x] R3. We need to define map layers and sidebar content for the report map
- [ ] R4. We need a timeline for when this needs to be in which production environment(s) @craigappl 
- [ ] R5. We to implement and it needs to be dynamically configurable 

### Planning
- [ ] P1. We need to define the MDA Plan definition
- [ ] P2. We need to know the Activities spec for MDA Plans @githengi 
- [ ] P2. We need a timeline for when this needs to be in which production environment(s) @craigappl 
- [ ] P3. We need to implement and it needs to be dynamically configurable 

### Assigning
- [ ] A1. We need to know if there are any special requirements for MDA assignments @craigappl 
- [ ] A2. We need a timeline for when this needs to be in which production environment(s) @craigappl 
- [ ] A3. We need to implement

## Questions
- [ ] Q0. In what order should we be prioritizing the above sections?
- [x] Q1. Are multiple jurisdictions allowed in an MDA Plan (similar to IRS?)
- [ ] Q2. Do we need Family Management modules (similar to Organizations/Practitioners)? @mberg 
- [ ] Q3. Are MDA Plans considered Focus Investigation plans? (Do we list them as a new table on the `Reporting > Focus Investigations` or should they be listed on their own page `Reporting > MDA`)? @mberg @rowo 
- [ ] Q4. What does `Add support for kicking off MDA round 2` imply for the webapp? @craigappl 
- [ ] Q5. What is the MDA adherence activity goal? Is it 100% of people in the OA? Or is there some other threshold or calculation we need to make?

(@mberg please assume all the questions have your name next to them!)

pld:
is there a timeline for these pieces and for the whole thing? please tag me when input for specific pieces is need, I'm going to unassign myself

cKellyDesign:
[...]

PR #597, opened 2019-11-15
Show MDA Plans in Assign View

cKellyDesign:
For #538 A3, allows users to open MDA type plans in the assignment view and assign teams to plan-jurisdictions.


src/containers/pages/IRS/assignments/index.tsx
```diff
@@ -153,7 +153,8 @@ const mapStateToProps = (state: Partial<Store>): DispatchedStateProps => {
   const planStatus = [PlanStatus.ACTIVE];
   const fiPlans = getPlanRecordsArray(state, InterventionType.FI, planStatus);
   const irsPlans = getPlanRecordsArray(state, InterventionType.IRS, planStatus);
-  const plans = [...fiPlans, ...irsPlans].sort(
+  const mdaPlans = getPlanRecordsArray(state, InterventionType.MDA, planStatus);
+  const plans = [...fiPlans, ...irsPlans, ...mdaPlans].sort(
     (a: PlanRecord, b: PlanRecord) => Date.parse(b.plan_date) - Date.parse(a.plan_date)
   );
```
cKellyDesign: Would be nice if we could just allow all types of plans render, though I'm not sure how to iterate over an `enum` like this. I think the argument against passing `null` instead of the plan type was to make sure and only ever show legitimate plans.

p-netm: > Would be nice if we could just allow all types of plans render,

Yeap

>  though I'm not sure how to iterate over an `enum` like this.

Not sure if the you are talking about the code itself, in which case i think you could do 
```Object.values``` or a `for ... in` loop.

> I think the argument against passing `null` instead of the plan type was to make sure and only ever show legitimate plans

yeah makes sense, maybe we can put the enum iteration in the `getPlanRecordsArray` and then have all the different InterventionTypes as the default argument. Adding more intervention types would require we only add to the said argument. I worked on something like this a while back #553 and i went with using a default of null did not know then about __make sure and only ever show legitimate plans.__

p-netm: >  maybe we can put the enum iteration in the

yeah, sorry, no iteration would be required in this case.

PR #601, opened 2019-11-18
Order IRS Reporting Jurisdiction Rows

cKellyDesign:
This PR simply adds a secondary sort parameter to the jurisdictions Superset slice request for IRS reporting drilldown tables.


src/containers/pages/IRS/JurisdictionsReport/index.tsx
```diff
@@ -91,7 +91,7 @@ const JurisdictionReport = (props: GenericJurisdictionProps & RouteComponentProp
           fetchJurisdictionsParams = superset.getFormData(
             SUPERSET_MAX_RECORDS,
             [{ comparator: planId, operator: '==', subject: 'plan_id' }],
-            { jurisdiction_depth: true }
+            { jurisdiction_depth: true, jurisdiction_name: true }
```
gstuder-ona: So this will still have the problem of "Eastern (other)", for example, sorting between real jurisdictions and not at the end?  The data coming from the DB has an "is_virtual_jurisdiction" sort added to it, maybe that would work here too?

cKellyDesign: great idea https://superset.reveal-zm.smartregister.org/superset/slice_json/160?form_data=%7B%22order_by_cols%22:%5B%22%5B%5C%22jurisdiction_depth%5C%22,+true%5D%22,%22%5B%5C%22is_virtual_jurisdiction%5C%22,+true%5D%22,%22%5B%5C%22jurisdiction_name%5C%22,+true%5D%22%5D,%22adhoc_filters%22:%5B%7B%22clause%22:%22WHERE%22,%22expressionType%22:%22SIMPLE%22,%22comparator%22:%229f1e0cfa-5313-49ff-af2c-f7dbf4fbdb9d%22,%22operator%22:%22==%22,%22subject%22:%22plan_id%22%7D%5D,%22row_limit%22:15000%7D

PR #602, opened 2019-11-19
Adds optional overwrite flag to team Assignment fetch action creators

### LINKED ISSUE #595: Deleted data still shows up on UI
_Reproducing_:
- create practitioner on UI,
- delete practitioner using postman or whichever tool
- back on the UI, perform any action that leads to new practitioners being fetched and rendered
- the deleted practitioner is still in store and thus shows up on the ui.

This bug happens due to how fetched response data is fetched from the api, ie. we are currently appending new fetched data to already existing records and thus deleted data in the store is never removed.

p-netm:
as per the issue that this pr closes #595 
this pr adds an overwrite flag to the :
- fetchPractitioners
- fetchOrganizations 
action creators.
now data that is retrieved for the organizationsList and practitionersList pages overwrites whatever data that was there, this ensures we do not retain stale data in the store and thus on the ui,...


src/store/ducks/opensrp/practitioners/index.ts
```diff
@@ -69,11 +70,14 @@ export type PractitionersActionTypes =
 
 /** Fetch practitioners action creator
  * @param {Practitioner []} practitionersList - practitioners array to add to store
+ * @param {boolean} overwrite - whether to replace the records in store for practitioners
  * @return {FetchPractitionersAction} - an action to add practitioners to redux store
  */
 export const fetchPractitioners = (
-  practitionersList: Practitioner[] = []
+  practitionersList: Practitioner[] = [],
+  overwrite: boolean = false
```
cKellyDesign: @p-netm I know I said `false`, but now I'm wondering which case is more common for `overwrite`? If `overwrite` is set to `true` more often then `false` (which it looks like maybe it is?), then perhaps the default should be `true`. I'll leave that decision to you :)

p-netm: Well i would choose false since most of the other fetch action creators outside team assignment append as opposed to overwrite newly fetched records to store.

PR #604, opened 2019-11-20
Issue#186 ensure user is logged out from opensrp

### LINKED ISSUE #186: Ensure superset user is logged out after logging out

FrankApiyo:
Fixes #186


src/components/page/Header/index.tsx
```diff
@@ -255,6 +260,17 @@ export class HeaderComponent extends React.Component<HeaderProps, State> {
     );
   }
 
+  private handleLogout() {
+    const logoutURL: string = 'https://reveal-stage.smartregister.org/opensrp/logout.do';
```
p-netm: lets put this in the env file,

FrankApiyo: we will indeed.

FrankApiyo:
> Am iffy about the sideeffects this imposes on the ui, will have to wait and see what @cKellyDesign thinks of it.

I think we could make the time taken before the window we open in closed automatically longer... then it'll be less weird.

src/components/page/Header/index.tsx
cKellyDesign: > lets put this in the env file

agreed, let's pipe in in from `.env` through `env.ts` and import it here, and let's add it to `.env.sample` too

FrankApiyo:
now that code is tested :ok_hand:

Issue #586, opened 2019-11-11
I18N for Reveal Web


craigappl:
We need to begin the process of scoping internationalization for the Reveal Web App. The scoping document will live on confluence: https://smartregister.atlassian.net/wiki/spaces/REV/pages/1229127681/Web+UI+Internationalization

What's involved (Please add bullets)
- Translate plan creation form so that input on targets and goals can support multiple languages
- Translate all UI elements in the web app (a single properties file)
- Translate task status and business status so that it can be supported
- Develop an architecture for accessing translations from data. Currently, the Android client's form information is stored as English values in the data warehouse. We need to figure out an architecture for referencing the translated value in a localized context. (This is a canopy problem, not a Reveal problem)
- Decision - We need to decide if we want to support switching languages in the UI or if we want to deploy the web UI in a specific language.

cKellyDesign:
@craigappl I've added bullets for how we will implement this in the page linked above.

- [ ] Install https://github.com/i18next/react-i18next in the WebApp
- [ ] Audit the rest of Webapp codebase to make sure all strings are contained in our constants file
- [ ] Update our constants file to translate all strings in the WebApp
- [ ] Add an environment variable to set the language of the WebApp when deployed
- [ ] Check business logic specific strings to make sure our conditionals aren't failing between the translated strings and the strings coming in from OpenSRP/Superset

p-netm:
@cKellyDesign Should we add an audit for concatenated strings, i for one am guilty of having done that  e.g [here](https://github.com/onaio/reveal-frontend/blob/3c3b9406e26d434a1453243c71d3aa2257c4660f/src/containers/pages/FocusInvestigation/map/planCompletion/index.tsx#L117)

cKellyDesign:
@p-netm good point, we should collect these in a central place (maybe not constants) and translate them as sentences instead of individual words. May need to use some simple builder functions which pass in the the dynamic parts into the sentence using the `<Trans />` tags [like this example](https://github.com/i18next/react-i18next#what-will-my-code-look-like).

cKellyDesign:
> Mop-up display strings

- [ ] Gatekeeper components (login process)
- [ ] Add/Edit Plan form priority dropdown (pending translation)
- [ ] Add/Edit Plan Activity Goal Units
- [ ] Plan Status in tables throughout the site
- [ ] FI Plans Table Headers
- [ ] Focus Investigation Manage page Routine and blue button "Add Focus Investigation"
[...]

PR #610, opened 2019-11-21
586 Migrate Display Strings to Constants

cKellyDesign:
This PR addresses the first task in #586 and should probably be merged independently.


cKellyDesign:
@p-netm there are a few places where it may make sense to run the internationalization functions in-file rather than all in constants (e.g. this settings files and some of the reporting helper files). They may end up in constants or we may need / want to re-organize our constants so I left that decision out of this PR.

PR #615, opened 2019-11-25
586 Internationalization

cKellyDesign:
This PR is for: #586 

This PR currently:
* Organizes display strings in the constants file
* Populates i18n english translations for transifex
* Ingest i18n configs as resources
* Migrates all display strings into `lang.ts`
* Translates existing display strings via i18next (from English to English)
* Removes redundant type declarations in the `constants` file


src/configs/strings/en.json
```diff
@@ -0,0 +1,202 @@
+{
```
moshthepitt: @cKellyDesign is this generated by hand?

src/constants.tsx
```diff
@@ -265,68 +265,17 @@ export const ADD = 'Add';
 export type ADD = typeof ADD;
```
moshthepitt: @cKellyDesign I think we should move "display" strings to something like `lang.ts` so as to completely separate them from other constants that should not be translated.

moshthepitt:
@cKellyDesign @p-netm let's have a small discussion:

- It looks like we have chosen [react.i18next](https://react.i18next.com/) - is this correct?
- If so, I think the usual usage of it is to mark translatable strings within React components e.g. something like `<div>{t('simpleContent')}</div>` - do you agree?
- Our approach seems to be that we will have display strings in some module e.g. constants.ts/lang.ts, and then mark these as translatable - is this so?  Could you give an indication of what you prefer, and why? i.e.
   - mark translatable strings in the lang file itself
   - mark translatable strings in the components
- What are your thoughts on testing all this?

p-netm:
@moshthepitt @cKellyDesign 
- yes, we are using react.i18next
- yes, that's how it works, an how i understand it is that for instance in your example `simpleContent` would not act as the content but as key 
- ¯\_(ツ)_/¯ :thinking: :thinking::thinking: 
```typescript
// i18next
import {ORGANIZATION_LABEL} from './constants.ts/lang.ts'

.init({
  resources: {
    en: {
      translations: {
        [ORGANIZATION_LABEL]: 'Team',
      }
    },
    sw: {
      translations: {
        [ORGANIZATION_LABEL]: 'Timu',
      }
    }
  },

  // some component
  import {ORGANIZATION_LABEL} from './constants.ts/lang.ts';
  const app = () => {
    return <div>t(ORGANIZATION_LABEL)</div>
  }

  // constants.ts/lang.ts
  export const ORGANIZATION_LABEL = 'Team';
```
- ¯\_(ツ)_/¯ :thinking:

src/configs/strings/en.json
cKellyDesign: yuuuup copied over by hand and given context for translation, it was a pretty interesting exercise actually and helped weed out unused strings!

cKellyDesign:
@p-netm an update from a discussion w/ @moshthepitt today:
1. Let's try dropping the i18next modules because they work by asynchronously returning translations based on the key, which feels like it could be accomplished synchronously with simple ternary conditionals like so:
```
// lang.ts - all strings get translated
import { LANGUAGE } from 'env';
import * as translations from 'strings';
const library = translations[LANGUAGE];

const WELCOME_TO_REVEAL = library['WELCOME_TO_REVEAL'] || 'Welcome to Reveal';
...
```

2. in an effort to avoid (lightly) refactoring half the repo, it would be nice if all the translations happened outside of React

3. All strings being rendered on the screen are subject to translation, 1 dictates nothing is translated in components. if strings don't have a translation then they will fallback to their original english values. This distinction means they should be defined in a different config (I like `lang.ts`), **my question is should we import/export these translated strings in and out of `constants` so we don't need to touch `lang` directly from the components?**

4. Let me circle back to the topic of tests...

src/constants.tsx
```diff
@@ -1,197 +1,314 @@
...
+export const STRUCTURE_LAYER = 'structure-layer';
+export type STRUCTURE_LAYER = typeof STRUCTURE_LAYER;
+export const POLYGON = 'Polygon';
+export type POLYGON = typeof POLYGON;
+export const MULTI_POLYGON = 'MultiPolygon';
+export type MULTI_POLYGON = typeof MULTI_POLYGON;
+export const POINT = 'Point';
+export type POINT = typeof POINT;
+export const FEATURE = 'Feature';
+export type FEATURE = typeof FEATURE;
```
cKellyDesign: @moshthepitt @p-netm - could we DRY this out by simply exporting all these strings... as strings? 

```
export const JURISDICTION_ID = 'jurisdiction_id';
export type JURISDICTION_ID = typeof JURISDICTION_ID;
```
becomes
```
export const JURISDICTION_ID = 'jurisdiction_id' as string;
```
🤔🤔🤔🤔🤔🤔

moshthepitt: @cKellyDesign yeah lets do this.  You do not even need the "as string" as Typescript will infer the type.

.env.sample
```diff
@@ -1,4 +1,5 @@
 SKIP_PREFLIGHT_CHECK=true
+REACT_APP_LANGUAGE=en
```
cKellyDesign: How the localization config is set, this must have an i18n JSON file in `src/configs/strings` or translations will fall back to English.

src/configs/env.ts
```diff
@@ -4,6 +4,10 @@
 export const WEBSITE_NAME = process.env.REACT_APP_WEBSITE_NAME || 'Reveal';
 export type WEBSITE_NAME = typeof WEBSITE_NAME;
 
+/** The website language */
+export const LANGUAGE = process.env.REACT_APP_LANGUAGE || 'en';
+export type LANGUAGE = typeof LANGUAGE;
```
cKellyDesign: If no localization language is specified the app will use English translated display strings.

src/configs/lang.ts
```diff
@@ -0,0 +1,359 @@
+// import the strings reference i18n file for the specified language
+import { LANGUAGE } from './env';
+// tslint:disable-next-line:no-var-requires
+const reference = require('./strings')[LANGUAGE] || {};
```
cKellyDesign: Import the i18n file for the configured language as `reference`.

src/configs/lang.ts
```diff
@@ -0,0 +1,359 @@
...
+/**
+ * @param {string} key The unique variable name of the display string
+ * @param {string} fallback The (original) English fallback string if reference doesn't include the display string
+ * @returns {string} The translated display string
+ */
+const translate = (key: string, fallback: string): string =>
+  (reference[key] && reference[key].message) || fallback;
```
cKellyDesign: Helper function to return translated display strings from the loaded reference file, falling back to original English strings if no translated display string is provided in the loaded i18n file.

src/containers/pages/FocusInvestigation/map/planCompletion/index.tsx
```diff
@@ -110,24 +108,16 @@ export class PlanCompletion extends React.Component<
     return (
       <div id="plan-completion-page-wrapper">
         <Helmet>
-          <title>{`${MARK} ${plan.plan_title} ${AS_COMPLETE}`}</title>
+          <title>{format(MARK_PLAN_AS_COMPLETE, plan.plan_title)}</title>
```
cKellyDesign: How translated display strings inject variables into placeholders.

src/containers/pages/FocusInvestigation/map/planCompletion/tests/__snapshots__/index.test.tsx.snap
```diff
@@ -25,14 +25,7 @@ exports[`@containers/pages/map/planCompletion/ renders correctly 1`] = `
...
   >
-    Mark 
-    <span
-      className="plan-name"
-    >
-      $
-      A1-Tha Luang Village 1 Focus 01
-    </span>
-     as complete
+    Mark A1-Tha Luang Village 1 Focus 01 as complete
```
cKellyDesign: This implementation does not support display strings with composite HTML.

src/containers/pages/IRS/Map/index.tsx
```diff
@@ -298,7 +298,13 @@ const IRSReportingMap = (props: IRSReportingMapProps & RouteComponentProps<Route
...
                   <p className="indicator-breakdown">
-                    {PROGRESS}: {row.numerator} {OF} {row.denominator} {`${row.unit || STRUCTURE}s`}{' '}
+                    {PROGRESS}:{' '}
+                    {format(
+                      NUMERATOR_OF_DENOMINATOR_UNITS,
+                      row.numerator,
+                      row.denominator,
+                      row.unit || STRUCTURES
+                    )}{' '}
                     ({row.value}%)
```
cKellyDesign: In this case, it felt like compound strings made sense given the nature of the information (`<metadata>:<textual data> (<numerical data>)`) and are translated separately (or not at all).

src/containers/pages/FocusInvestigation/map/planCompletion/tests/__snapshots__/index.test.tsx.snap
moshthepitt: @cKellyDesign what will we do about these?

cKellyDesign: Not too sure unless we actually bring in `react-i18next`.

cKellyDesign:
> the "reason" field for planActivities has missing translations

@moshthepitt not sure where this is missing from, could you link one of the other activity fields that's not missing?

> should change RACD_ACTIVITY to "REGISTER FAMILY ACTIVITY"

Done and updated in the [translation sheet](https://docs.google.com/spreadsheets/d/1ONHpPpuPxNW9_-vP8gTTcrVHTbKTT3WaN-cfdySeMCs/edit?folder=1MN5J-PK6KUOHC6gw3KKmf12vV-GQFPkr#gid=0).

> we should add a test that ensures that there are no untranslated strings in any component

Interesting, not sure how we would do this 🤔, any ideas? May just need to create a ticket for it.

moshthepitt:
@cKellyDesign 

> not sure where this is missing from, could you link one of the other activity fields that's not missing?

Like here https://github.com/onaio/reveal-frontend/blob/1f1968281f2acd623a78299dc52c8fe1e02f3ef9/src/configs/settings.ts#L296 and https://github.com/onaio/reveal-frontend/blob/1f1968281f2acd623a78299dc52c8fe1e02f3ef9/src/configs/settings.ts#L310

moshthepitt:
@cKellyDesign we also need to test the PlanForm with translation turned on to ensure that while the drop-down fields are translated, they are always saved (sent to OpenSRP) in English.  I think this is the most sensitive part of this PR, and the one which could break things in the dashboards, Canopy, NiFi and maybe elsewhere that depend on these values.

cKellyDesign:
> one which could break things in the dashboards, Canopy, NiFi and maybe elsewhere that depend on these values

@moshthepitt - then the values should not be saved as translated, like we need to decouple what is being rendered from what is being saved. In these settings it's not immediately clear what is display text vs what effects business logic.

cKellyDesign:
@moshthepitt In this schema, which values need to be decoupled?
```
IRS: {
    action: {
      code: 'IRS', // business or display? 
      description: IRS_ACTIVITY_DESCRIPTION, 
      goalId: 'IRS', // business or display?
      identifier: '',
      prefix: 7,
      reason: ROUTINE, // business or display?
      subjectCodableConcept: {
        text: 'Residential_Structure', // seems like just business
      },
      taskTemplate: 'Spray_Structures',  // seems like just business
      timingPeriod: {
        end: '',
        start: '',
      },
      title: IRS_ACTIVITY,
    },
    goal: {
      description: IRS_GOAL_DESCRIPTION,
      id: 'IRS', // business or display?
      priority: MEDIUM_PRIORITY, // business only
      target: [
        {
          detail: {
            detailQuantity: {
              comparator: '>=',
              unit: GoalUnit.PERCENT,
              value: 90,
            },
          },
          due: '',
          measure: IRS_GOAL_MEASURE,
        },
      ],
    },
  },
```

moshthepitt:
@cKellyDesign I have highlighted below the fields that **have to be translated but have to be saved in English**:

Update Plan  A1 - Ban Huai Haeng OA - 2019-09-09  deprecated    Reveal Stage

These are:

- Plan Intervention type
- Plan status (not shown in screenshot above)
- Focus investigation status
- Focus investigation reason
- The Goal Unit
- The Goal priority
- The Action Reason

moshthepitt:
@cKellyDesign these are fields that are "business only" i.e. they dont need to be translated, but if they are they must be saved in English:

1. action code
2. action goalId
3. goal id

---

Which brings us to:

```
IRS: {
    action: {
      code: 'IRS', // BUSINESS ONLY
      description: IRS_ACTIVITY_DESCRIPTION,
      goalId: 'IRS', // BUSINESS ONLY
      identifier: '',
      prefix: 7,
      reason: ROUTINE, // BOTH BUSINESS & DISPLAY
      subjectCodableConcept: {
        text: 'Residential_Structure', // BUSINESS ONLY
      },
      taskTemplate: 'Spray_Structures',  // BUSINESS ONLY
      timingPeriod: {
        end: '',
        start: '',
      },
      title: IRS_ACTIVITY,
    },
    goal: {
      description: IRS_GOAL_DESCRIPTION,
      id: 'IRS', // BUSINESS ONLY
      priority: MEDIUM_PRIORITY, // BOTH BUSINESS & DISPLAY
      target: [
        {
          detail: {
            detailQuantity: {
              comparator: '>=',
              unit: GoalUnit.PERCENT,  // BOTH BUSINESS & DISPLAY
              value: 90,
            },
          },
          due: '',
          measure: IRS_GOAL_MEASURE,
        },
      ],
    },
  },
```

moshthepitt:
@cKellyDesign for the select/dropdown fields that are both business and display I think it may be easy to do something like 

```jsx
<select>
  <option value={{ BUSINESS VALUE }}>{{ DISPLAY VALUE }}</option>
</select>
```

cKellyDesign:
Awesome thanks @moshthepitt! And yeah the drop downs should be easy, and I'll just add a bit of logic to deal with the goal unit.

cKellyDesign:
Screen Shot 2020-01-09 at 11 54 39 AM

@moshthepitt @p-netm any idea why these tests are failing? Did some packages get minified or something? Not sure what changed here.

src/components/LinkAsButton/test/__snapshots__/index.test.tsx.snap
```diff
@@ -3,7 +3,7 @@
 exports[`components/LinkToNewPlans Matches snapshot 1`] = `
 <LinkAsButton
   classNameProp="focus-investigation btn btn-primary float-right mt-5"
-  text="Add Focus Investigation"
+  text="Add Plan"
```
moshthepitt: @cKellyDesign should we expect this change and the one below?

src/components/Table/TableHeaders/tests/__snapshots__/index.test.tsx.snap
```diff
@@ -167,7 +167,7 @@ exports[`components/page/Header renders routine headers correctly 1`] = `
             >
               <LinkAsButton
                 classNameProp="focus-investigation btn btn-primary float-right mt-5"
-                text="Add Focus Investigation"
+                text="Add Plan"
```
moshthepitt: @cKellyDesign I have asked about this change elsewhere so this comment is just to flag it for me.

src/containers/pages/IRS/Map/tests/__snapshots__/index.test.tsx.snap
```diff
@@ -57,14 +42,9 @@ Array [
...
-    : 
-    19
-     
-    of
-     
-    22
+    :
      
-    rooms
+    19 of 22 room
```
moshthepitt: It seems like we lost the plural form of room @cKellyDesign

src/components/LinkAsButton/test/__snapshots__/index.test.tsx.snap
cKellyDesign: I think probably depends on the context the LinkAsButton is embedded in, checking now

moshthepitt: It seems there are a bunch of places where it changed like this.

cKellyDesign: Yeah, I think I will change it back and open a ticket to update this b/c it feels like the default props for such a generic component should actually be more generic than `Add Plan` and shouldn't be breaking other, more specific tests where the component is being used.

Good catch, thanks @moshthepitt!

src/containers/pages/IRS/Map/tests/__snapshots__/index.test.tsx.snap
cKellyDesign: changed to `room(s)`

src/components/LinkAsButton/test/__snapshots__/index.test.tsx.snap
cKellyDesign: Issue created at https://github.com/onaio/reveal-frontend/issues/634

cKellyDesign:
This should be ready to merge and tag `v0.3.0`

PR #618, opened 2019-11-26
Fix for mark plan as complete error

### LINKED ISSUE #616: Mark Plan as complete errors out
issue first raised [here](https://smartregister.atlassian.net/browse/RVL-699)

Clicking confirm on the plan's mark as complete initiates a PUT fetch api call to `rest/plans` that errors out with `500` status.

p-netm:
closes #616

Refactors type of received response during 'get' request to 'rest/plans/<identifier>, Fixes the payload to PUT accordingly.


src/containers/pages/FocusInvestigation/map/planCompletion/index.tsx
```diff
@@ -69,7 +69,8 @@ export class PlanCompletion extends React.Component<
     const service = new serviceClass(`${OPENSRP_PLANS}`);
 
     // get the plan as it exists in the OpenSRP Server
-    await service.read(planToconfirm.plan_id).then((planPayload: PlanPayload) => {
+    await service.read(planToconfirm.plan_id).then((fullPayload: PlanPayload[]) => {
+      const planPayload = fullPayload[0];
```
moshthepitt: @p-netm why do we need to make this change?  Has the OpenSRP response changed, or did this comment just never work before (it did).

p-netm: @moshthepitt Before the `planPayload` was assumed to be a single object but on rechecking, it seems like even for a single plan, the response is actually a jsonArray, 
I have no idea why it worked, if it did work, 
maybe what i can do is add a check to make it more robust such that it will handle both json Objects and json array,

moshthepitt: @p-netm cool thanks.  I think we should not do that workaround.

src/containers/pages/FocusInvestigation/map/planCompletion/index.tsx
```diff
@@ -87,11 +90,11 @@ export class PlanCompletion extends React.Component<
           })
           .catch(e => {
             // catch the error if one is returned from OpenSRP
-            alert(PLAN_STATUS_UPDATE_ERROR);
+            growl(`{PLAN_STATUS_UPDATE_ERROR}: ${e.message}`, { type: toast.TYPE.ERROR });
```
moshthepitt: @p-netm how can we test these grown notifications?

p-netm: investigating this,

 currently i don't have a good way of observing how they behave from within a test. 

What i tried doing was using `querySelector` on the document.
While this might capture the toast, the toast does not seem to behave as expected in terms of auto closing, The library itself uses redux action to force the toast to auto close,
I tried using jest timers but those did not help,
I also tried simulating clicks on the toast's close button from within the test , but this only appended css classes for the auto close transitions, but the toast will still be there on follow up `document.querySelector`.

moshthepitt: Is it possible to mock the growl function and test it that way?

p-netm: this i can do

src/containers/forms/PractitionerForm/UserIdSelect/tests/index.test.tsx
```diff
@@ -128,6 +128,22 @@ describe('src/*/forms/userIdSelect', () => {
       ],
     ]);
   });
+
+  it('options are sorted in descending', async () => {
```
p-netm: @moshthepitt Tests for sorted openMRSUsers

Issue #156, opened 2019-07-01
Write a spec for more secure token store mechanism


moshthepitt:
Following the discussion in https://github.com/onaio/reveal-frontend/issues/39, we now need to write a specification document that covers:

- The intermediary service that will handle oAuth2 access tokens aka "front end proxy" 
- The reveal backend, if any

moshthepitt:
ORY Hydra looks really interesting: https://github.com/ory/hydra

Oathkeeper as well: https://github.com/ory/oathkeeper

moshthepitt:
Setting up Express with Typescript: https://developer.okta.com/blog/2018/11/15/node-express-typescript

moshthepitt:
Probably related in that we may use that structure: https://github.com/OpenSRP/opensrp-web/issues/7

moshthepitt:
#### Steps to implement a minimal backend server to handle authentication/authorization

- Set up a tiny express.js backend server
- Switch to doing the OAuth 2 authentication/authorization on the server side using the [authorization code OAth2 flow](https://oauth.net/2/grant-types/authorization-code/)
- Store the OAth2 credentials on the server side
- Implement server side Redux store - https://redux.js.org/recipes/server-rendering
- Initialize the Redux store from the server side and set the initial state to contain the OAth2 credentials stored in the backend (if any)
- In theory it should now just work, and one should be logged in for however long the OAuth credentials stored in the backend are valid

cKellyDesign:
@moshthepitt Looks solid!

cKellyDesign:
> Switch to doing the OAuth 2 authentication/authorization on the server side using the authorization code OAth2 flow

@moshthepitt how are we doing AuthZ for non-OpenSRP resources? Webapp views? (Auth'ing more than OpenSRP resources may be outside the scope of this ticket)

moshthepitt:
> how are we doing AuthZ for non-OpenSRP resources? Webapp views? (Auth'ing more than OpenSRP resources may be outside the scope of this ticket)

@cKellyDesign the plan is for nothing else to change for now: just how we get the OAuth credentials.  Everything else will come in a future PR.

Issue #117, opened 2019-06-19
Redirect user after login


moshthepitt:
We should add a way to redirect users to the page they were trying to access before they logged in.

cKellyDesign:
@moshthepitt do we have a proposed solution / level of effort for this?

moshthepitt:
This is very very low on my priority list.  It would be easiest to implement after #156

PR #624, opened 2019-12-13
Tiny backend for OAuth

moshthepitt:
This PR implements a tiny express backend that persists OAuth authentication credentials and pre-loads them into the redux store.

Fixes: #156
Fixes: #117

### How it works

- OAuth is handled via the express backend (we switch to the authorization code flow)
- Once the credentials are received, they are persisited in the backend server
- The backend server injects the persisted state in the window.__PRELOADED_STATE__ object
- This is picked up by the client side code and hydrated

### TODO

- Cleanup the code (it is possible to make the server into an independent module)
- Right now the state is persisted to file - we need to evaluate if this is good enough, but it probably is not
- Implement some mechanism to force refreshing the OAuth credentials after they have expired
- Implement a logout mechanism

### How to test

- run `yarn build`
- set these env vars REACT_APP_OPENSRP_CLIENT_ID and REACT_APP_OPENSRP_CLIENT_SECRET (they are not picked from the .env file)
- run `yarn babel-node server/server.ts --extensions ".ts"`


moshthepitt:
@cKellyDesign @craigappl 

### Updated TODO

1.  I think feeding in the OAuth credentials via preloaded state does not work as this remains in the HTML served and its absolutely unacceptable
2. We therefore need to make it available to the frontend in another way.  We can do this with secure "httponly" cookies, I think.  And have the client side code use the value of the cookie as the source of preloaded state.  
3.  We need to modify the login page to show urls that start the authorization OAuth flow - [example here](https://github.com/onaio/reveal-frontend/pull/624/files#diff-d0088817ab06a53f242d8f267f2945dcR44).  We can achieve this by providing a custom Login component to render on the login page (which itself can be done by either modifying gatekeeper to support it, or including it in the server package - see 6 below)
4. We can drop using gatekeeper from the client side and only use it in the backend.  This means we need to find a place to redirect to (on the client side) from the [server side callback](https://github.com/onaio/reveal-frontend/pull/624/files#diff-d0088817ab06a53f242d8f267f2945dcR48)
5. We need to modify server.ts to persist the OAuth credentials aka preloaded state in a database like redis or postgres.  Right now it persists to file, which is nice but can be better
6. (nice to have) We need to package server.ts in a re-usable way and probably move it to js-tools

cKellyDesign:
> And have the client side code use the value of the cookie as the source of preloaded state

@moshthepitt - To my knowledge client side apps cannot programmatically access the values of `httponly` cookies, and that such cookies are automatically used by the browser when making requests to the domain(s) specified by the cookie(/response headers?/server-side route-handler?).

p-netm:
@moshthepitt @cKellyDesign Is it possible to make the api calls from the server. Not sure though how the web app requests would be mapped to the opensrp service, maybe have a single express endpoint that takes in an object representing a request, the server would apply the token and make an official request to opensrp. As for the response!!? guess it would just have to go through back the way the request was made.

moshthepitt:
> To my knowledge client side apps cannot programmatically access the values of httponly cookies, and that such cookies are automatically used by the browser when making requests to the domain(s) specified by the cookie(/response headers?/server-side route-handler?). 

@cKellyDesign my bad, you are right.  Cookies cannot help here, we'll have to get the credentials to the client another way.

moshthepitt:
> @moshthepitt @cKellyDesign Is it possible to make the api calls from the server. Not sure though how the web app requests would be mapped to the opensrp service, maybe have a single express endpoint that takes in an object representing a request, the server would apply the token and make an official request to opensrp. As for the response!!? guess it would just have to go through back the way the request was made.

@p-netm @cKellyDesign so this is what we want to do eventually, based on our discussions last year.  However, I want to go towards it slowly so that's why this PR does not at all include anything that goes in the direction of using the backend as a proxy.

moshthepitt:
I have been working on this, albeit slowly.  The current plan:

1.  OAuth will completely be handled by the backed server.  We will be switching to the authorization code flow.
2. The backend server will store the credentials and make them accessible via a JSON view, something like `/api/oauth-state`
3. The backend server will be responsible for serving both its views and all other views (i.e. it will serve the requests made to the client, by passing them to the client).
4. We'll change the client to look for the credentials by accessing `/api/oauth-state`, and use that to populate the relevant redux stores
5. The server will maintain a session for authenticated users `/api/oauth-state` will only be available for authenticated users.
6. In the event that a user refreshes the browser while logged in, we will ensure that the client:
    - recognizes the user is logged in
    - fetches the OAuth credentials and puts them into its redux store
    - redirects the user to the url that the user had wanted to access
7. This means that the gatekeeper package will need modifications, these are being done here https://github.com/onaio/js-tools/pull/114

---

Please let me know if you think this makes sense, or if I have overlooked anything @cKellyDesign @p-netm?

moshthepitt:
@p-netm this is what you need to run the server for now:

- Yarn install
- Create a .env file in the server directory
- Run `yarn build` so that you compile the front end (client)
- Run the express server with `yarn babel-node -r node_modules/dotenv/config server/server.ts --extensions ".ts" dotenv_config_path=/home/mosh/ona/reveal-frontend/server/.env`
- These three views are exposed:
   - `/oauth/opensrp` to initiate server side login
  - `/oauth/callback/OpenSRP` server side OAuth callback
  - `/oauth/state` renders the OAuth state JSON
  - Everything else is handled by the client (via the compiled client code from step 3)

cKellyDesign:
@moshthepitt @p-netm - I'm trying to run this locally and it looks like I'm still missing something, I've gotten env vars from the client env and [from slack](https://onaio.slack.com/archives/CBSU8L2SC/p1576265426012800?thread_ts=1576265407.012500&cid=CBSU8L2SC). Any ideas what's not happening?

Screen Shot 2020-01-27 at 12 28 23 PM

moshthepitt:
> Can we add this to the package.json scripts?

@p-netm ^^

moshthepitt:
> @moshthepitt @p-netm - I'm trying to run this locally and it looks like I'm still missing something, I've gotten env vars from the client env and from slack. Any ideas what's not happening?

Not sure but one thing I noticed was that state in the OAuth2 options was unidentified.  Are you using something like this for envs?

```
EXPRESS_OPENSRP_ACCESS_TOKEN_URL=https://reveal-stage.smartregister.org/opensrp/oauth/token
EXPRESS_OPENSRP_AUTHORIZATION_URL=https://reveal-stage.smartregister.org/opensrp/oauth/authorize
EXPRESS_OPENSRP_CALLBACK_URL=http://localhost:3000/oauth/callback/OpenSRP/
EXPRESS_OPENSRP_USER_URL=https://reveal-stage.smartregister.org/opensrp/user-details
EXPRESS_OPENSRP_OAUTH_STATE=opensrp
EXPRESS_OPENSRP_CLIENT_ID=???
EXPRESS_OPENSRP_CLIENT_SECRET=???
EXPRESS_PRELOADED_STATE_FILE=/tmp/revealState.json
```

cKellyDesign:
Yup looks like I was missing a few of those, will try again, thanks!

moshthepitt:
@cKellyDesign I made it just a bit more streamlined and hopefully easier to deal with: https://github.com/onaio/reveal-frontend/pull/644

cKellyDesign:
Great thanks @moshthepitt I will check it out! Also at some point it would be good to have a flow diagram describing how this all works, especially as we look to standardize and modularize this implementation for other projects :D

moshthepitt:
> Great thanks @moshthepitt I will check it out! Also at some point it would be good to have a flow diagram describing how this all works, especially as we look to standardize and modularize this implementation for other projects :D

@p-netm I think you understand it well enough to to draw this up?  You can add it to some README in the server directory - can you create an issue for this?

moshthepitt:
> @moshthepitt can we also change this line to node_modules/ to ignore node_modules directories anywhere in the repo?

@cKellyDesign done!

moshthepitt:
@p-netm sometimes getting this error:

```{
message: "ENOENT: no such file or directory, stat '/home/reveal-web-stage/express-versioned/build/index.html'",
status: "error",
statusCode: 404
}```

cKellyDesign:
@p-netm @moshthepitt are there updated steps to set this up locally? Or should i refer to existing comments again?

cKellyDesign:
@moshthepitt @p-netm - It would be really nice to see a flow-diagram describing the authentication and logout flows implemented by this PR. Also any implementation docs would be good, but maybe these belong in the Gatekeeper package itself?

moshthepitt:
> It would be really nice to see a flow-diagram describing the authentication and logout flows implemented by this PR. Also any implementation docs would be good, but maybe these belong in the Gatekeeper package itself?

@p-netm will create this sometimes next week

cKellyDesign:
(removed `has pr` label because this is a PR, and added label to the connected issues)

PR #625, opened 2019-12-16
Add AbortSignal to OpenSRPServer service

### LINKED ISSUE #356: Fix UpdatePlan bug
```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
    in PlanDefinitionList (created by Context.Consumer)
    in Connect(PlanDefinitionList) (created by Route)
```

### LINKED ISSUE #554: Refactor setLoading usage
currently, some components use a setState call to track if the component should load. i.e
```typescript
    const SomeComponent = (props) => {
    const [loading, setLoading] = useState<boolean>(props.someArrayProps.length > 1)

    if(loading){
        return <Loading />
    }
    
    return <h1>Loaded</h1>
}
```
Where `loading` is dependent on the presence of records inside `someArayProp`.

__Problem__:

Possibility for a memory leak resulting from setting state on an unmounted component 
__how__: 
if say there is an asynchronous call to get `someArrayProp` data with a callback that sets loading to false, when the component gets unmounted before such the async data loading call resolves, the setState call still runs effectively trying to set state of  an unmounted component. 

__Proposed solution__:

`loading` can be derived from the props and does not have to be tracked in the state

```typescript
    const SomeComponent = (props) => {

    // data loading async call

    const loading = props.someArrayProps.length > 1

    if(loading){
        return <Loading />
    }
    
    return <h1>Loaded</h1>
}
```
Here the take away being that we don't have to use `setState` hooks for variables that can be derived from data that is already present either in state or in the props

p-netm:
closes #554, close #356 
- Refactors the openSRPService to include an abort signal that can be used to abort pending Promises.
- Adds growl to a few components.
- <del>Add cleanup functions to Functional components that aborts fetch calls to opensrp</del> reason : 16eba72


src/containers/forms/OrganizationSelect/index.tsx
```diff
@@ -77,18 +79,26 @@ const OrganizationSelect = (props: OrganizationSelectProps) => {
...
-        /** TODO - find something to do with error */
+        growl(err.message, { type: toast.TYPE.ERROR });
       });
   };
 
   useEffect(() => {
-    loadOrganizations(serviceClass);
+    loadOrganizations(serviceClass, signal);
+    return () => controller.abort();
   }, []);
```
cKellyDesign: Nice! Good use of `useEffect`'s cleanup callback!

src/containers/pages/IRS/assignments/index.tsx
```diff
@@ -79,19 +83,19 @@ const IRSAssignmentPlansList = (props: PlanAssignmentsListProps) => {
...
 
   useEffect(() => {
     loadData();
+    return () => controller.abort();
   }, []);
 
-  if (loading) {
+  if (plans.length < 1) {
     return <Loading />;
   }
```
cKellyDesign: Feels like this should be an if-else-if block so we don't render the spinner when we're done waiting for the response.
```
if (loading) {
  return <Loading />
} else if (plans.length < 1) {
  return <p>Sorry, no plans were found</p>
}
```
(without magical strings of course!)

src/containers/pages/InterventionPlan/PlanDefinitionList/index.tsx
```diff
@@ -65,21 +68,19 @@ const PlanDefinitionList = (props: PlanListProps) => {
...
 
   useEffect(() => {
     loadData();
+    return () => controller.abort();
   }, []);
 
-  if (loading === true) {
+  if (plans.length < 1) {
     return <Loading />;
   }
```
cKellyDesign: Same as above

src/containers/pages/InterventionPlan/UpdatePlan/index.tsx
```diff
@@ -54,25 +57,23 @@ const UpdatePlan = (props: RouteComponentProps<RouteParams> & UpdatePlanProps) =
...
 
   useEffect(() => {
     loadData();
+    return () => controller.abort();
   }, []);
 
-  if (loading === true) {
+  if (plan === null) {
     return <Loading />;
   }
```
cKellyDesign: Same as above, may need a message like "Plan not found" or something

src/containers/pages/PractitionerViews/PractitionerListView/index.tsx
```diff
@@ -118,7 +119,10 @@ const PractitionersListView = (props: PropsTypes) => {
 
   /** hook to load all practitioners and dispatch to them to store */
   useEffect(() => {
-    loadPractitioners(serviceClass, fetchPractitionersCreator);
+    loadPractitioners(serviceClass, fetchPractitionersCreator, signal);
+    return () => {
+      controller.abort();
+    };
```
cKellyDesign: Might it be DRY'er to have a util function that does this instead of re-writing the anon function?
```
const controllerAborter = (controller) => { controller.abort() };
...
useEffect(() => {
  ...
  return controllerAborter(controller);
}, []);
```
(but with a better name than `controllerAborter`?)

src/containers/pages/IRS/assignments/index.tsx
p-netm: I agree that would be ideal, however for this component, loading is a "computed variable" of `plans.length`.

src/containers/pages/InterventionPlan/PlanDefinitionList/index.tsx
p-netm: `loading` state variable is computed from `plans.length` here too

src/containers/pages/PractitionerViews/PractitionerListView/index.tsx
p-netm: hmmm... maybe I could try and see if a custom hook will serve us better here.

src/containers/pages/IRS/assignments/index.tsx
p-netm: #798

src/containers/pages/InterventionPlan/UpdatePlan/index.tsx
p-netm: Created #798, to help guid discussions and work for this

src/containers/pages/InterventionPlan/UpdatePlan/index.tsx
```diff
@@ -56,22 +55,19 @@ const UpdatePlan = (props: RouteComponentProps<RouteParams> & UpdatePlanProps) =
...
-      setLoading(false);
     }
   }
 
   useEffect(() => {
     loadData().catch(err => displayError(err));
   }, []);
 
-  if (loading === true) {
+  if (plan === null) {
```
moshthepitt, line 70: why this change?

src/containers/pages/OrganizationViews/helpers/tests/servicehooks.test.ts
```diff
@@ -47,9 +47,7 @@ describe('src/containers/pages/OrganizationViews/helpers/servicehooks', () => {
       mockClass,
       fetchPractitionerRolesMock,
       fetchPractitionersMock
-    ).catch(e => {
-      throw e;
-    });
+    ).catch(err => err);
```
moshthepitt, line 50: Do we want to use `displayError` here?

src/components/forms/PractitionerForm/UserIdSelect/index.tsx
```diff
@@ -96,8 +98,10 @@ export const UserIdSelect: React.FC<Props> = props => {
   const loadUnmatchedUsers = async () => {
     const practitioners: Practitioner[] = await new props.serviceClass(
       OPENSRP_PRACTITIONER_ENDPOINT
-    ).list();
-    const allOpenMRSUsers = await loadOpenMRSUsers();
+    )
+      .list()
+      .catch((err: Error) => displayError(err));
```
moshthepitt, line 103: Do we want to add a test that `displayError` is, in fact, called?

src/containers/pages/IRS/assignments/index.tsx
```diff
@@ -99,7 +98,7 @@ const IRSAssignmentPlansList = (props: PlanAssignmentsListProps) => {
     loadData().catch(error => displayError(error));
   }, []);
 
-  if (loading) {
+  if (plans.length < 1) {
```
moshthepitt, line 101: If we get 0 plans from the API, does this mean we will show the loading indicator forever?

src/containers/pages/InterventionPlan/PlanDefinitionList/index.tsx
```diff
@@ -63,21 +62,18 @@ const PlanDefinitionList = (props: PlanListProps) => {
...
-      setLoading(false);
     }
   }
 
   useEffect(() => {
     loadData().catch(err => displayError(err));
   }, []);
 
-  if (loading === true) {
+  if (plans.length < 1) {
```
moshthepitt, line 76: If we get 0 plans from the API, does this mean we will show the loading indicator forever?

src/containers/pages/IRS/assignments/index.tsx
p-netm: unfortunately yes, the alternative is to start implementing the changes that arise from the discussions here https://github.com/onaio/reveal-frontend/issues/806. This change only attempts to stop with setting state error, but still retains the same logic for when the loader is going to be displayed.

src/containers/pages/OrganizationViews/helpers/tests/servicehooks.test.ts
p-netm: Naah!, this is part of the test, I want the error to actually fail the test.

src/containers/pages/IRS/assignments/index.tsx
p-netm: i was thinking about this wrong. let me refactor.

moshthepitt:
@p-netm can we add notes on how to proceed, and then possibly close this?

moshthepitt:
Going to close this.

Issue #641, opened 2020-01-27
Stop un-mounting containers when GA Tracking


cKellyDesign:
Our current implementation react-ga's page tracking follows suggested practices, but is forcing our container components (and all of their children components) to unmount and remount every time the URL changes.

### Implementation Options
A. Tweak [current implementation](https://github.com/onaio/reveal-frontend/blob/master/src/components/page/WithGATracker/index.tsx#L51-L89) to avoid unmounting containers, or
B. Refactor to use a [different suggested implementation](https://vanja.gavric.org/blog/integrate-google-analytics-with-react-router-v4/#routecomponentimhobestapproach)

Additionally:
C. If a GA code is not provided via env variables then do not initialize GA tracking.

moshthepitt:
This issue's successful resolution must be accompanied by https://github.com/onaio/js-tools/issues/113

kelvin-muchiri:
@moshthepitt @cKellyDesign I don't seem to understand or know how to produce this. I tried but seem to find no problem. What steps should I use?

cKellyDesign:
@kelvin-muchiri do you have `GA_CODE` included in your `env` variables?

### Reproduction Steps
1. run locally with GA enabled
2. go to `Focus Investigation` > [any FI Plan] (at this point the map is mounted and loaded)
3. click a metric in the sidebar to update the url (seeing the map re-initialize indicates re-mounting)

Issue #637, opened 2020-01-24
FI map breaks in Thailand production


moshthepitt:
Reveal web version: 0.3.0

The map view in the monitor page does not load for Family Registration and Bednet Indicators (it loads for the other indicators)

![image (5)](https://user-images.githubusercontent.com/372073/73060786-aeb95c80-3ea9-11ea-9d4f-105e0a9fe82b.png)

moshthepitt:
You can reproduce by trying to load this plan: `A1 Thailand test site BVBD 2 2020-01-09 RK release 3.1.2 การสอบสวน` on Thailand production

moshthepitt:
It seems to work fine locally (but tested in English)

moshthepitt:
@cKellyDesign note from Pierre:

> I can confirm that v0.2.9.2 works. To note - that version does not reload the map view, while the newer version does

(he means reload the map when you click on a goal)

cKellyDesign:
@moshthepitt @craigappl Ok so there are a few things going on and a few things that are going to be very very difficult (if not impossible) to test in staging.

#### A. GA is forcing our components to re-mount
Our current implementation seems to be forcing our container components to unmount and remount, forcing every component on the page to unmount and remount. Ideally this should amount only to a performance issue as the components should have all the information they need to render correctly.

#### B. GisidaWrapper is not re-initializing the map after remounting 
This is the original issue raised above, and so far it is only in TH Prod v0.3.0. I have yet to reproduce this issue locally or see it live, and without reproducing the issue it will be very difficult to actually debug.

#### C. The reported plan has unexpected geometries (points) for structures
I'm not sure how related this is, but the geometries being returned from the structures slice in Superset is returning Points instead of Polygons, which the Mapbox map will not render because we are only rendering fill and line layers for structures. This should be a simple fix, but it will be hard to test locally b/c I don't know if there are any plan-jurisdictions in staging with points for structures.

### Proposed Next Steps
1. Add a circle layer to render structures that are points to resolve C. I don't know if this will effect B (and it won't effect A), but implementing this change would either rule it out or confirm it as the underlying issue for B. 
_Note: Unless we have a plan-jurisdiction with points as structures, we won't be able to test until we release to production_

2. Find or make a new plan in staging which actually reproduces B, then debug from there.
3. Update the GA functionality to only use the GA HOC component if the GA code is provided as `v0.3.1` (so we can avoid A while we...)
[...]

PR #642, opened 2020-01-27
Do not use GA HOC if GA code is not provided

cKellyDesign:
This PR addresses:
- [x] Point C from https://github.com/onaio/reveal-frontend/issues/641
- [x] Point 3 from https://github.com/onaio/reveal-frontend/issues/637#issuecomment-578252808
- [ ] Default to `test` if `REACT_APP_GA_ENV` is not provided
- [ ] Adds tests

After Merging:
- [ ] Include in release `v0.3.1`
- [ ] Deploy `v0.3.1` to Thailand Prod without a GA code


src/components/page/WithGATracker/tests/index.test.tsx
```diff
@@ -78,5 +78,22 @@ describe('components/WithGATracker', () => {
...
+      <Provider store={store}>
+        <ConnectedRouter history={history}>
+          <App />
+        </ConnectedRouter>
+      </Provider>
+    );
+    history.push(PLAN_LIST_URL);
+    expect(GoogleAnalytics.pageview).toBeCalledTimes(0);
+    wrapper.unmount();
   });
```
cKellyDesign: @moshthepitt @p-netm @FrankApiyo - What's the best way to set an env variable for a specific test?

The `GA_CODE` from `env.ts` is being imported into from `WithGATracker` and needs to be tested with a `GA_CODE` value of `''`.

moshthepitt: @cKellyDesign we typically use this file to mock them https://github.com/onaio/reveal-frontend/blob/master/src/configs/__mocks__/env.ts

moshthepitt: Here is an example usage in a test https://github.com/onaio/reveal-frontend/blob/b8e46731601574489ba9a4d1a679ec47885fc3a5/src/containers/forms/OrganizationSelect/tests/index.test.tsx#L16

cKellyDesign: yeah the `jest.mock()` is there on line 86, I thought the mock file was up-to-date but perhaps it wasn't.

What I'm wondering then is how to test both cases of:
1. the `GA_CODE` is provided via env and so mounts and page tracks as normal
2. the `GA_CODE` is not provided via env and so GA tracking doesn't happen

moshthepitt: @cKellyDesign I think we would need to have two test files: one that mocks the envs and one that does not

cKellyDesign: aaah ok that's a great idea, i'll do that today, thanks @moshthepitt!

src/components/page/WithGATracker/tests/withoutGA.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+      </Provider>
+    );
+    history.push(PLAN_LIST_URL);
+
+    // even though this env variable is set up correctly for the test,
+    // env.ts is picking up the local process.env.REACT_APP_GA_CODE,
+    // so by the time WithGACode imports GA_CODE it's already set from local process.env
+    expect(process.env.REACT_APP_GA_CODE).toBe('');
+    expect(GA_CODE).toBe('');
+    expect(GoogleAnalytics.pageview).toBeCalledTimes(0);
```
cKellyDesign: Despite moving into a file that does not mock `env.ts`, the test is still picking up the local env vars (`process.env.REACT_APP_GA_CODE` => `env.ts` => `WithGATracker` => `WitoutGATracker.test.ts`).

```
expect(process.env.REACT_APP_GA_CODE).toBe(''); // passes because it is reset before the test
expect(GA_CODE).toBe(''); // fails with the local process.env.REACT_APP_GA_CODE
expect(GoogleAnalytics.pageview).toBeCalledTimes(0); // fails because it is called twice
```

src/components/page/WithGATracker/tests/withoutGA.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+  it('renders components correctly without HOC wrapper', () => {
+    const wrapper = mount(
+      <Provider store={store}>
+        <ConnectedRouter history={history}>
+          <App />
+        </ConnectedRouter>
+      </Provider>
+    );
+    expect(wrapper.find('App').length).toEqual(1);
+    expect(wrapper.find('HOC').length).toEqual(0);
```
cKellyDesign: This fails because HOC is being mounted and GA Tracking is being initialized with the locally set `process.env.REACT_APP_GA_CODE`.

moshthepitt:
Seems like its superseded by https://github.com/onaio/reveal-frontend/pull/657

PR #643, opened 2020-01-27
Add default circle layer for FI structures point features

cKellyDesign:
This PR:
- [x] Adds a default structures circle layer (addressing point C from https://github.com/onaio/reveal-frontend/issues/637#issuecomment-578252808)
- [ ] Adds tests

After Merging:
- [ ] Include in release `v0.3.1`
- [ ] Deploy `v0.3.1` to Thailand Prod without a GA code

### DIFF
--- a/src/components/GisidaWrapper/index.tsx
+++ b/src/components/GisidaWrapper/index.tsx
@@ -14,8 +14,7 @@ import {
   lineLayerConfig,
   symbolLayerConfig,
 } from '../../configs/settings';
-import { APP, MAIN_PLAN, MAP_ID, STRUCTURE_LAYER } from '../../constants';
-import { displayError } from '../../helpers/errors';
+import { APP, MAIN_PLAN, MAP_ID, POINT, STRUCTURE_LAYER } from '../../constants';
 import { EventData } from '../../helpers/mapbox';
 import { ConfigStore, FeatureCollection, FlexObject } from '../../helpers/utils';
 import store from '../../store';
@@ -24,6 +23,7 @@ import { Jurisdiction, JurisdictionGeoJSON } from '../../store/ducks/jurisdictio
 import { StructureGeoJSON } from '../../store/ducks/structures';
 import { TaskGeoJSON } from '../../store/ducks/tasks';
 import './gisida.css';
+import { displayError } from '../../helpers/errors';
 
 /** handlers Interface */
 export interface Handlers {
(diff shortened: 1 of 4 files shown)


src/components/GisidaWrapper/index.tsx
```diff
@@ -339,39 +339,80 @@ class GisidaWrapper extends React.Component<GisidaProps, GisidaState> {
...
+      // console.log(structures.features.map(f => f.geometry && f.geometry.type));
+      const structurePolygons = { ...structures, features: [] as StructureGeoJSON[] };
+      const structurePoints = { ...structures, features: [] as StructureGeoJSON[] };
+      structures.features.forEach((structure: StructureGeoJSON) => {
+        const featureType = structure.geometry && structure.geometry.type;
+        switch (featureType) {
+          case POINT: {
+            structurePoints.features.push(structure);
+          }
+          default: {
```
moshthepitt: Is there a danger that we may end up pushing things that aren't "structures"?

cKellyDesign: I think that we should assume that if the geoms are included in the Superset slice response that they are indeed structures.

src/containers/pages/FocusInvestigation/map/active/tests/index.test.tsx
```diff
@@ -348,4 +348,43 @@ describe('containers/pages/FocusInvestigation/activeMap', () => {
...
+        <Router history={history}>
+          <ConnectedMapSingleFI {...props} />
+        </Router>
+      </Provider>
+    );
+
+    // structures prop
+    const singleActiveWrapperProps = wrapper.find('SingleActiveFIMap').props();
+    expect((singleActiveWrapperProps as MapSingleFIProps).plan).toEqual(fixtures.plan1);
+    expect((singleActiveWrapperProps as MapSingleFIProps).structures).not.toBeNull();
```
cKellyDesign: @moshthepitt @p-netm @FrankApiyo - Please assist me here, I cannot figure out why these are failing...

Structures is null because the plan is null because for some reason the component isn't getting a Plan from the store when it maps state to props. I've [copied this test implementation from directly above]((singleActiveWrapperProps as MapSingleFIProps).plan) so I don't understand why `(singleActiveWrapperProps as MapSingleFIProps).plan` is `null`.

src/containers/pages/FocusInvestigation/map/active/tests/index.test.tsx
```diff
@@ -352,4 +362,48 @@ describe('containers/pages/FocusInvestigation/activeMap', () => {
...
+
+    store.dispatch(fetchGoals([fixtures.goal3]));
+    store.dispatch(fetchJurisdictions([fixtures.jurisdictions[0]]));
+    store.dispatch(fetchPlans([fixtures.plan1]));
+    store.dispatch(fetchTasks(fixtures.tasks));
+    store.dispatch(setStructures(structures as structureDucks.Structure[]));
+
+    // check that we can get a plan from the store
+    expect(store.getState().plans.plansById[fixtures.plan1.id]).toEqual(fixtures.plan1); // passes
+    expect(getPlanById(store.getState(), fixtures.plan1.id)).toEqual(fixtures.plan1); // fails
```
cKellyDesign: @moshthepitt @p-netm @kelvin-muchiri @ciremusyoka 

Any idea why this test is failing to get a plan from the store usig the `getPlanById()` method?

cKellyDesign: Somehow `getPlanById` is returning `undefined`, which I'm not even sure how that's possible (since it should fall back to `null`). Also, when I copy the function itself into the test it works, but not when it's imported.

Screen Shot 2020-02-24 at 8 03 14 AM

ciremusyoka: @cKellyDesign Have noticed when you comment out the other tests the test is passing. My guess is some tests above are affecting the test but have not been able to figure out why.

cKellyDesign: @ciremusyoka if we do that are there any `expect`s passing that look at the component props? If not then it can/will "pass" even though the plan mapped to props is `null` causing the mounted component to not render a map.

moshthepitt:
Was superseded by #917

PR #644, opened 2020-01-28
Refactor express server

moshthepitt:
- Clean up express app
- Refactor package.json
    - Add scripts
    - Explicitly install all required packages
- Add file based session storage
- Store OAuth credentials in session
- Add login protection for `/oauth/state`
- Add error handler middleware
- Add logout view

### How to test

1. `run yarn build` to build frontend
2. cd into server directory
3. create appropriate .env using .env.sample
    - EXPRESS_REACT_BUILD_PATH is supposed to point to the build directory produced in (1) above
3. run `yarn dev:start`
4. visit `http://localhost:3000/oauth/state` ==> you should get not authorized
5. visit `http://localhost:3000/oauth/opensrp` to start auth process
6. you should end up on a page that says "success" after coming back from OpenSRP
7. visit `http://localhost:3000/oauth/state` ==> you should get OAuth credentials
8. visit `http://localhost:3000/oauth/state` in incognito ==> you should get not authorized

#### To test logout

9. visit `http://localhost:3000/oauth/state` (not in incognito) ==> you should get OAuth credentials
10. visit `http://localhost:3000/logout (not incognito)  ==> you should get logged out
11. visit `http://localhost:3000/oauth/state` (not incognito)  ==> you should get OAuth credentials


moshthepitt:
> you should end up on a page that says "success" after coming back from OpenSRP

@p-netm is going to change this to redirect back to the client/frontend so that we pass the auth state to that

PR #655, opened 2020-02-03
Fix Google Analytics tests

moshthepitt:
This PR attempts to fix tests for the Google Analytics component.

So far the following has been done by this PR:

- Add mock for env.ts in the WithGATracker tests
- Remove environment variables from mocks module (Because the mocks should be static and not rely on getting data from anywhere else.)
- Removes the snapshot test for WithGATracker - because it was very large and did not seem like it was necessary.


src/configs/__mocks__/env.ts
```diff
@@ -31,10 +31,10 @@ export type PLAN_UUID_NAMESPACE = typeof PLAN_UUID_NAMESPACE;
 export const ACTION_UUID_NAMESPACE = '1bb3c217-af1c-4749-b3df-4f6108582e3c';
 export type ACTION_UUID_NAMESPACE = typeof ACTION_UUID_NAMESPACE;
 
-export const GA_CODE = process.env.REACT_APP_GA_CODE || '';
+export const GA_CODE = '';
```
moshthepitt: @cKellyDesign what is the value of this variable that should be used in tests?

src/configs/__mocks__/env.ts
```diff
@@ -31,10 +31,10 @@ export type PLAN_UUID_NAMESPACE = typeof PLAN_UUID_NAMESPACE;
...
 export type GA_CODE = typeof GA_CODE;
 
-export const GA_ENV = process.env.REACT_APP_GA_ENV || 'dev';
+export const GA_ENV = 'test';
```
moshthepitt: @cKellyDesign what is the value of this variable that should be used in tests?

src/configs/__mocks__/env.ts
cKellyDesign: @moshthepitt I just added the GA_CODE from the staging analytics

src/configs/__mocks__/env.ts
cKellyDesign: Yup should be `test` to make it work in the testing environment

src/components/page/WithGATracker/tests/index.test.tsx
```diff
@@ -54,7 +56,7 @@ describe('components/WithGATracker', () => {
         </ConnectedRouter>
       </Provider>
     );
-    expect(toJson(wrapper)).toMatchSnapshot();
+    expect(wrapper.find('App').length).toEqual(1);
```
cKellyDesign: Nice

PR #657, opened 2020-02-04
Do not use GA HOC if GA code is not provided

cKellyDesign:
This PR:
* adds a simple conditional to disallow the WithGATracking component from mounting the HOC.
* adds simple conditions to disallow GA Tracking helper methods from tracking without a GA Code
* adds a utility function to extract the GA Code from the WithGATracking `options` param if provided, even if the code is an empty string
* adds tests to make sure page tracking is not happening in the absence of a GA_CODE
* references the GA_CODE and GA_ENV from options only
* adds `FlexComponent` and `ConnectedFlexComponent` to `types.ts`
* adds correct typings for WithGATracking

This PR addresses:
* Point C from https://github.com/onaio/reveal-frontend/issues/641
* Point 3 from https://github.com/onaio/reveal-frontend/issues/637#issuecomment-578252808

### DIFF
--- a/src/components/page/WithGATracker/index.tsx
+++ b/src/components/page/WithGATracker/index.tsx
@@ -12,9 +12,16 @@ let username = (getUser(store.getState()) || {}).username || '';
 
 /**
  * helper function to get the Google Analytics tracking code
- * @param options tracking options for the page view
+ * @param {FlexObject|undefined} options tracking options for the page view
+ * @returns {string}
  */
-export const getGaCode = (options: FlexObject = {}) => (options && options.GA_CODE) || GA_CODE;
+export const getGaCode = (options: FlexObject | undefined) => {
+  const gaCode =
+    (options && !!!options.GA_CODE && options.GA_CODE) ||
+    ((!options || !!options.GA_CODE) && GA_CODE) ||
+    '';
+  return gaCode;
+};
 
 /**
  * helper function to set the Google Analytics dimension for username
--- a/src/components/page/WithGATracker/index.tsx
+++ b/src/components/page/WithGATracker/index.tsx
@@ -43,12 +43,15 @@ export const getGAusername = (): string => username;
  * @param {string} page the url string of the page view being tracked
  * @param {FlexObject} options tracking options for the page view
  */
-export const trackPage = (page: string, options: FlexObject = {}): void => {
-  GoogleAnalytics.set({
-    page,
-    ...options,
-  });
-  GoogleAnalytics.pageview(page);
+export const trackPage = (page: string, options?: FlexObject): void => {
+  const gaCode = getGaCode(options);
+  if (gaCode && gaCode.length) {
+    GoogleAnalytics.set({
+      page,
+      ...options,
+    });
+    GoogleAnalytics.pageview(page);
+  }
 };
 
 /**
--- a/src/components/page/WithGATracker/index.tsx
+++ b/src/components/page/WithGATracker/index.tsx
@@ -73,7 +73,7 @@ const WithGATracker = (WrappedComponent: any, options: FlexObject = {}) => {
     return WrappedComponent;
   }
 
-  const HOC = class extends Component<Props> {
+  const WithGATrackerHOC = class extends Component<Props> {
     public componentDidMount() {
       // update the username dimension
       const user = (getUser(store.getState()) || {}) as FlexObject;
@@ -106,7 +106,7 @@ const WithGATracker = (WrappedComponent: any, options: FlexObject = {}) => {
     }
   };
 
-  return HOC;
+  return WithGATrackerHOC;
 };
 
 export default WithGATracker;
--- a/src/components/page/WithGATracker/index.tsx
+++ b/src/components/page/WithGATracker/index.tsx
@@ -3,6 +3,7 @@ import { getUser } from '@onaio/session-reducer';
 import React, { Component } from 'react';
 import GoogleAnalytics from 'react-ga';
 import { RouteComponentProps } from 'react-router';
+import * as env from '../../../configs/env';
 import { TEST } from '../../../constants';
 import { FlexObject, RouteParams } from '../../../helpers/utils';
 import store from '../../../store';
@@ -17,6 +18,11 @@ export interface TrackingOptions {
   GA_CODE: string;
   GA_ENV?: string;
 }
+
+const defaultTrackingOptions: TrackingOptions = {
+  GA_CODE: env.GA_CODE || '',
+  GA_ENV: env.GA_ENV || TEST,
+};
 
 /**
  * helper function to set the Google Analytics dimension for username
@@ -38,7 +44,10 @@ export const getGAusername = (): string => username;
  * @param {string} page the url string of the page view being tracked
  * @param {FlexObject} options tracking options for the page view
  */
-export const trackPage = (page: string, options: TrackingOptions): void => {
+export const trackPage = (
+  page: string,
+  options: TrackingOptions = defaultTrackingOptions
+): void => {
   const { GA_CODE } = options;
   if (GA_CODE && GA_CODE.length) {
     GoogleAnalytics.set({
@@ -54,7 +63,9 @@ export const trackPage = (page: string, options: TrackingOptions): void => {
  * @param {TrackingOptions} options tracking options for the page view
  * @returns {TrackingOptions}
  */
-export const initGoogleAnalytics = (options: TrackingOptions): TrackingOptions => {
+export const initGoogleAnalytics = (
+  options: TrackingOptions = defaultTrackingOptions
+): TrackingOptions => {
   const { GA_CODE, GA_ENV } = options;
   if (GA_CODE && GA_CODE.length) {
     GoogleAnalytics.initialize(GA_CODE, {
@@ -65,7 +76,7 @@ export const initGoogleAnalytics = (options: TrackingOptions): TrackingOptions =
       username,
     });
   }
-  return options;
+  return { ...options };
 };
 
 /**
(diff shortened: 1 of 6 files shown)


moshthepitt:
@cKellyDesign so I was looking at this deeper and I think that we can make it more testable by making `WithGATracker` take the google analytics code as a param which defaults to `GA_CODE`.  In tests we can then supply null/whatever to test what happens when the google analytics code is not supplied.  A side effect would be to make `WithGATracker` more flexible in that it is not tied to the hip to a variable that comes only from env vars.

What are your thoughts on this?

Additional small nitpick: In `const HOC = class extends Component<Props>` can we change "HOC" to something more identifiable?  I dont like how it shows up as HOC in snapshot tests etc, and I think it should show up as something like GATrackerHOC or whatever else.

src/components/page/WithGATracker/index.tsx
```diff
@@ -10,6 +10,19 @@ import store from '../../../store';
...
+ * @returns {string}
+ */
+export const getGaCode = (options: FlexObject | undefined) => {
+  const gaCode =
+    (options && !!!options.GA_CODE && options.GA_CODE) ||
+    ((!options || !!options.GA_CODE) && GA_CODE) ||
+    '';
+  return gaCode;
+};
+
```
cKellyDesign: This conditional checks the existing optional `options` param for `GA_CODE` and uses that instead of `env.GA_CODE`.

src/components/page/WithGATracker/index.tsx
```diff
@@ -30,32 +43,47 @@ export const getGAusername = (): string => username;
...
+  if (gaCode && gaCode.length) {
+    GoogleAnalytics.initialize(GA_CODE, {
+      testMode: GA_ENV === 'test',
+    });
+    GoogleAnalytics.set({
+      env: GA_ENV,
+      username,
+    });
+  }
+};
```
cKellyDesign: The initialization of Google Analytics should happen at runtime (called from `App.tsx`) instead of at the time of compiling, so it can be passed an optional `options` object containing an over-writing code.

src/components/page/WithGATracker/index.tsx
```diff
@@ -30,32 +43,47 @@ export const getGAusername = (): string => username;
...
  * @param {FlexObject} options tracking options for the page view
  * @returns HOC rendering the WrappedComponent
  */
-const WithGATracker = (WrappedComponent: any, options: FlexObject = {}) => {
-  const HOC = class extends Component<Props> {
+const WithGATracker = (WrappedComponent: any, options?: FlexObject) => {
+  const gaCode = getGaCode(options);
+  if (!gaCode || !gaCode.length) {
+    return WrappedComponent;
+  }
```
cKellyDesign: This is the actual change which allows us to turn off GA tracking by not providing a GA_CODE.

src/components/page/WithGATracker/index.tsx
```diff
@@ -88,7 +116,7 @@ const WithGATracker = (WrappedComponent: any, options: FlexObject = {}) => {
     }
   };
 
-  return HOC;
+  return WithGATrackerHOC;
```
cKellyDesign: More descriptive name 🤗

src/components/page/WithGATracker/tests/index.test.tsx
```diff
@@ -80,5 +81,15 @@ describe('components/WithGATracker', () => {
...
+  });
+
+  it('tracks nothing when no GA Code is provided', () => {
+    GoogleAnalytics.pageview = jest.fn();
+    const gaOptions: FlexObject = { GA_CODE: '' };
+    initGoogleAnalytics(gaOptions);
+    trackPage('/', gaOptions);
+    trackPage(NEW_IRS_PLAN_URL, gaOptions);
+    expect(GoogleAnalytics.pageview).not.toBeCalled();
   });
```
cKellyDesign: Tests that the page tracking is not happening when the `GA_CODE` is set to `''`

src/components/page/WithGATracker/index.tsx
```diff
@@ -30,32 +43,47 @@ export const getGAusername = (): string => username;
  * @param {string} page the url string of the page view being tracked
  * @param {FlexObject} options tracking options for the page view
  */
-export const trackPage = (page: string, options: FlexObject = {}): void => {
-  GoogleAnalytics.set({
-    page,
-    ...options,
-  });
-  GoogleAnalytics.pageview(page);
+export const trackPage = (page: string, options?: FlexObject): void => {
```
moshthepitt: - I still think that `options` does not need to be of type `FlexObject`.
- I feel that it is weird that `options` is ... optional for `trackPage`.  Why does it need to be this way?  What would it mean to call `trackPage` and supply no options i.e. what would one be trying to achieve when they do that?

src/components/page/WithGATracker/index.tsx
moshthepitt: - I still maintain that `options` should not be a FlexObject.
- I think that you should not get `GA_ENV` from the global scope in this module.  Is there a compelling reason why we don't simple get it from `options` or another param supplied to `initGoogleAnalytics`?  A nice side effect of doing so would be to make `initGoogleAnalytics` more flexible AND testable.

moshthepitt:
@cKellyDesign do we need to have a test around the different values of GA_ENV i.e. I am wondering if things happen differently if GA_ENV="test" Vs GA_ENV="prod", or even say GA_ENV="mary"

src/components/page/WithGATracker/index.tsx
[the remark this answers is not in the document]
cKellyDesign: The intent was to default to the global `env. GA_CODE` value if the `options.GA_CODE` was not provided, so that we don't need to pass the `GA_CODE` to every single route in `App.tsx`. If you think it's better we can just pass it in every time to simplify this and make `options` a required param.

src/components/page/WithGATracker/index.tsx
cKellyDesign: As it is, if we didn't pass anything in as `options` then it would use the global value for `GA_CODE`, but we can make it required and only care about what is passed in with `options`.

src/components/page/WithGATracker/index.tsx
cKellyDesign: We can pass `GA_ENV` in through the `options` param.

cKellyDesign:
@moshthepitt Thanks for the review, I had left `options` as FlexObject for speed but you're right they should be specified Types which are updated as more options are defined for GA.

As for `GA_ENV`, the only purpose that serves is for testing: if it is anything else in a test environment the tests will break, if it is anything else outside of a test environment there should be no effect. Shouldn't be an issue to pass this value in through `options` as well.

src/components/page/WithGATracker/index.tsx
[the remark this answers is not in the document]
moshthepitt: I think we should consider getting GA_CODE from options only.  It would make the function [easier to reason about and more flexible](https://stackify.com/solid-design-open-closed-principle/).

cKellyDesign:
> As for GA_ENV, the only purpose that serves is for testing

That's wrong, it's also being set as a custom `env` dimension for the GA tracking. While this dimension has no functional effect in this repo, as our analytics solution matures we may be adding more dimensions.

src/components/page/WithGATracker/index.tsx
moshthepitt: I think this is similar to `getGaCode` i.e. I think its better to have just one entry point for values to get into a function and that should be the params that the function accepts.

src/components/page/WithGATracker/index.tsx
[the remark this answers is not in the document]
cKellyDesign: It is so, thanks for the read!

src/components/page/WithGATracker/index.tsx
```diff
@@ -30,32 +38,49 @@ export const getGAusername = (): string => username;
...
 /**
  * Higher Order Component (HOC) which handles Google Analytics page view tracking
  * @param {any} WrappedComponent the component to be wrapped by the HOC component
- * @param {FlexObject} options tracking options for the page view
+ * @param {TrackingOptions} options tracking options for the page view
  * @returns HOC rendering the WrappedComponent
  */
-const WithGATracker = (WrappedComponent: any, options: FlexObject = {}) => {
-  const HOC = class extends Component<Props> {
+const WithGATracker = (WrappedComponent: any, options: TrackingOptions) => {
```
moshthepitt: @cKellyDesign how do you feel about giving options a default value here, to avoid the `component={WithGATracker(ConnectedPractitionersListView, trackingOptions)}` kind of changes in `App.tsx`?

cKellyDesign: ah ok good idea, think those defaults should pull from the global env then?

moshthepitt: Lets also not use `any` here for `WrappedComponent` - I believe there is a type for React Components that we can use e.g. the `React.ElementType` kind of types from React

src/constants.tsx
```diff
@@ -41,6 +41,7 @@ export const BLOOD_SCREENING_ACTIVITY_CODE = 'bloodScreening';
 export const BEDNET_DISTRIBUTION_ACTIVITY_CODE = 'bednetDistribution';
 export const LARVAL_DIPPING_ACTIVITY_CODE = 'larvalDipping';
 export const MOSQUITO_COLLECTION_ACTIVITY_CODE = 'mosquitoCollection';
+export const TEST = 'test';
```
moshthepitt: Do we need to rename this to somehow indicate it is used for GA, in case we think there's a danger it may be inadvertently changed/translated etc?

src/components/page/WithGATracker/index.tsx
moshthepitt: > ah ok good idea, think those defaults should pull from the global env then?

Yeah

cKellyDesign: think i got it :)

cKellyDesign:
@moshthepitt I believe this is ready for review again :)

cKellyDesign:
Oh snap got it, thanks man!

cKellyDesign:
I think this should a part of `v0.4.1` so as to not hold up `v0.4.0` while testing and getting sign-off on staging.

PR #674, opened 2020-02-06
Exclude src/index.tsx from the coverage report

### LINKED ISSUE #671: Increase test coverage for src/index.tsx
Increase test coverage to be >= 85% for:

`src/index.tsx`

Part of #658

DavisRayM:
# Changes proposed in this PR

- Exclude `src/index.tsx` from the coverage report

Closes #671

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,6 +74,12 @@
     "lint-js": "eslint 'src/**/*.{js,jsx}'",
     "lint-ts": "tslint -c tslint.json 'src/**/*.{ts,tsx}'"
   },
+  "jest": {
+    "collectCoverageFrom": [
+      "src/**/*.{ts,tsx}",
+      "!src/index.tsx"
+    ]
+  },
   "husky": {
     "hooks": {
       "pre-commit": "lint-staged"


moshthepitt:
@DavisRayM was the change in circle ci configs necessary?

DavisRayM:
> @DavisRayM was the change in circle ci configs necessary?

I'm not quite sure, trying out an alternate option (setting the jest configuration in `package.json`).

DavisRayM:
> @DavisRayM was the change in circle ci configs necessary?

It's not necessary, the change in the `package.json` is more than enough to exclude the `index.tsx` from the coverage report. I've made the change in the latest commit.

moshthepitt:
I mean that it looks like you already have tests.  Is it necessary to go ahead and still exclude the file from coverage?

DavisRayM:
> I mean that it looks like you already have tests. Is it necessary to go ahead and still exclude the file from coverage?

The test I've written doesn't execute any code within `index.tsx` as such the coverage isn't affected at all. (It'll still be 0% with my test)

moshthepitt:
> The test I've written doesn't execute any code within index.tsx as such the coverage isn't affected at all. (It'll still be 0% with my test)

Cool, understood

DavisRayM:
> > The test I've written doesn't execute any code within index.tsx as such the coverage isn't affected at all. (It'll still be 0% with my test)
> 
> Cool, understood

Yea, 😅 I'm not quite sure if there is a better way to test the `index.tsx` at this point. But I can add a `DNM` tag for now while I research a bit more.

DavisRayM:
I haven't been able to find any alternative to the current implementation.... I think it would be better to remove the `index.tsx` file from the coverage report and forgo the test.

CC: @moshthepitt

moshthepitt:
@DavisRayM sounds good to me

DavisRayM:
> @DavisRayM sounds good to me

Nice, I've dropped the commit that added the test. This PR should now be excluding the `src/index.tsx` from the coverage report.

PR #675, opened 2020-02-06
Fix log out

p-netm:
Feature to logout user out of store and invalidate session in express server

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -46,6 +46,7 @@ REACT_APP_OPENSRP_AUTHORIZATION_URL=https://reveal-stage.smartregister.org/opens
 REACT_APP_OPENSRP_USER_URL=https://reveal-stage.smartregister.org/opensrp/user-details
 REACT_APP_OPENSRP_LOGOUT_URL='https://reveal-stage.smartregister.org/opensrp/logout.do'
 REACT_APP_OPENSRP_OAUTH_STATE=opensrp
+EXPRESS_OAUTH_LOGOUT_URL='http://localhost:3000/logout'
 EXPRESS_OAUTH_GET_STATE_URL=http://localhost:3000/oauth/state
 # Onadata oAuth
 REACT_APP_ENABLE_ONADATA_OAUTH=false
--- a/.env.sample
+++ b/.env.sample
@@ -46,8 +46,8 @@ REACT_APP_OPENSRP_AUTHORIZATION_URL=https://reveal-stage.smartregister.org/opens
 REACT_APP_OPENSRP_USER_URL=https://reveal-stage.smartregister.org/opensrp/user-details
 REACT_APP_OPENSRP_LOGOUT_URL='https://reveal-stage.smartregister.org/opensrp/logout.do'
 REACT_APP_OPENSRP_OAUTH_STATE=opensrp
-EXPRESS_OAUTH_LOGOUT_URL='http://localhost:3000/logout'
-EXPRESS_OAUTH_GET_STATE_URL=http://localhost:3000/oauth/state
+REACT_APP_EXPRESS_OAUTH_LOGOUT_URL='http://localhost:3000/logout'
+REACT_APP_EXPRESS_OAUTH_GET_STATE_URL=http://localhost:3000/oauth/state
 # Onadata oAuth
 REACT_APP_ENABLE_ONADATA_OAUTH=false
 REACT_APP_ONADATA_CLIENT_ID=hunter2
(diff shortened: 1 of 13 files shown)


.env.sample
```diff
@@ -46,6 +46,7 @@ REACT_APP_OPENSRP_AUTHORIZATION_URL=https://reveal-stage.smartregister.org/opens
 REACT_APP_OPENSRP_USER_URL=https://reveal-stage.smartregister.org/opensrp/user-details
 REACT_APP_OPENSRP_LOGOUT_URL='https://reveal-stage.smartregister.org/opensrp/logout.do'
 REACT_APP_OPENSRP_OAUTH_STATE=opensrp
+EXPRESS_OAUTH_LOGOUT_URL='http://localhost:3000/logout'
 EXPRESS_OAUTH_GET_STATE_URL=http://localhost:3000/oauth/state
```
moshthepitt: Lets maintain the naming convention of prefixing `REACT_APP_` to all env vars for the React app

moshthepitt:
@p-netm how does logout work after this PR?

.env.sample
p-netm: ok

cKellyDesign: Yeah they must be prefixed with `REACT_APP_` in order to be picked up by the Create React App scripts right?

p-netm:
> @p-netm how does logout work after this PR?


1. dispatches an action to logout user from session slice of the store, its not a container since by the time a user is logged in, there is the quarantee that the session reducer will have been registered.
2. window.location.href will force a page navigation to `<domain>/logout`, this will be picked by the express server routes, which will invalidate the session and redirect user-agent to login page.

moshthepitt:
@p-netm thanks.  Let me know when you are ready for a re-review.

moshthepitt:
@p-netm keep in mind that the previous implementation called the logout url on the OpenSRP server if it was supplied

cKellyDesign:
> which will invalidate the session and redirect user-agent to login page

Echoing [what Mosh said](https://github.com/onaio/reveal-frontend/pull/675#issuecomment-583379271): Does this PR [logout the user from OpenSRP](https://reveal-stage.smartregister.org/opensrp/logout.do)?

p-netm:
> Echoing [what Mosh said](https://github.com/onaio/reveal-frontend/pull/675#issuecomment-583379271): Does this PR [logout the user from OpenSRP](https://reveal-stage.smartregister.org/opensrp/logout.do)?

@cKellyDesign unless there be strong sentiments as to retaining that feature, i am inclined to leave it out in this pr, the reason being; logging out of the authorization server should ideally be something in the resource-owner's(superset-user) domain and not the clients(the web app)

cKellyDesign:
We can keep it out of this PR if want to discuss further, but I do that think since we're going for single sign-on it feels like we should also have single sign-out.

p-netm:
> but I do that think since we're going for single sign-on it feels like we should also have single sign-out.

@moshthepitt what do you say ?

p-netm:
@cKellyDesign This pr includes a string that might need transalations, i have added the string to the json files and also exported it from `Lang.ts`, what are the next steps?

moshthepitt:
@p-netm @cKellyDesign I am fine with creating an issue to discuss single sign out further.  Lets merge this one for now.

src/configs/lang.ts
```diff
@@ -404,3 +404,4 @@ export const ORGANIZATION_LABEL = TEAM;
 export const ORGANIZATIONS_LABEL = TEAMS;
 
 export const ORGANIZATION_NAME_TITLE = translate('ORGANIZATION_NAME_TITLE', 'Team Name');
+export const WELCOME_BACK = translate('WELCOME_BACK', 'Welcome back');
```
cKellyDesign: Let's create a small ticket to add TH translations for this.

src/configs/strings/th.json
```diff
@@ -830,5 +830,9 @@
   "ORGANIZATION_NAME_TITLE": {
     "message": "ทีมปฏิบัติงาน",
     "description": "Organization list column header"
+  },
+  "WELCOME_BACK": {
+    "message": "",
+    "description": "prompt after a user has logged in"
```
cKellyDesign: Let's remove this until we have an actual translation so it falls back to English. With this it will render an empty string instead of `Welcome Back` when run with TH translations.

cKellyDesign: (I'll make a ticket)

src/configs/lang.ts
cKellyDesign: ticket created at https://github.com/onaio/reveal-frontend/issues/680

src/configs/strings/th.json
p-netm: makes sense, lemmy remove it

Issue #634, opened 2020-01-10
Generalize and decouple LinkAsButton defaultProps.text


cKellyDesign:
Currently the default text for the general `LinkAsButton` component is specific to adding focus investigation plans and should be generalized.

- [x] Change defaultProps.text to `Link` (and update the component's snapshot)
- [x] Specify `text` prop in references using the default props (so their snapshots don't break)
- [x] Add ticket for migrating the LinkAsButton to a `js-tools` package ([ticket](https://github.com/onaio/js-tools/issues/128))

moshthepitt:
@ciremusyoka once we have an issue on js-tools created, we can close this.

cc: @cKellyDesign

cKellyDesign:
Ticket created at https://github.com/onaio/js-tools/issues/128

PR #678, opened 2020-02-07
Generalize LinkAsButton text prop

ciremusyoka:
part of #634

### DIFF
--- a/src/components/LinkAsButton/index.tsx
+++ b/src/components/LinkAsButton/index.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { Link } from 'react-router-dom';
-import { ADD_FOCUS_INVESTIGATION } from '../../configs/lang';
+import { LINK } from '../../configs/lang';
 import { NEW_PLAN_URL } from '../../constants';
 
 /** interface for LinkAsButton component props */
@@ -23,7 +23,7 @@ const LinkAsButton = (props: Props) => {
 /** default props for LinkAsButton component */
 const defaultProps: Props = {
   classNameProp: 'focus-investigation btn btn-primary float-right mt-5',
-  text: ADD_FOCUS_INVESTIGATION,
+  text: LINK,
   to: NEW_PLAN_URL,
 };
 LinkAsButton.defaultProps = defaultProps;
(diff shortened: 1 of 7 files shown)


cKellyDesign:
Agreed, something is up with the App snapshot, looks like the same issue seen here: https://github.com/onaio/reveal-frontend/pull/679

src/components/LinkAsButton/index.tsx
```diff
@@ -6,24 +6,24 @@ import { NEW_PLAN_URL } from '../../constants';
...
 const LinkAsButton = (props: Props) => {
-  const { to, classNameProp, text } = props;
+  const { to, classNameProp, link } = props;
   return (
     <Link to={to} className={classNameProp}>
-      {text}
+      {link}
     </Link>
   );
 };
```
cKellyDesign: @ciremusyoka Why are we changing `props.text` to `props.link`?

It feels a little confusing b/c the actual "link" in this case is a URL which is passed in through `props.to`, whereas the child of `<Link />` is a "text" node so it may make more sense to keep that as `text`.

Seems to me like if anything props should change here it should be `to` => `link`.

@moshthepitt @p-netm what are your thoughts?

ciremusyoka:
My bad i miss understood the issue let me make the right changes.

src/components/LinkAsButton/index.tsx
ciremusyoka: My bad i miss understood the issue let me make the right changes.

src/configs/strings/en.json
```diff
@@ -842,5 +842,9 @@
   "ORGANIZATION_NAME_TITLE": {
     "message": "Team Name",
     "description": "Organization list column header"
+  },
+  "LINK": {
+    "message": "Link",
+    "description": "URL"
```
cKellyDesign: Let's just change this description to be a bit more specific:
`Button text: describes a URL leading to another page`

src/containers/pages/FocusInvestigation/active/index.tsx
```diff
@@ -356,7 +357,7 @@ class ActiveFocusInvestigation extends React.Component<
                     <h3 className="mb-3 mt-5 page-title">{ROUTINE_TITLE}</h3>
                   </Col>
                   <Col xs="6">
-                    <LinkAsButton />
+                    <LinkAsButton text={ADD_FOCUS_INVESTIGATION} />
```
cKellyDesign: 👍

Issue #639, opened 2020-01-24
Add tests for each translated language


moshthepitt:
We need to write tests that run against each language that we support 

Context: https://onaio.slack.com/archives/C0482AJP6/p1579863718032900

PR #679, opened 2020-02-07
Make IRS report legend colors customizable

### LINKED ISSUE #646: RVL-734 Change Colour Gradient Range for Dashboards
New color gradients should be:

- Grey: 0%
- Red: 0 to 70%
- Yellow: 70 to 80%
- Light Green: 80 to 85%
- Green: >85%

Current values: https://github.com/onaio/reveal-frontend/blob/master/src/configs/settings.ts#L848

This change should only affect Namibia.

More info: https://smartregister.atlassian.net/browse/RVL-734

kelvin-muchiri:
Closes #646

### DIFF
--- a/src/App/tests/__snapshots__/App.test.tsx.snap
+++ b/src/App/tests/__snapshots__/App.test.tsx.snap
@@ -15,7 +15,7 @@ exports[`App renders App correctly 1`] = `
   <Router
     history={
       Object {
-        "action": "POP",
+        "action": "REPLACE",
         "block": [Function],
         "createHref": [Function],
         "go": [Function],
@@ -25,7 +25,8 @@ exports[`App renders App correctly 1`] = `
         "listen": [Function],
         "location": Object {
           "hash": "",
-          "pathname": "/",
+          "key": "nmani6",
+          "pathname": "/login",
           "search": "",
           "state": undefined,
         },
@@ -80,7 +81,7 @@ exports[`App renders App correctly 1`] = `
                   authenticated={false}
                   history={
                     Object {
-                      "action": "POP",
+                      "action": "REPLACE",
                       "block": [Function],
                       "createHref": [Function],
                       "go": [Function],
@@ -90,7 +91,8 @@ exports[`App renders App correctly 1`] = `
                       "listen": [Function],
                       "location": Object {
                         "hash": "",
-                        "pathname": "/",
+                        "key": "nmani6",
+                        "pathname": "/login",
                         "search": "",
                         "state": undefined,
                       },
@@ -101,7 +103,8 @@ exports[`App renders App correctly 1`] = `
                   location={
                     Object {
                       "hash": "",
-                      "pathname": "/",
+                      "key": "nmani6",
+                      "pathname": "/login",
                       "search": "",
                       "state": undefined,
                     }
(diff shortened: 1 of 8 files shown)


moshthepitt:
@kelvin-muchiri it would be mice to have a way, in tests, to show that:

- Namibia colors are rendered correctly when using Namibia settings
- default colors are rendered correctly when not using Namibia settings

kelvin-muchiri:
> * Please fix broken tests
> * Why is the App component affected at all? (might be an issue with your env vars)

There is one test failing. @p-netm confirmed it was failing before. Though he said he'll look into it

moshthepitt:
> There is one test failing. @p-netm confirmed it was failing before. Though he said he'll look into it

@kelvin-muchiri @p-netm as far as I know we have no failing tests on the master branch.

cKellyDesign:
@kelvin-muchiri was this branched off of `master`? Not sure how it would have existing failing tests.

src/configs/settings.ts
```diff
@@ -870,6 +870,45 @@ export const indicatorThresholdsIRS: IndicatorThresholds = {
...
+    name: 'Red',
+    orEquals: true,
+    value: 0.7,
+  },
+  YELLOW_THRESHOLD: {
+    color: '#FFDC00',
+    name: 'Yellow',
+    value: 0.8,
+  },
+};
```
cKellyDesign: Can we pass the `name` strings in from the `lang.ts` file so we are sure they are translated?

src/App/tests/__snapshots__/App.test.tsx.snap
```diff
@@ -25,7 +25,8 @@ exports[`App renders App correctly 1`] = `
         "listen": [Function],
         "location": Object {
           "hash": "",
-          "pathname": "/",
+          "key": "nmani6",
+          "pathname": "/login",
```
cKellyDesign: Looks like the App redirecting to `/login` when it mounts in the test?

src/App/tests/__snapshots__/App.test.tsx.snap
```diff
@@ -846,370 +851,161 @@ exports[`App renders App correctly 1`] = `
...
+                          path="/login"
+                          render={[Function]}
                         >
-                          <PrivateRoute
-                            authenticated={false}
-                            component={[Function]}
-                            computedMatch={
+                          <OauthLogin
+                            ProviderLinksComponent={[Function]}
+                            history={
```
cKellyDesign: Seems like these shouldn't be changing

kelvin-muchiri:
> @kelvin-muchiri it would be mice to have a way, in tests, to show that:
> 
> * Namibia colors are rendered correctly when using Namibia settings
> * default colors are rendered correctly when not using Namibia settings

Yea..sure..

moshthepitt:
> Great thanks @kelvin-muchiri - just one more step to add these to the i18n formatted en.json file.

Good catch @cKellyDesign.  What can we do to ensure that something like this is caught in a failing test in the future?

src/configs/settings.ts
cKellyDesign: I've added these to a ticket requesting TH translations for new display strings.

cKellyDesign:
> What can we do to ensure that something like this is caught in a failing test in the future?

@moshthepitt is that covered by this ticket? https://github.com/onaio/reveal-frontend/issues/639

moshthepitt:
> @moshthepitt is that covered by this ticket? #639

Yes, my bad!

Issue #664, opened 2020-02-06
Increase test coverage for IRS Plan


moshthepitt:
Increase test coverage to be >= 85% for:

`src/containers/pages/InterventionPlan/IRS/plan/index.tsx ==> 6.21%`

Part of #658

cKellyDesign:
@moshthepitt @p-netm @kelvin-muchiri - for the [IRS Jurisdiction Selection/Assignment](https://github.com/onaio/reveal-frontend/blob/master/src/containers/pages/InterventionPlan/IRS/plan/index.tsx) we are using quite a few API calls via the [OpenSRP service](https://github.com/onaio/reveal-frontend/blob/master/src/services/opensrp/index.ts) ([1](https://github.com/onaio/reveal-frontend/blob/master/src/containers/pages/InterventionPlan/IRS/plan/index.tsx#L260), [2](https://github.com/onaio/reveal-frontend/blob/master/src/containers/pages/InterventionPlan/IRS/plan/index.tsx#L275), [4](https://github.com/onaio/reveal-frontend/blob/master/src/containers/pages/InterventionPlan/IRS/plan/index.tsx#L348)) and the [SupersetConnector service](https://github.com/onaio/reveal-frontend/blob/master/src/services/superset/index.ts) ([3](https://github.com/onaio/reveal-frontend/blob/master/src/containers/pages/InterventionPlan/IRS/plan/index.tsx#L292)), both use `fetch` but only the OpenSRP service is picking up mock fetch responses in the tests. The current tests don't fail, but they also don't mount anything other than the spinner component b/c none of the calls are being successfully mocked.

Any suggestions on how to mock all these calls together?

moshthepitt:
>@moshthepitt @p-netm @kelvin-muchiri - for the IRS Jurisdiction Selection/Assignment we are using quite a few API calls via the OpenSRP service (1, 2, 4) and the SupersetConnector service (3), both use fetch but only the OpenSRP service is picking up mock fetch responses in the tests. The current tests don't fail, but they also don't mount anything other than the spinner component b/c none of the calls are being successfully mocked. (my progress so far)

> Any suggestions on how to mock all these calls together?

[...]

PR #683, opened 2020-02-10
[DNM] Increase IRS Plan Test Coverage

cKellyDesign:
Closes: #664

### DIFF
--- a/src/containers/pages/InterventionPlan/IRS/plan/tests/__snapshots__/index.test.tsx.snap
+++ b/src/containers/pages/InterventionPlan/IRS/plan/tests/__snapshots__/index.test.tsx.snap
@@ -1,5 +1,121 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`containers/pages/IRS/plan calls the correct API endpoints to load map and table: IRS PLAN 1`] = `
+<IrsPlan
+  allJurisdictionIds={Array []}
+  assignmentsArray={Array []}
+  fetchAllJurisdictionIdsActionCreator={[Function]}
+  fetchAssignmentsActionCreator={[Function]}
+  fetchJurisdictionsActionCreator={[Function]}
+  fetchOrganizationsActionCreator={[Function]}
+  fetchPlansActionCreator={[Function]}
+  history={
+    Object {
+      "action": "POP",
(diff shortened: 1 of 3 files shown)


src/containers/pages/InterventionPlan/IRS/plan/tests/__snapshots__/index.test.tsx.snap
```diff
@@ -1,5 +1,121 @@
...
+  supersetService={[Function]}
+>
+  <Ripple
+    borderColor="#ff5d00"
+    borderStyle="solid"
+    borderWidth="4px"
+    height="64px"
+    minHeight="80vh"
+    width="64px"
+  >
```
cKellyDesign: So we can see that the map and table are not mounting in the tests because the ripple is still in the snapshot.

src/containers/pages/InterventionPlan/IRS/plan/tests/index.test.tsx
```diff
@@ -117,4 +126,48 @@ describe('containers/pages/IRS/plan', () => {
...
+
+    const mock: any = jest.fn();
+    const { identifier: id } = irsPlanRecordResponse1;
+    // const supersetServiceMock: any = jest.fn();
+    // supersetServiceMock.mockImplementation(async () => jurisdictionsArray as FlexObject[]);
+
+    fetch
+      .mockResponseOnce(JSON.stringify([] as Organization[])) // opensrp (2) - get organizations: the only successful mock response
+      .mockResponseOnce(JSON.stringify(jurisdictionsArray as FlexObject[])) // superset (3) - get all jurisdictions
+      .mockResponseOnce(JSON.stringify([jurisdictionsArray[0]])); // opensrp (4) - get level 0 jurisdiction
```
cKellyDesign: After dropping console logs in `componentDidMount` I can see that no combination of these works. In fact 

[According to the docs](https://www.npmjs.com/package/jest-fetch-mock#mocking-multiple-fetches-with-different-responses) we should be able to simply chain these responses for the multiple `fetch`s. However, I can see that the only `mockResponseOnce` being returned is `opensrp (2)` which is the first fetch being called in the test.

moshthepitt:
I think this one has been superseded by #789

PR #684, opened 2020-02-12
467 use progressbar component from js-tools

### LINKED ISSUE #467: Use ProgressBar Component from onaio package
Currently we are [using a local ProgressBar component in the IRS Map page](https://github.com/onaio/reveal-frontend/blob/irs-reporting-table/src/containers/pages/IRS/Map/index.tsx#L25), which should be replaced by the `ProgressIndicators` package being introduced by onaio/js-tools#95.

- [ ] Install package ([@onaio/progress-indicators](https://github.com/onaio/js-tools/tree/master/packages/ProgressIndicators))
- [ ] Update component props to pass in correctly structured thresholds model

RayceeM:
Fixes #467


moshthepitt:
@cKellyDesign - @RayceeM and I looked at it last week, I think she now has a way forward.  At least I hope so @RayceeM ?

src/containers/pages/IRS/Map/tests/__snapshots__/index.test.tsx.snap
```diff
@@ -151,51 +151,34 @@ Array [
...
-          "color": "#FF4136",
-          "name": "Red",
-          "orEquals": true,
-          "value": 0.75,
-        },
-        "YELLOW_THRESHOLD": Object {
-          "color": "#FFDC00",
-          "name": "Yellow",
-          "value": 0.9,
-        },
```
moshthepitt: Do you think there is value in maintaining these older style thresholds?  For example, how would we now deal with a threshold that is `>= 80%`?

src/containers/pages/IRS/Map/tests/__snapshots__/index.test.tsx.snap
```diff
@@ -151,51 +151,34 @@ Array [
...
     value={50}
   >
     <div
       className="progress"
-      style={
-        Object {
-          "height": "10px",
-          "marginBottom": "15px",
-        }
-      }
```
moshthepitt: Does the removal of this imply that we no longer need this CSS?

src/containers/pages/IRS/Map/tests/__snapshots__/index.test.tsx.snap
RayceeM: if it's greater than the color value then it takes the default lineColor which is "#0000FF"

moshthepitt: @RayceeM I mean how will we handle the possibility of having `>=` as well as just `>` for the thresholds?

src/containers/pages/IRS/Map/tests/__snapshots__/index.test.tsx.snap
RayceeM: We do, height is also a prop in the js library  that had a default value of `10px`

cKellyDesign:
@RayceeM have you tried discarding the incoming changes to yarn.lock (from `master`) and re-running `yarn install` to resolve this conflict?

RayceeM:
@cKellyDesign  no not yet, have been trying to start reveal but I keep getting an issue with `babel-loader` version

cKellyDesign:
@RayceeM oh interesting, I was having issues with babel-loader yesterday in another project as well...

At any rate, you may try dumping your `node_modules` and `yarn.lock` file before re-installing to see if that clears things up.

```sh
rm -rf node_modules && rm -f yarn.lock && yarn install
```

moshthepitt:
@RayceeM please update this branch.  A huge amount of stuff has changed :)

RayceeM:
Here is a screenshot of a page using the progressbar component from js-library
nchelenge2

RayceeM:
Screenshot of progressbar in a focus investigstion map using the js-component after passing lineColor
FI3

src/containers/pages/FocusInvestigation/map/active/index.tsx
```diff
@@ -343,6 +349,16 @@ const SingleActiveFIMap = (props: MapSingleFIProps & RouteComponentProps<RoutePa
 
   const mapBounds = getMapBounds(jurisdiction);
 
+  function getLineColor(decimalValue: number) {
```
moshthepitt, line 352: Do you think that we should make [something like this](https://github.com/onaio/reveal-frontend/blob/8fcee000dc30c07ece47f898b9962b226cc92b0a/src/configs/settings.ts#L1617) for FI and then use it here?

RayceeM: Yeap, we can and pass `lineColorThreshold` instead of `lineColor`

moshthepitt: Cool, lets make one for Focus Investigations.  You can make it in the same settings file and can copy how we did it for [Namibia IRS](https://github.com/onaio/reveal-frontend/blob/8fcee000dc30c07ece47f898b9962b226cc92b0a/src/configs/settings.ts#L1644) if it makes sense.

src/containers/pages/FocusInvestigation/map/active/index.tsx
```diff
@@ -432,7 +432,11 @@ const SingleActiveFIMap = (props: MapSingleFIProps & RouteComponentProps<RoutePa
                         ({goalReport.prettyPercentAchieved})
                       </p>
                       <br />
-                      <ProgressBar value={goalReport.percentAchieved} max={1} />
+                      <ProgressBar
+                        value={0.012}
```
moshthepitt, line 436: Looks like this is left over from testing?  We should revert back to `value={goalReport.percentAchieved}`

moshthepitt (changes requested):

src/containers/pages/FocusInvestigation/map/active/index.tsx
RayceeM: Yeah I have reverted back in the next commit

moshthepitt (approved):

p-netm (approved):

PR #686, opened 2020-02-12
Add tests for DrillDownTableLinkedCell

### LINKED ISSUE #672: Increase test coverage for DrillDownTableLinkedCell
Increase test coverage to be >= 85% for:

`src/components/DrillDownTableLinkedCell/index.tsx`

Part of #658

kelvin-muchiri:
Closes #672 

The `DrillDownTableLinkedCell` component expects `cellValue` prop to be of type `Node`. At first, I tried `const cellValue: Text = new Text('value')` but it did not work. Changing to `const cellValue: string = 'some string'` does not work either.

I went through the repo and implemented as it is in line https://github.com/onaio/js-tools/blob/master/packages/DrillDownTable/src/index.tsx#L139  `const cellValue = cell.value` and it worked.

Still trying to figure out why.


moshthepitt:
@kelvin-muchiri thanks for this.  Please rebase your work with current master.

src/components/DrillDownTableLinkedCell/tests/index.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+    </Router>
+  );
+
+  it('receives props correctly', () => {
+    expect(wrapper.find(DrillDownTableLinkedCell).props()).toEqual(props);
+  });
+
+  it('renders correctly if prop hasChildren is true', () => {
+    expect(wrapper.find(DrillDownTableLinkedCell).html()).toEqual(
+      '<div><a href="/http://example.com/path/2019-11-20">2019-11-20</a></div>'
```
moshthepitt: @kelvin-muchiri do you have any insight on why this is happening? i.e. adding a `/` to the URL.  Do we need to create an issue to ensure that this component can actually successfully render links to external URLs?

kelvin-muchiri: Well, that's a nice thought. I understand in react linking to an external URL you'll have to use anchor tag and not `Link`. Maybe we should update the test and use urlPath say as `/path` to avoid confusion.

moshthepitt: @kelvin-muchiri not sure what you meant, but try it out and lets see

Issue #669, opened 2020-02-06
Increase test coverage for InterventionPlan/IRS/index.tsx


moshthepitt:
Increase test coverage to be >= 85% for:

`src/containers/pages/InterventionPlan/IRS/index.tsx`

Part of #658

moshthepitt:
Fixed in #789

PR #689, opened 2020-02-13
Load draft plans on default IRS Intervention Page

### LINKED ISSUE #685: IRS Plan page just spins on reveal-stage
The IRS plan page just spins on the Reveal-stage UI:

https://web.reveal-stage.smartregister.org/intervention/irs

image

cKellyDesign:
This PR renders the IRS draft plans list page at `/intervention/irs` in addition to `/intervention/irs/drafts`.

Fixes #685


moshthepitt:
@cKellyDesign what was the problem that caused the endless loading, that this is solving?  Also is it possible to get some test for this?

src/containers/pages/InterventionPlan/IRS/index.tsx
```diff
@@ -240,7 +240,8 @@ interface DispatchedStateProps {
 }
 
 const mapStateToProps = (state: Partial<Store>, ownProps: any): DispatchedStateProps => {
-  const isDraftsList = ownProps.path === INTERVENTION_IRS_DRAFTS_URL;
+  const isDraftsList =
+    ownProps.path === INTERVENTION_IRS_DRAFTS_URL || ownProps.path === INTERVENTION_IRS_URL;
```
cKellyDesign: Initial bug was caused by this conditional failing to recognize `/intervention/irs` as an IRS path that should render draft plans, so [the container didn't fetch plans from OpenSRP](https://github.com/onaio/reveal-frontend/blob/master/src/containers/pages/InterventionPlan/IRS/index.tsx#L111-L129).

cKellyDesign: An alternate fix would simply be removing the `/intervention/irs` from the App and updating the IRS link on the Home page to point to `/intervention/irs/drafts`.

cKellyDesign: @moshthepitt as for tests it looks like they're currently pretty limited, I imagine we could count the number of plans returned in the props and simply mount the component with different routes (relates to https://github.com/onaio/reveal-frontend/issues/669).

moshthepitt: @cKellyDesign for a test is it not possible to recreate what happened (to cause the issue) and confirm that now we get a different result?  It just feels weird to me that we are adding a fix that could later be removed accidentally or otherwise and we wouldn't know until "they" tell us oh no this bug is back magically.

cKellyDesign: @moshthepitt sorry, I mean we can count rows in both draft cases, I think to achieve the different cases we can mount `App` and push the history to navigate to all the URLs for which this container is rendered.

moshthepitt: > sorry, I mean we can count rows in both draft cases, I think to achieve the different cases we can mount App and push the history to navigate to all the URLs for which this container is rendered.

🙈 not clear what you mean, but I'll get the idea when we have some code.

cKellyDesign: 🙊 pole, yes I think it will be more clear when I commit some code :)

p-netm:
superseded by #766

PR #690, opened 2020-02-13
Test logout

### LINKED ISSUE #676: Test Logout
testing logout functionality added in #675

p-netm:
closes #676 
-  tests current logout functionality


src/configs/__mocks__/env.ts
```diff
@@ -89,3 +89,9 @@ export type LANGUAGE = typeof LANGUAGE;
 
 export const OAUTH_GET_STATE_URL = 'http://localhost:3000/oauth/state';
 export type OAUTH_GET_STATE_URL = typeof OAUTH_GET_STATE_URL;
+
+export const DISABLE_LOGIN_PROTECTION = false;
+
+export const REACT_APP_EXPRESS_OAUTH_GET_STATE_URL = 'http://localhost:3000/oauth/state';
+
+export const REACT_APP_EXPRESS_OAUTH_LOGOUT_URL = 'http://localhost:3000/logout';
```
moshthepitt: Do we need the `REACT_APP_` prefix?

cKellyDesign: Good catch

p-netm: refactored this. Just would like to add that the variables had been named like that in the envs.ts file, and that's why the test did not fail

Issue #656, opened 2020-02-03
Refactor ActiveFocusInvestigation


moshthepitt:
In support of #651, we need to refactor the `ActiveFocusInvestigation` component to work with jurisdictions.  i.e. this component should take a jurisdiction id and render that jurisdiction's FI activities.

moshthepitt:
Fixed in #692

PR #692, opened 2020-02-14
Refactor ActiveFocusInvestigation

moshthepitt:
Fix: #656

This PR refactors ActiveFocusInvestigation in the following ways:

- ActiveFocusInvestigation is now based off a jurisdiction as opposed to a plan.  This means that we retrieve a jurisdiction id (instead of plan id) from the URL and use it to fetch from APIs and render the component.
- ActiveFocusInvestigation is now a functional component
- The little map section is broken into its own component.
- The selectors use Reselect.  Reselect was itself introduced in this PR
- Various performance and code improvements within ActiveFocusInvestigation were made


src/helpers/cbv.tsx
```diff
@@ -0,0 +1,140 @@
...
+ * higher order component that has been connected to the redux store, and which
+ * knows how to fetch data for display.
+ *
+ * The end goal is a quick and convenient way to display a list of objects without
+ * having to worry about boilerplate code that deal with fetching objects and
+ * storing them in a Redux store.
+ *
+ * Every method in this class can and should be overridden to cater to custom needs.
+ */
+export class ObjectList {
```
moshthepitt: @cKellyDesign @p-netm this is a first attempt at solving part of https://github.com/onaio/js-tools/issues/126

We have a basic working version, but I need critical eyes looking at it.

- Does it meet the needs that we stated in that issue?
- What can we improve?
- Is it easy and natural to use?
- What would you do differently?

PS - it is definitely not done, and the final version will be a package in js-tools.

src/containers/pages/FocusInvestigation/jurisdiction/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+  messages: Message[];
+}
+
+const options = {
+  dispatchPropName: 'loadMessages',
+  listPropName: 'messages',
+};
+
+/** this is our dumb component */
+const FIJurisdiction = (props: Props) => {
```
moshthepitt: @p-netm @cKellyDesign this is sample usage of `ObjectList`. How do you feel about this?  What can be better?

p-netm: I think the reducerName would have to be placed somewhere different, at the moment, i feel like its more natural to have the dumb component either statically define the reducerName or pass an option to the HOC that tells it what reducer slice to register or use

src/helpers/cbv.tsx
p-netm: - What would i do differently
I would probably keep data fetching in the dumb component. I am thinking of places where we have more than a single fetch data request, places where the request is part of the component's lifecycle or as a result of a user interaction.

i understand we can override the current `fetchData` . I at the moment i can only speculate that this would result in a bit of added complexity.

**How would i work around this**
data fetching would remain colocated with dumb component as it is now.
The dumb component would thus need to be given the necessary actionCreators, that it needs

moshthepitt: @p-netm I thought about that a bit.  The way I thought about it was that the process of fetching the data goes like this:

- call some URL in a very specific way (this varies a lot between components)
- while the URL is loading, display a loading indicator
- once done, show the data

I think we should DRY this up and standardize it, and by "this" I mean the whole process.  i.e. you may want to call a specific URL in a specific way which is cool, but we want to standardize what happens while you call it and after.

Thoughts on that?

src/containers/pages/FocusInvestigation/jurisdiction/index.tsx
moshthepitt: You're right.  This part I had not worked on yet.  It also ties in nicely with what we talked about the other day i.e. generic ducks

moshthepitt:
@p-netm @cKellyDesign after `yanr start`, you can see this working by visiting http://localhost:3000/focus-investigation/jurisdiction

cKellyDesign:
> http://localhost:3000/focus-investigation/jurisdiction

@moshthepitt I'm getting an authentication error when trying to load this page(after logging in).
Screen Shot 2020-02-14 at 9 14 50 AM

Am I missing some env variables or something?

moshthepitt:
@ckellydesign yeah you can safely ignore that, it's caused by something else and the fact that this page doesn't require log in yet

src/helpers/cbv.tsx
cKellyDesign: > I would probably keep data fetching in the dumb component.

@p-netm and @moshthepitt - I agree and disagree with this, I think there are two cases in which we are calling the API:
1. When a component mounts (everything on inital mounting)
2. After a component mounts (later lifecycle methods, user interaction, etc)
In my opinion, `2` should definitely remain in the dumb components while `1` should be the responsibility of the `cbv`.

To accomplish this while keeping things modular and DRY, could there be a way to pass options into `cbv` which tell it:
* which API endpoints to to call on behalf of the dumb component
* what options should be used when making the call
* what to do when the call is completed
* what to do if the call fails

If we pass an array of such options, then the `cbv` can generate a list of Promises to `resolve` or `reject`. With this approach it seems the `cbv` could also manage an `isLoading` state which is passed into the dumb component props.

What do you think of this idea?

src/containers/pages/FocusInvestigation/jurisdiction/index.tsx
cKellyDesign: > i feel like its more natural to have the dumb component either statically define the reducerName or pass an option to the HOC that tells it what reducer slice to register or use

Agreed!

src/helpers/cbv.tsx
cKellyDesign: What gets tricky is the business logic that often comes with the pages, like if the results from request A are required to configure request B.

moshthepitt: @p-netm @cKellyDesign okay so I'll leave this out for now.  Feels like we would be trying to make something DRY before we fully understand how we will be using this.  We'll revisit this when the usage patterns and opportunities for optimization become clearer.

p-netm: I am satisfied with this decision

src/containers/pages/FocusInvestigation/jurisdiction/JurisdictionList.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { ObjectList } from '../../../../helpers/cbv';
+import {
+  InterventionType,
+  makePlansArraySelector,
+  PlanFilters,
+  PlanStatus,
+} from '../../../../store/ducks/plans';
+
+/** Custom implementation of ObjectList where we override the getMapStateToProps method */
+export default class JurisdictionList<
```
moshthepitt: @cKellyDesign @p-netm for more complex cases like this component, you have to override one or more methods on ObjectList like I am doing here.  What are your feelings/thoughts about this?

src/containers/pages/FocusInvestigation/jurisdiction/JurisdictionList.tsx
```diff
@@ -0,0 +1,46 @@
...
+  PropsType
+> extends ObjectList<ObjectType, ActionType, SelectorType, PropsType> {
+  /** Custom getMapStateToProps
+   * Gets plans for the current jurisdiction
+   */
+  public getMapStateToProps() {
+    return (state: Registry, ownProps: any) => {
+      const getPlansArray = makePlansArraySelector();
+
+      const jurisdictionId =
```
moshthepitt: This is hard coded for now, during dev.  Will be removed in final build.

src/containers/pages/FocusInvestigation/jurisdiction/index.tsx
```diff
@@ -0,0 +1,370 @@
...
+ * given jurisdiction.  It is meant to display for the focus area i.e. the
+ * lowest level jurisdiction
+ */
+const FIJurisdiction = (props: FIJurisdictionProps & RouteComponentProps<RouteParams>) => {
+  const [loading, setLoading] = useState(true);
+
+  const { caseTriggeredPlans, fetchPlansActionCreator, routinePlans, supersetService } = props;
+
+  const jurisdictionId =
+    '59ad4fa0-1945-4b50-a6e3-a056a7cdceb2' || props.match.params.jurisdictionId;
```
moshthepitt: This is hard coded for now, during dev.  Will be removed in final build.

src/containers/pages/FocusInvestigation/jurisdiction/index.tsx
```diff
@@ -0,0 +1,370 @@
...
+    { comparator: InterventionType.FI, operator: '==', subject: 'plan_intervention_type' },
+  ]);
+
+  useEffect(() => {
+    supersetService(SUPERSET_PLANS_SLICE, supersetParams)
+      .then((result: Plan[]) => {
+        if (result) {
+          fetchPlansActionCreator(result);
+        } else {
+          displayError(new Error('An error occurred'));
```
moshthepitt: @cKellyDesign @p-netm  Noticed that we don't show any errors in cases of HTPP errors (think 4xx/5xx).  This needs to be generalised some more, and I might do that in this issue.  Ideas welcome.

src/containers/pages/FocusInvestigation/jurisdiction/index.tsx
```diff
@@ -0,0 +1,370 @@
...
+      })
+      .finally(() => setLoading(false))
+      .catch(err => displayError(err));
+  }, []);
+
+  if (loading === true) {
+    return <Loading />;
+  }
+
+  /** UNKNOWN */
```
moshthepitt: Just here during dev.  Will not be in final build.

src/containers/pages/FocusInvestigation/jurisdiction/JurisdictionList.tsx
cKellyDesign: Seems like a pretty straight forward inheritance pattern, I love it!

src/containers/pages/FocusInvestigation/jurisdiction/index.tsx
cKellyDesign: For service errors (4xx/5xx upon fetch), we could parameterize the error messages by passing in props/object to override default error messages based on the response code.

```
const customErrorMessages:ErrorMessages = {
  500: 'Whoops, something went way wrong!',
  401: 'Whoops, looks like you need to log in again!',
  400: () => { return 'Whoops, looks like `${jurisdiction_name}` doesn't have any active tasks`; }
}

File: src/helpers/cbv.tsx [hunk 8]
```diff
@@ -0,0 +1,108 @@
...
+ */
+export class ObjectList<ObjectType, ActionType, SelectorType, PropsType, RootState = Registry> {
+  public Component: React.ElementType;
+  public options: ObjectListOptions<ActionType, SelectorType>;
+
+  /** constructor */
+  constructor(component: React.ElementType, options: ObjectListOptions<ActionType, SelectorType>) {
+    this.Component = component;
+    this.options = options;
+  }
```
cKellyDesign: Cool so the actual `JurisdictionList` component gets passed in here

src/helpers/cbv.tsx
```diff
@@ -0,0 +1,108 @@
...
+   * This function returns a Higher Order component whose job is to wrap around
+   * the target component, and pass on props to it.  The props in this case are:
+   *     - the list of objects to be displayed
+   *     - the action creator dispatch function
+   */
+  public getHOC() {
+    return (props: ObjectListProps<ActionType, ObjectType>) => {
+      return <this.Component {...props} />;
+    };
+  }
```
cKellyDesign: And the `Component` get's wrapped in the `HOC` here

src/helpers/cbv.tsx
```diff
@@ -0,0 +1,108 @@
...
+  public getConnectedHOC() {
+    /** map dispatch to props */
+    const mapDispatchToProps = this.getMapDispatchToProps();
+    /** map state to props */
+    const mapStateToProps = this.getMapStateToProps();
+    /** connect to store */
+    return connect(
+      mapStateToProps,
+      mapDispatchToProps
+    )(this.getHOC());
```
cKellyDesign: And the `HOC` gets connected to the `store` here

src/helpers/cbv.tsx
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  /**
+   * This function returns the connected higher order component.
+   */
+  public render() {
+    return this.getConnectedHOC();
+  }
```
cKellyDesign: And the `connectedHOC` is what is finally returned on `render()` for the `FocusInvestigation/jurisdiction` container default.

src/helpers/cbv.tsx
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * The class's render method will then return the same component wrapped in a
+ * higher order component that has been connected to the redux store, and which
+ * knows how to fetch data for display.
+ *
+ * The end goal is a quick and convenient way to display a list of objects without
+ * having to worry about boilerplate code that deal with fetching objects and
+ * storing them in a Redux store.
+ *
+ * Every method in this class can and should be overridden to cater to custom needs.
```
cKellyDesign: `overridden` => `overwritten`

src/containers/pages/FocusInvestigation/jurisdiction/index.tsx
```diff
@@ -0,0 +1,371 @@
...
+
+/** ObjectList options */
+const objectListOptions = {
+  actionCreator: fetchPlans,
+  dispatchPropName: 'fetchPlansActionCreator',
+  listPropName: 'routinePlans',
+  selector: getPlansArray,
+};
+
+const cbv = new JurisdictionList<
```
cKellyDesign: Can we namespace this `const` something a bit more specific? 🙏

src/helpers/cbv.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+  /**
+   * The mapStateToProps function
+   * You may override this for more custom needs.
+   */
+  public getMapStateToProps() {
+    return (state: RootState, ownProps: PropsType) => {
+      // we have to use a Type Guard to check if this.options.selector is a
+      // callable/function otherwise Typescript will infer its type as "unknown"
+      // TODO: look into whether there is a better fix for this
```
cKellyDesign: 👀

src/containers/JurisdictionMap/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+/** SingleObject options */
+const objectListOptions = {
+  actionCreator: fetchJurisdictions,
+  dispatchPropName: 'fetchJurisdictionsActionCreator',
+  returnPropName: 'jurisdiction',
+  selector: jurisdictionByIdSelector,
+};
+
+const ConnectedJurisdictionMap = new SingleObject<
```
moshthepitt: @p-netm @cKellyDesign this is a simple usage of `SingleObject` - in this case you don't need to override anything in the class.

src/helpers/CBV/ObjectList.tsx
```diff
@@ -0,0 +1,74 @@
...
+ * higher order component that has been connected to the redux store, and which
+ * knows how to fetch data for display.
+ *
+ * The end goal is a quick and convenient way to display a list of objects without
+ * having to worry about boilerplate code that deal with fetching objects and
+ * storing them in a Redux store.
+ *
+ * Every method in this class can and should be overwritten to cater to custom needs.
+ */
+export class ObjectList<
```
moshthepitt: @cKellyDesign @p-netm ObjectList now extends "Base"

src/helpers/CBV/SingleObject.tsx
```diff
@@ -0,0 +1,74 @@
...
+ * knows how to fetch data for display.
+ *
+ * The end goal is a quick and convenient way to display a singular objects without
+ * having to worry about boilerplate code that deal with fetching it and
+ * storing it in a Redux store.
+ *
+ * Every method in this class can and should be overwritten to cater to custom needs.
+ */
+// export class SingleObject<ObjectType, ActionType, SelectorType, PropsType, RootState = Registry> {
+export class SingleObject<
```
moshthepitt: @cKellyDesign @p-netm we now have `SingleObject` that is concerned with components that only want one object from the store.

moshthepitt:
@cKellyDesign @p-netm it needs cleanup, tests, etc but `FIJurisdiction` now works:

Focus Investigations in ra_SND_51   Reveal

If you get a chance please look at it and point out any big mis-steps or mistakes.  Particularly regarding:

- ObjectList class and usage
- SingleObject class and usage
- Reselect usage

src/configs/strings/en.json
```diff
@@ -870,5 +870,13 @@
...
     "message": "Link",
     "description": "URL leading to another page"
+  },
+  "AN_ERROR_OCCURRED": {
+    "message": "An error occurred",
+    "description": "Generic error message"
+  },
+  "JURISDICTION_LOADING_ERROR": {
+    "message": "An error prevented this jurisdiction from loading.",
+    "description": "Error message displayed when jurisdiction cannot be displayed correctly"
```
cKellyDesign: Let's also be sure to request these translations 🙏

src/containers/pages/FocusInvestigation/jurisdiction/index.tsx
cKellyDesign: Ooh let's use the translated string here `AN_ERROR_OCCURRED` 🙏

src/containers/pages/FocusInvestigation/jurisdiction/index.tsx
```diff
@@ -0,0 +1,434 @@
...
+  useEffect(() => {
+    supersetService(SUPERSET_PLANS_SLICE, supersetParams)
+      .then((result: Plan[]) => {
+        if (result) {
+          fetchPlansActionCreator(result);
+        } else {
+          displayError(new Error('An error occurred'));
+        }
+      })
+      .finally(() => setLoading(false))
```
cKellyDesign: Nice use of `finally()`!

moshthepitt:
@cKellyDesign tests for the CBV are still pending.  Will do those tomorrow.  Thanks for the review so far!

src/configs/strings/en.json
moshthepitt: How do we do that usually @cKellyDesign ?

cKellyDesign: I think [we can just add them to the spreadsheet here](https://docs.google.com/spreadsheets/d/1ONHpPpuPxNW9_-vP8gTTcrVHTbKTT3WaN-cfdySeMCs/edit?ts=5e0d6060#gid=0) and let @craigappl and/or @samkanga know that there are more display strings to be translated so they can let the client know to provide them.

moshthepitt:
@cKellyDesign @p-netm  this is now ready for re-Review.  Note that this is being merged not into master but its own parent branch where I'll need to write more tests there.  The things to review for now are:

- ActiveFocusInvestigation used to be based off SingleFI which this PR removes
- ActiveFocusInvestigation now is based on a new component called JurisdictionFI.   This means that we retrieve a jurisdiction id (instead of plan id) from the URL and use it to fetch from APIs and render the component.
- The little map section is broken into its own component called JurisdictionMap.
- JurisdictionFI and JurisdictionMap are powered by ObjectList and SingleObject respectively
- The new selectors use Reselect. Reselect was itself introduced in this PR
- Tests have been added for:
    - JurisdictionFi
    - JurisdictionMap
    - ObjectList
    - SingleObject
    - All new selectors

moshthepitt:
@cKellyDesign @p-netm  I have added tests for CBV here.  But once we merge to the other branch, I'll move it to js-tools.

src/helpers/CBV/tests/ObjectList.test.tsx
```diff
@@ -0,0 +1,130 @@
...
+    for (const Element of toTest) {
+      store.dispatch(removeMessagesAction());
+      store.dispatch(sendMessage({ user: 'bob', message: 'hello' }));
+      store.dispatch(sendMessage({ user: 'bobbie', message: 'hello hello' }));
+
+      const wrapper = mount(
+        <Provider store={store}>
+          <Element />
+        </Provider>
+      );
```
cKellyDesign: Do we need to / should we unmount `wrapper` at the end of this `for` block?

src/helpers/CBV/tests/ObjectList.test.tsx
```diff
@@ -0,0 +1,130 @@
...
+/** register reducers */
+reducerRegistry.register(reducerName, reducer);
+
+interface TestProps {
+  actionCreator?: typeof sendMessage;
+  callbackFunc?: () => void;
+  messages?: Message[];
+}
+
+const TestComponent = (_: TestProps) => <div>mosh</div>;
```
cKellyDesign: Would it make sense to have this `TestComponent` be a list of elements driven by `props.messages` since it's a test for `ObjectList`?

src/helpers/CBV/tests/ObjectList.test.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+      expect(wrapper.find('Connect(HoC)').props()).toEqual({});
+      expect(wrapper.find('Connect(HoC)>HoC').props()).toEqual({
+        ...expected,
+        actionCreator: expect.any(Function),
+      });
+      expect(wrapper.find('Connect(HoC)>HoC>TestComponent').props()).toEqual({
+        ...expected,
+        actionCreator: expect.any(Function),
+      });
```
cKellyDesign: 👏 Nice props testing! If the `TestComponent` renders a list of objects, could we add a line or two testing that `TestComponent` is rendering correctly with the props it's given? If the component renders a simple list, maybe something like:

```
expect(wrapper.find('li').length).toEqual(2)
```

Not sure if this actually increases test coverage though?

src/helpers/CBV/tests/SingleObject.test.tsx
```diff
@@ -0,0 +1,128 @@
...
+/** register reducers */
+reducerRegistry.register(reducerName, reducer);
+
+interface TestProps {
+  actionCreator?: typeof sendMessage;
+  callbackFunc?: () => void;
+  messages?: Message[];
+}
+
+const TestComponent = (_: TestProps) => <div>mosh</div>;
```
cKellyDesign: Should the `TestComponent` render something from the props? Does that increase test coverage or is that kind of redundant after testing the props alone?

p-netm:
by the way, what does cbv stand for, am assuming  its an acronym

cKellyDesign:
> what does cbv stand for

@p-netm I believe it stands for Class Based View

@moshthepitt Is that right? Perhaps we could include this in a `src/helpers/CBV/readme.md` along with some simple docs on how to use it?

src/helpers/CBV/tests/ObjectList.test.tsx
moshthepitt: I don't think that we need to do so, but I'll do it anyway.

src/helpers/CBV/tests/ObjectList.test.tsx
moshthepitt: I'll do that for clarity in the test.  Thanks.

src/helpers/CBV/tests/ObjectList.test.tsx
moshthepitt: It will not improve the test coverage (its already 100%) but I think it is useful for clarity.  I'll add this.

src/helpers/CBV/tests/SingleObject.test.tsx
moshthepitt: It will not improve the test coverage (its already 100%) but I think it is useful for clarity.  I'll add this.

moshthepitt:
> Is that right? Perhaps we could include this in a src/helpers/CBV/readme.md along with some simple docs on how to use it?

@p-netm @cKellyDesign this is correct.  We'll have the README when we have this as a package in js-tools.

PR #696, opened 2020-02-18
Router prompt tests

### LINKED ISSUE #609: Add tests for router prompts in team assignment
Currently we are using alert prompts to notify the user they are navigating away from unsaved changes and we need a way to add tests to this functionality.

ciremusyoka:
Closes #609 

Testing the actual functionality of the `React Router Prompt component` is somehow tricky because have not found a way to simulate the `Prompt component` buttons. However, i think that react has already tested that hence this test checks whether it is possible for a users to navigate to other routes when they have unsaved changes.


src/containers/pages/OrganizationViews/AssignPractitioners/tests/index.test.tsx
```diff
@@ -216,6 +216,59 @@ describe('src/pages/*/AssignPractitioners', () => {
...
+    // simulate single value change function
+    const entry = fixtures.practitioner2;
+    (select.instance() as any).selectOption({
+      label: `${entry.username} - ${entry.name}`,
+      value: entry.identifier,
+    });
+    wrapper.update();
+
+    // try navigating to login
+    history.push(FI_URL);
```
moshthepitt: Does `FI_URL` represent the login URL?

ciremusyoka: No, have rectified the comment

src/containers/pages/OrganizationViews/AssignPractitioners/tests/index.test.tsx
```diff
@@ -216,6 +216,59 @@ describe('src/pages/*/AssignPractitioners', () => {
...
+    (select.instance() as any).selectOption({
+      label: `${entry.username} - ${entry.name}`,
+      value: entry.identifier,
+    });
+    wrapper.update();
+
+    // try navigating with usaved changes
+    history.push(FI_URL);
+    let locIsLogin = location.pathname === FI_URL;
+    expect(locIsLogin).toBeFalsy();
```
moshthepitt: @ciremusyoka at this point we should have an alert pop up, right?  It should be possible to test that this happened, and that the alert contained the correct message.

cKellyDesign: Agreed, this far it doesn't look like these tests are testing for the alert.

PR #698, opened 2020-02-18
Add tests for component AssignTeamPopover

### LINKED ISSUE #517: Add more tests for TeamAssignment popover
Currently the TeamAssignmentPopover has only a shallow tests and needs more substantial unit tests, however when mounting in jest the Component throws an error complaining it can't find its `target` in the DOM.

```
  it('renders correctly', () => {
    const mock: any = jest.fn();
    const props:AssignTeamPopoverProps = {
      formName: 'plan-assignment-form-outpost-number-one',
      isActive: true,
      jurisdictionId: 'outpost-number-one',
      onClearAssignmentsButtonClick: mock,
      onSaveAssignmentsButtonClick: mock,
      onToggle: mock,
      organizationsById: null,
      planId: 'alpha',
      target: 'plan-assignment-outpost-number-one',
    }

    // attempted fix from https://github.com/reactstrap/reactstrap/issues/773#issuecomment-373451256
    var div = document.createElement('div');
    document.body.appendChild(div);

    // this is breaking on mount because the underlying Popover component can't find the button
    const wrapper = mount((
      <div>
        <button id="plan-assignment-outpost-number-one">Target Button</button>
        <AssignTeamPopover {...props} />
      </div>
    ), { attachTo: div });
  })
```

kelvin-muchiri:
Closes #517

### DIFF
--- a/src/components/AssignTeamPopover/tests/__snapshots__/index.test.tsx.snap
+++ b/src/components/AssignTeamPopover/tests/__snapshots__/index.test.tsx.snap
@@ -182,3 +182,39 @@ exports[`/components/AssignTeamPopover renders and passes props correctly 1`] = 
   </TooltipPopoverWrapper>
 </Popover>
 `;
+
+exports[`/components/AssignTeamPopover renders correctly when organizationsById is not null: button clear props 1`] = `
+Object {
+  "children": "Clear",
+  "color": "default",
+  "onClick": [MockFunction],
+  "outline": true,
+  "size": "xs",
+  "tag": "button",
+}
+`;
+
+exports[`/components/AssignTeamPopover renders correctly when organizationsById is not null: button save props 1`] = `
+Object {
+  "children": "Save",
+  "color": "primary",
+  "onClick": [MockFunction],
+  "size": "xs",
+  "tag": "button",
+}
+`;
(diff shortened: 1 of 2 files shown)


src/components/AssignTeamPopover/tests/__snapshots__/index.test.tsx.snap
```diff
@@ -0,0 +1,184 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
moshthepitt: Do you think that this snapshot can be made smaller, in the spirit of https://kentcdodds.com/blog/effective-snapshot-testing ?

kelvin-muchiri: > Do you think that this snapshot can be made smaller, in the spirit of https://kentcdodds.com/blog/effective-snapshot-testing ?

Yes. After reading some articles I learnt that one should make the assumption that the children components were tested and the parent should only care if they exist

kelvin-muchiri: I will be pushing an update and tell me what you think

PR #701, opened 2020-02-18
Reduce snapshot size

kelvin-muchiri:
Reduce snapshot sizes. Use more explicit assertions.


moshthepitt:
@kelvin-muchiri was this linked to an open issue?  If so, please close that issue.

Issue #702, opened 2020-02-18
Replace getPlanRecordsArray


moshthepitt:
Following #699, we now need to fully replace `getPlanRecordsArray` with `plansArraySelector`.

ciremusyoka:
@moshthepitt  The plans filter for `plansArraySelector` is done on `plansById`  while for `getPlanRecordsArray` is done on `planRecordsById`. There is a significant difference between [plansById](https://github.com/onaio/reveal-frontend/blob/8108f5f4d843c925ec2e934afb8a82d11efbeb4f/src/store/ducks/plans.ts#L70-L77) and [planRecordsById](https://github.com/onaio/reveal-frontend/blob/8108f5f4d843c925ec2e934afb8a82d11efbeb4f/src/store/ducks/plans.ts#L53-L66) making it hard to combine them.

To accommodate filtering of `planRecordsById` or any other objects inside plans, have refactored the [makePlansArraySelector](https://github.com/onaio/reveal-frontend/blob/8108f5f4d843c925ec2e934afb8a82d11efbeb4f/src/store/ducks/plans.ts#L667-L678) function and it's input-selectors. 

Any thoughts on how to make it better or is it OK?

moshthepitt:
@ciremusyoka looks like it can work.  Go ahead and complete it and lets see what the PR looks like.

moshthepitt:
Fixed in #733

PR #651, opened 2020-01-31
Add link to focus area from add plan page

moshthepitt:
Fix: #649

Adds a link to the affected focus area when adding a new FI plan.

PR #703, opened 2020-02-18
Add Reselect

### LINKED ISSUE #699: Use Reselect
We should adopt [Reselect](https://github.com/reduxjs/reselect) for [these reasons](https://github.com/reduxjs/reselect) i.e. a summary:

- Selectors can compute derived data, allowing Redux to store the minimal possible state.
- Selectors are efficient. A selector is not recomputed unless one of its arguments changes.
- Selectors are composable. They can be used as input to other selectors.

This will also greatly help in implementing: https://github.com/onaio/js-tools/issues/126

moshthepitt:
This PR adds Reselect, thereby Fixes #699.

Reselect is added in the plans reducer module in a way that the new selector, `plansArraySelector`,  effectively can fully replace `getPlanRecordsArray`.

`getPlanRecordsArray` should be fully replaced via #702

This is meant to support #692 and #651

### DIFF
--- a/src/store/ducks/plans.ts
+++ b/src/store/ducks/plans.ts
@@ -588,6 +588,18 @@ export const getPlansArrayByParentJurisdictionId = createSelector(
     )
 );
 
+export const plansArraySelector = createSelector(
+  [
+    getPlansArrayByInterventionType,
+    getPlansArrayByJurisdictionIds,
+    getPlansArrayByStatus,
+    getPlansArrayByReason,
+    getPlansArrayByParentJurisdictionId,
+  ],
+  (plans, plans2, plans3, plans4, plans5) =>
+    intersect([plans, plans2, plans3, plans4, plans5], JSON.stringify)
+);
+
 export const getPlansArrayByInterventionTypeAndJurisdictionId = createSelector(
   [getPlansArrayByInterventionType, getPlansArrayByJurisdictionIds],
   (plans, plans2) => intersect([plans, plans2], JSON.stringify)
--- a/src/store/ducks/plans.ts
+++ b/src/store/ducks/plans.ts
@@ -599,8 +599,3 @@ export const plansArraySelector = createSelector(
   (plans, plans2, plans3, plans4, plans5) =>
     intersect([plans, plans2, plans3, plans4, plans5], JSON.stringify)
 );
-
-export const getPlansArrayByInterventionTypeAndJurisdictionId = createSelector(
-  [getPlansArrayByInterventionType, getPlansArrayByJurisdictionIds],
-  (plans, plans2) => intersect([plans, plans2], JSON.stringify)
-);
--- a/src/store/ducks/plans.ts
+++ b/src/store/ducks/plans.ts
@@ -522,7 +522,7 @@ export function getPlanRecordById(state: Partial<Store>, id: string): PlanRecord
 
 export interface PlanFilters {
   interventionType?: InterventionType;
-  jurisdictionId?: string;
+  jurisdictionIds?: string[];
   parentJurisdictionId?: string;
   reason?: FIReasonType;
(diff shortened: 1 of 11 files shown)


src/store/ducks/plans.ts
```diff
@@ -515,3 +518,150 @@ export function getPlanRecordsIdArray(
...
+ *    - plan status
+ *    - FI plan reason
+ *    - plan jurisdiction parent_id
+ *
+ * These filter params are all optional and are supplied via the prop parameter.
+ *
+ * This selector is meant to be a memoized replacement for getPlanRecordsArray.
+ *
+ * @param state - the redux store
+ * @param props - the plan filters object
```
cKellyDesign: @moshthepitt Let's be sure to add the Typing for `props` here, especially for these created Selectors since the props/returns of the resultant Selector are not intuitive from what's being passed into `createSelector()`

moshthepitt: @cKellyDesign good catch!

moshthepitt:
> @moshthepitt I like the approach here, it seems like a good way to apply compound filters to get very select plans. Maybe we can include some further documentation somewhere on how to set up other ducks modules in this way, but that doesn't necessarily need to happen in this PR

I can do this, but how/where do you think this can go?

cKellyDesign:
@moshthepitt how about in a new `/src/store/ducks/readme.md`?

moshthepitt:
@cKellyDesign ready for re-review

p-netm:
@moshthepitt @cKellyDesign Is this something we would want to consider:
 https://github.com/reduxjs/reselect#sharing-selectors-with-props-across-multiple-component-instances

p-netm:
> @moshthepitt how about in a new `/src/store/ducks/readme.md`?

@cKellyDesign @moshthepitt in what would this be different to the reselect documentation itself?

moshthepitt:
> @moshthepitt @cKellyDesign Is this something we would want to consider:
https://github.com/reduxjs/reselect#sharing-selectors-with-props-across-multiple-component-instances

Glad you brought this up

cKellyDesign:
> > @moshthepitt how about in a new /src/store/ducks/readme.md?
>
> @cKellyDesign @moshthepitt in what would this be different to the reselect documentation itself?

@p-netm Personally I'm finding the reselect documentation to be pretty dense and hard to understand the nuances. Perhaps such a readme could include:
* a more light-weight example in the context of our dux modules / containers
* some sort of simple flow diagram to describe the relation between `reselect. createSelector`, the selectors which are actually returned, the getters, and the store
* links to specific parts of the reselect documentation which might otherwise be easy to miss (like [the point you brought up above](https://github.com/onaio/reveal-frontend/pull/703#issuecomment-587515828))

How does that sound?

p-netm:
> > > @moshthepitt how about in a new /src/store/ducks/readme.md?
> > 
> > 
> > @cKellyDesign @moshthepitt in what would this be different to the reselect documentation itself?
> 
> @p-netm Personally I'm finding the reselect documentation to be pretty dense and hard to understand the nuances. Perhaps such a readme could include:
> 
>     * a more light-weight example in the context of our dux modules / containers
> 
>     * some sort of simple flow diagram to describe the relation between `reselect. createSelector`, the selectors which are actually returned, the getters, and the store
> 
>     * links to specific parts of the reselect documentation which might otherwise be easy to miss (like [the point you brought up above](https://github.com/onaio/reveal-frontend/pull/703#issuecomment-587515828))
> 
> 
> How does that sound?

Yeah, totally agree.

moshthepitt:
@cKellyDesign @p-netm documentation here https://github.com/onaio/reveal-frontend/pull/706

moshthepitt:
@p-netm please re-review this, I added one creator selector function.

PR #708, opened 2020-02-19
Solve bug map not loading on page refresh

### LINKED ISSUE #694: RVL-880 Map does not load on refreshing page
map view in SiingleActiveFI loads fine on the first load, however when you either
- refresh the page or
- copy the url and paste in a new tab and load it

the map does not render, it just keeps on loading

more info: https://smartregister.atlassian.net/browse/RVL-880

kelvin-muchiri:
Closes #694 

The bug was solved by checking if props.plans is null or not in componentDidMount. If not null, then we know the plan was received from the state when the user accessed the page via the link from the list of plans. If props.plans is null, then the page was refreshed, fetch plans and then fetch the rest of the data required.


moshthepitt:
@kelvin-muchiri could you include a little info in this PR on what the bug was and how this solves it?

kelvin-muchiri:
> @kelvin-muchiri could you include a little info in this PR on what the bug was and how this solves it?

@moshthepitt I have updated the description

kelvin-muchiri:
> How can we add a test for this fix?

I think we can add tests by checking when props.plan is null and when props.plan is not null should give the same result.

server/package.json
```diff
@@ -15,6 +15,7 @@
     "@types/helmet": "^0.0.45",
     "compression": "^1.7.4",
     "cookie-parser": "^1.4.4",
+    "cors": "^2.8.5",
```
cKellyDesign: Is this package required for the resolution of #694?

server/src/index.ts
```diff
@@ -35,12 +36,12 @@ const opensrpAuth = new ClientOAuth2({
   scopes: ['read', 'write'],
   state: EXPRESS_OPENSRP_OAUTH_STATE,
 });
-
 const loginURL = EXPRESS_SESSION_LOGIN_URL || '/';
 const sessionName = EXPRESS_SESSION_NAME || 'session';
 
 const app = express();
 
+app.use(cors());
```
cKellyDesign: I don't know what the default configs are for [cors](https://www.npmjs.com/package/cors), but should we be passing in options to limit / control resource requests?

src/containers/pages/FocusInvestigation/map/active/index.tsx
```diff
@@ -162,51 +162,27 @@ class SingleActiveFIMap extends React.Component<
...
-        SUPERSET_MAX_RECORDS,
-        [{ comparator: plan.plan_id, operator: '==', subject: 'plan_id' }],
-        { action_prefix: true }
-      );
-      /** Implement Ad hoc Queries since jurisdictions have no plan_id */
-      await supersetService(SUPERSET_STRUCTURES_SLICE, jurisdictionsParams).then(
-        (structuresResults: Structure[]) => {
-          fetchStructuresActionCreator(structuresResults);
-        }
+      await supersetService(SUPERSET_PLANS_SLICE, supersetParams).then((result: Plan[]) =>
```
cKellyDesign: If this is the last thing happening in `componentDidMount` then it doesn't need to `await` and we can drop `async` on the life-cycle method. 👍

src/containers/pages/FocusInvestigation/map/active/index.tsx
```diff
@@ -162,51 +162,27 @@ class SingleActiveFIMap extends React.Component<
...
-      });
-      await supersetService(SUPERSET_TASKS_SLICE, supersetParams).then((result4: Task[]) => {
-        fetchTasksActionCreator(result4);
-      });
+    } else {
+      /**
+       * Plans are available in state since the plans were fetched by the
+       * the previous view
+       */
+      this.fetchData();
```
cKellyDesign: Noice! 🎉

src/containers/pages/FocusInvestigation/map/active/index.tsx
```diff
@@ -217,6 +193,13 @@ class SingleActiveFIMap extends React.Component<
     }
   }
 
+  public componentDidUpdate(prevProps: any) {
+    if (!prevProps.plan && this.props.plan) {
+      /** Page was reloaded, fetch data again */
+      this.fetchData();
+    }
+  }
```
cKellyDesign: Can we add this conditional to check `prevProps.plan` and `this.props.plan` IDs so we can fetch data again if the `plan` changes?

src/containers/pages/FocusInvestigation/map/active/index.tsx
```diff
@@ -403,6 +386,53 @@ class SingleActiveFIMap extends React.Component<
...
+      fetchStructuresActionCreator,
+      fetchTasksActionCreator,
+      plan,
+      supersetService,
+    } = this.props;
+
+    if (plan && plan.plan_id) {
+      /** define superset filter params for jurisdictions */
+      const jurisdictionsParams = superset.getFormData(SUPERSET_MAX_RECORDS, [
+        { comparator: plan.jurisdiction_id, operator: '==', subject: 'jurisdiction_id' },
```
cKellyDesign: Is `jurisdiction_id` a magic string in this case? I think so, please move the `subject` strings into the `/src/constants.tsx` file 🙏

server/package.json
kelvin-muchiri: > Is this package required for the resolution of #694?

I think we had added that with @p-netm when troubleshooting when I had trouble setting `tiny-server` up.  we later found out the problem was `REACT_APP_DISABLE_LOGIN_PROTECTION=false` had been set to true

server/src/index.ts
```diff
@@ -79,7 +78,7 @@ class HttpException extends Error {
     this.message = message;
   }
 }
-
+/* tslint:disable:no-unused-variable */
```
moshthepitt: Please remove the unused variable or replace with `_` instead of disabling the linter.

src/containers/pages/FocusInvestigation/map/active/index.tsx
```diff
@@ -161,52 +163,28 @@ class SingleActiveFIMap extends React.Component<
...
-      });
+      supersetService(SUPERSET_PLANS_SLICE, supersetParams)
+        .then((result: Plan[]) => fetchPlansActionCreator(result))
+        .catch(error => displayError(error));
+    } else {
+      /**
+       * Plans are available in state since the plans were fetched by the
+       * the previous view
+       */
+      this.fetchData().catch(error => displayError(error));
```
cKellyDesign: Does this actually catch the errors if no promises are being returned in `this.fetchData()`?

kelvin-muchiri: @cKellyDesign Tslint raises a flag if there's no catch block
Screenshot from 2020-02-21 18-03-48

cKellyDesign: @kelvin-muchiri I definitely think it should have a `catch` block, I just want to make sure that if an error is encountered within `fetchData()`, it's thrown correctly and get's caught by this `catch`. Can we add a test for such a scenario?

moshthepitt: God bless this more aggressive `tslist` :) #amiright? :)

kelvin-muchiri: > Does this actually catch the errors if no promises are being returned in `this.fetchData()`?

Apparently we have to add a `then` or `catch` block else the compiler thinks we forgot to handle our async code. TSlint was raising errors (https://stackoverflow.com/a/43981878/3247914)

kelvin-muchiri: > @kelvin-muchiri I definitely think it should have a `catch` block, I just want to make sure that if an error is encountered within `fetchData()`, it's thrown correctly and get's caught by this `catch`. Can we add a test for such a scenario?

Added tests

PR #709, opened 2020-02-19
Add readme detailing the authorization code flow

### LINKED ISSUE #705: Add flow diagram for the tiny backend server
This comes from [here](https://github.com/onaio/reveal-frontend/pull/624#issuecomment-583019650).

>  It would be really nice to see a flow-diagram describing the authentication and logout flows implemented by this PR. Also any implementation docs would be good, but maybe these belong in the Gatekeeper package itself?

This diagram and a little README should go into the server directory.

p-netm:
- Add a basic flow diagram 
- Add readme detailing the authorization code flow to supplement a forementioned basic
- Fix #705


moshthepitt:
I havent yet looked into it in depth, but I would name this file README.md

p-netm:
> I havent yet looked into it in depth, but I would name this file README.md

ok

moshthepitt:
> Also - Does the Express app actually render the view or is it just routing? If we're not doing server-side React rendering then it feels like it should be the later.

@cKellyDesign it needs to render the view so that the front-end/client can have access to the express session and be able to access the `/oauth/state/` express endpoint.  Otherwise the client/front-end will be treated like an un-authorized third party.

We aren't doing any server side rendering, but that is something we should consider - we are already set up for it now.

p-netm:
> this is great! Just one change request on the flow diagram: remove `Nothing Happens???` and add a red arrow pointing back to `Express renders react app login`

@cKellyDesign This is actually what happens, see #717 , should i still go ahead and apply this changes

p-netm:
> In the flow diagram, we are missing a path back to the login page when the user starts at `User types in a URL or refreshes tab`

@moshthepitt I think there i handled that, when a user types in a url, the next point of contact is the connected-private-route, the login is shown after a couple of steps . I think this part covers it, no?
typeInTab

cKellyDesign:
> > remove Nothing Happens??? and add a red arrow pointing back to Express renders react app login
>
> @cKellyDesign This is actually what happens, see #717 , should i still go ahead and apply this changes

@p-netm Yes please 🙏

p-netm:
> > > remove Nothing Happens??? and add a red arrow pointing back to Express renders react app login
> > 
> > 
> > @cKellyDesign This is actually what happens, see #717 , should i still go ahead and apply this changes
> 
> @p-netm Yes please pray

ok

PR #710, opened 2020-02-20
Persist selected plan on browser refresh

### LINKED ISSUE #697: RVL-878 - Selected plan should persist after refreshing 'Update Plan' page
When a user refreshes the Update Plan page for an existing plan (e.g. pressing control-R) the page reloads with no information. Note that the urls are the same , and that if the plan is saved, it overwrites the original plan

To reproduce:

1. Navigate to  Manage Plans
2. Click on a Plan
3. Click control-R when on the Update Plan page

As reported here https://smartregister.atlassian.net/browse/RVL-878

ciremusyoka:
Closes #697

### DIFF
--- a/src/containers/pages/InterventionPlan/UpdatePlan/index.tsx
+++ b/src/containers/pages/InterventionPlan/UpdatePlan/index.tsx
@@ -50,7 +50,8 @@ const UpdatePlan = (props: RouteComponentProps<RouteParams> & UpdatePlanProps) =
     try {
       setLoading(plan === null); // only set loading when there are no plans
       const planFromAPI = await apiService.read(planIdentifier);
-      fetchPlan(planFromAPI);
+      const currentPlan = Array.isArray(planFromAPI) ? planFromAPI[0] : planFromAPI;
+      fetchPlan(currentPlan);
     } catch (e) {
       // do something with the error?
     } finally {
--- a/src/containers/pages/InterventionPlan/UpdatePlan/index.tsx
+++ b/src/containers/pages/InterventionPlan/UpdatePlan/index.tsx
@@ -43,6 +43,7 @@ const UpdatePlan = (props: RouteComponentProps<RouteParams> & UpdatePlanProps) =
   if (!planIdentifier) {
     return null; /** we should make this into a better error page */
   }
+  const apiService = new service(OPENSRP_PLANS);
 
   /** async function to load the data */
   async function loadData() {
@@ -65,7 +66,6 @@ const UpdatePlan = (props: RouteComponentProps<RouteParams> & UpdatePlanProps) =
     return <Loading />;
   }
 
-  const apiService = new service(OPENSRP_PLANS);
   const pageTitle: string = plan ? `${UPDATE_PLAN}: ${plan.title}` : UPDATE_PLAN;
 
   const breadcrumbProps = {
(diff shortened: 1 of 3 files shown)


src/containers/pages/InterventionPlan/UpdatePlan/index.tsx
```diff
@@ -43,13 +43,15 @@ const UpdatePlan = (props: RouteComponentProps<RouteParams> & UpdatePlanProps) =
...
+  const apiService = new service(OPENSRP_PLANS);
 
   /** async function to load the data */
   async function loadData() {
     try {
       setLoading(plan === null); // only set loading when there are no plans
       const planFromAPI = await apiService.read(planIdentifier);
-      fetchPlan(planFromAPI);
+      const currentPlan = Array.isArray(planFromAPI) ? planFromAPI[0] : planFromAPI;
+      fetchPlan(currentPlan);
```
ciremusyoka: The data we are getting from API is a `list` but we were using it as an `object` thus when filtering using the `planIdentifier` it could not be found.

src/containers/pages/InterventionPlan/UpdatePlan/index.tsx
```diff
@@ -43,13 +43,15 @@ const UpdatePlan = (props: RouteComponentProps<RouteParams> & UpdatePlanProps) =
   if (!planIdentifier) {
     return null; /** we should make this into a better error page */
   }
+  const apiService = new service(OPENSRP_PLANS);
```
ciremusyoka: Also with the `apiService` defined before being called  was causing a problem which have not yet been able to understand fully.

moshthepitt: This probably came from the more aggressive linter that we are now using.

moshthepitt:
Created a related issue to get this fixed API-side https://github.com/OpenSRP/opensrp-server-web/issues/213

ciremusyoka:
> @ciremusyoka nice catch, well done. 
> 
> Please add a test for this. In my view, tests are especially important for bugfixes so that we avoid regressions.

@moshthepitt I think the test for this case should be covered on `Server side` to ensure consistency of data type being returned. To test the data type, i will be using a set of predefined mock data but if the `Server`  keeps on returning different data types our tests will not be able handle that. They will pass but still the bug will be there.

moshthepitt:
@ciremusyoka while what you are saying makes sense, from the point of view of our code we still need a const currentPlan = Array.isArray(planFromAPI) ? planFromAPI[0] : planFromAPI;
      fetchPlan(currentPlan);test to ensure that our fix is in place and works.  i.e.

```ts
const currentPlan = Array.isArray(planFromAPI) ? planFromAPI[0] : planFromAPI;
fetchPlan(currentPlan);
```

^^ this should be testable, right?

ciremusyoka:
> @ciremusyoka while what you are saying makes sense, from the point of view of our code we still need a const currentPlan = Array.isArray(planFromAPI) ? planFromAPI[0] : planFromAPI;
> fetchPlan(currentPlan);test to ensure that our fix is in place and works. i.e.
> 
> ```ts
> const currentPlan = Array.isArray(planFromAPI) ? planFromAPI[0] : planFromAPI;
> fetchPlan(currentPlan);
> ```
> 
> ^^ this should be testable, right?

@moshthepitt Yes it does.

moshthepitt:
@ciremusyoka cool, lets add the test so that we can merge and close this.

PR #713, opened 2020-02-20
Stop rendering NULL in focus Area field in ActiveFI

### LINKED ISSUE #550: Dont render NULL jurisdiction in FI Plan
NULL jurisdictions

p-netm:
fixes #550 

- cleans the focus area field to an empty string for 'NULL' values
- Removes util function `transformValues` -> Used to take an object and a list of propertyNames and would an optional list of invalid values, It would check if the values of the listed property names are included in the list of invalid values and replace them with an empty string if found.

It was primarily and only used to replaces jurisdiction values that took on a null shape in the focusInvestigation plan pages. With the resolution to remove plans with invalid jurisdiction_ids, this util function would no longer be used.


moshthepitt:
@p-netm what causes the NULL?

p-netm:
> @p-netm what causes the NULL?

Ill get back to you on this, not sure what the root of this is

src/containers/pages/FocusInvestigation/active/index.tsx
```diff
@@ -206,7 +206,13 @@ class ActiveFocusInvestigation extends React.Component<
...
               let thisItem = extractPlan(item);
               // transform values of this properties if they are null
-              const propertiesToTransform = ['village', 'canton', 'district', 'province'];
+              const propertiesToTransform = [
+                'village',
+                'canton',
+                'district',
+                'province',
+                'focusArea',
+              ];
```
cKellyDesign: These look like magic strings which should be moved to `constants` please

src/containers/pages/FocusInvestigation/active/index.tsx
```diff
@@ -269,7 +275,7 @@ class ActiveFocusInvestigation extends React.Component<
                     Cell: (cell: CellInfo) => {
                       return (
                         <div>
-                          {cell.original.focusArea.trim() && (
+                          {cell.value.trim() && (
```
cKellyDesign: 👍

src/containers/pages/FocusInvestigation/active/index.tsx
p-netm: @cKellyDesign This actually contains domain logic, they are not just strings, see [here](https://github.com/onaio/reveal-frontend/blob/24e4230c76784a5d0f4df9181e382f3e7b0009e2/src/helpers/utils.tsx#L334).

p-netm:
> @p-netm what causes the NULL?

[slack conversation](https://onaio.slack.com/archives/DJM16C4LU/p1572014185001300)

moshthepitt:
Thanks @p-netm - could you create an issue on the Canopy repo that includes the JSON from the slcak converation?  Assign it to me.

Meanwhile, if we get something whose jurisidiction_id is NULL/" NULL" then my view is that we should filter it out from what is shown on the table.

cc: @cKellyDesign @craigappl

cKellyDesign:
> if we get something whose jurisidiction_id is NULL/" NULL" then my view is that we should filter it out from what is shown on the table.

Agreed kabisa, I think `null` values should be filtered up as up-stream as possible. Ideally there are no `null` values in the slices, also good would be filtering out `null` values in the actual superset requests, and as a last resort we can filter `null` values in the client.

src/containers/pages/FocusInvestigation/active/index.tsx
```diff
@@ -203,11 +207,9 @@ class ActiveFocusInvestigation extends React.Component<
...
-            const thePlans = plansArray.map((item: Plan) => {
-              let thisItem = extractPlan(item);
-              // transform values of this properties if they are null
-              const propertiesToTransform = ['village', 'canton', 'district', 'province'];
-              thisItem = transformValues(thisItem, propertiesToTransform);
+            const jurisdictionValidPlans = removeNullJurisdictionPlans(plansArray);
+            const thePlans = jurisdictionValidPlans.map((item: Plan) => {
+              const thisItem = extractPlan(item);
               return thisItem;
             });
```
cKellyDesign: Can we simplify this map function to one line?

src/containers/pages/FocusInvestigation/single/index.tsx
```diff
@@ -232,10 +222,7 @@ class SingleFI extends React.Component<RouteComponentProps<RouteParams> & Single
       [currentReactivePlansArray, currentRoutinePlansArray].forEach((plansArray: Plan[]) => {
         if (plansArray && plansArray.length) {
           const thePlans = plansArray.map((item: Plan) => {
-            let thisItem = extractPlan(item);
-            // transform values of this properties if they are null
-            const columnsToTransform = ['village', 'canton', 'district', 'province'];
-            thisItem = transformValues(thisItem, columnsToTransform);
+            const thisItem = extractPlan(item);
             return thisItem;
```
cKellyDesign: Same here, since the extra logic was moved out of this block, it seems like this could be a one-liner.

src/helpers/utils.tsx
```diff
@@ -426,31 +426,19 @@ export function getColor(taskObject: InitialTask): Color {
...
-  propertyNames.forEach(propertyName => {
-    let preOldValue = (thisObj as any)[propertyName];
-    // preprocess received old value: trim and ensure lower case
-    if (typeof preOldValue === 'string') {
-      preOldValue = preOldValue.toLowerCase().trim();
-    }
-    if ((thisObj as any)[propertyName] !== undefined && oldValue.indexOf(preOldValue) > -1) {
-      (thisObj as any)[propertyName] = newValue;
-    }
+export const removeNullJurisdictionPlans = (plansArray: Plan[]): Plan[] => {
```
cKellyDesign: 👍

p-netm:
Thanks @cKellyDesign , please re-review

src/containers/pages/FocusInvestigation/active/tests/__snapshots__/index.test.tsx.snap
```diff
@@ -479,11 +479,11 @@ Object {
     "canton": "Canton Tha Luang",
     "caseClassification": null,
     "caseNotificationDate": "2019-06-18",
-    "district": "",
+    "district": null,
     "focusArea": "TLv1_02",
     "plan_status": "complete",
     "plan_title": "A1-Tha Luang Village 1 Focus 01",
-    "province": "",
+    "province": null,
```
moshthepitt: @p-netm what is causing these changes here?

src/containers/pages/FocusInvestigation/active/index.tsx
```diff
@@ -269,7 +270,7 @@ class ActiveFocusInvestigation extends React.Component<
                     Cell: (cell: CellInfo) => {
                       return (
                         <div>
-                          {cell.original.focusArea.trim() && (
+                          {cell.value.trim() && (
```
moshthepitt: @p-netm why this change?

src/containers/pages/FocusInvestigation/single/tests/index.test.tsx
```diff
@@ -49,15 +49,29 @@ describe('containers/pages/SingleFI', () => {
         path: `${FI_SINGLE_URL}/:id`,
         url: `${FI_SINGLE_URL}/13`,
       },
-      planById: fixtures.plan1,
+      planById: null,
```
moshthepitt: @p-netm why this change?

src/containers/pages/FocusInvestigation/single/tests/index.test.tsx
```diff
@@ -101,43 +115,6 @@ describe('containers/pages/SingleFI', () => {
     wrapper.unmount();
   });
 
-  it('renders SingleFI correctly for null plan jurisdction id', () => {
```
moshthepitt: @p-netm was it necessary to completely remove this test?

src/helpers/tests/utils.test.tsx
```diff
@@ -231,29 +232,6 @@ describe('helpers/utils', () => {
     expect(color).toEqual(YELLOW);
   });
 
-  it('updates property names correctly', () => {
```
moshthepitt: Why was it necessary to completely remove this?

src/helpers/utils.tsx
```diff
@@ -426,31 +426,19 @@ export function getColor(taskObject: InitialTask): Color {
...
- * {string[]} propertyNames - list of property names in obj whose values will change
- * {any} newValue - the value to be assigned the property names
- * {any[]} old_value - replace property name if old_value is among specified here
+
+/** filters out plans whose jurisdiction id is null
+ * @param {Plan []} plansArray - a list of pre-filtered plans
+ *
+ * @return {Plan []} - A list of filtered plans
  */
-export function transformValues<T>(
```
moshthepitt: @p-netm could you add a little note in this PR for the reason why we removed this?

src/store/ducks/tests/fixtures.ts
```diff
@@ -288,8 +288,8 @@ export const planRecordsById = {
     plan_date: '2019-05-24',
     plan_effective_period_end: '2019-09-30',
     plan_effective_period_start: '2019-05-27',
-    plan_fi_reason: '',
-    plan_fi_status: '',
+    plan_fi_reason: ' NULL',
```
moshthepitt: Why are we doing this? 

- is it possible to get plans with the Fi reason and status as `''`?
- is it possible to get plans with the Fi reason and status as `' NULL'`?

src/store/ducks/tests/fixtures.ts
```diff
@@ -317,8 +317,8 @@ export const sortedPlanRecordArray = [
...
```
moshthepitt: Why are we doing this? 

- is it possible to get plans with the Fi reason and status as `''`?
- is it possible to get plans with the Fi reason and status as `' NULL'`?

moshthepitt:
> Coverage decreased (-0.1%) to 66.598% when pulling 38a8820 on 550-nullJurisdictions into 6173b74 on master.

@p-netm what caused the coverage drop, and how can you fix it?

src/containers/pages/FocusInvestigation/active/tests/__snapshots__/index.test.tsx.snap
p-netm: I removed the util function that was cleaning up the jurisdiction data. `TransformValues`. it removed null values and replaced them with an empty string.I believe Its functionality is no longer required since the plans it was cleaning wont be used for display anymore.

moshthepitt: @p-netm thanks.  In the future please do remember to have such "why" information included in the commit message.  This one would have gone here https://github.com/onaio/reveal-frontend/pull/713/commits/8b831639390fa08973af83c1c737e36beff9371a

moshthepitt: https://chris.beams.io/posts/git-commit/

src/containers/pages/FocusInvestigation/active/index.tsx
p-netm: An old change from when i had tried replacing ' NULL' jurisdiction ids with an empty string, it would stop displaying the the plan title after changing the focusArea value to an empty string. I can revert this.

src/containers/pages/FocusInvestigation/active/tests/__snapshots__/index.test.tsx.snap
p-netm: ok

src/containers/pages/FocusInvestigation/single/tests/index.test.tsx
p-netm: To increase coverage. for this test, we would also be testing that the loading component was shown. Later on in the same testCase i replace the null value with a valid planId.

src/containers/pages/FocusInvestigation/single/tests/index.test.tsx
p-netm: I believe so, it tested a case that we would not be supported any longer, i.e plans that had jurisdiction_ids with ' NULL' value.

src/helpers/utils.tsx
p-netm: ok

src/store/ducks/tests/fixtures.ts
p-netm: seems like we need to cleanup the fixtures data. @moshthepitt i do not know if there could be cases where fi_plan_reason: and plan_fi_status is ' NULL'. If this is not a possible case , I could remove all plans in fixtures that has the named fields as ' NULL'.

PR #715, opened 2020-02-20
Fixing login page showing after login

### LINKED ISSUE #687: Login page is still accessible to an authenticated user
__Currently__:
An authenticated user could route to the login page successfully

__Ideally__:
A user should only be able to route to the login page if they are not logged in.

__scenarios__:
- an already logged in user; types the 'login' url in the address bar
- a user logs in from tab `a`; logs out from tab  `a`, thus the page on tab `a` is the login page. opens a new tab `b`, and successfully logs in, comes back to tab `a`. tab `a` is still on the login page

_Related to this_:
if a user successfully logs in from tab `a`, opens a new tab `b` while still logged in, switches back to tab `a` and logs out, if they circle back to tab `b` they will still be logged in

ciremusyoka:
Closes #687 
Added a check for session validity before accessing `login component`.


ciremusyoka:
> @ciremusyoka not sure how I feel about this. Currently, the login page is handled by the React app. Do you think it is possible to have a React-only fix (i.e. not in express)? (This is not a request, it is a question).
> 
> cc: @p-netm for his thoughts on this

@moshthepitt Have looked around but am not getting a better fix for this but have created an issue for moving the fix. #718

PR #728, opened 2020-03-03
Refactor focus investigation map

### LINKED ISSUE #638: Refactor FI maps
The FI map page is older code and may not be as elegant as, for instance, the IRS map pages.

We need to:

- [ ] Rewrite to be better

- [ ] Add more tests

Context: https://onaio.slack.com/archives/C0482AJP6/p1579863718032900

kelvin-muchiri:
Closes #638 

Refactored to use function based component. Added more tests

### DIFF
--- a/src/containers/pages/FocusInvestigation/map/active/helpers/utils.tsx
+++ b/src/containers/pages/FocusInvestigation/map/active/helpers/utils.tsx
@@ -13,7 +13,7 @@ import { CASE_CLASSIFICATION_LABEL, END_DATE, START_DATE } from '../../../../../
 import { JURISDICTION_ID, PLAN_ID } from '../../../../../../constants';
 import { ROUTINE } from '../../../../../../constants';
 import { displayError } from '../../../../../../helpers/errors';
-import { popupHandler } from '../../../../../../helpers/handlers';
+import { PopHandler, popupHandler } from '../../../../../../helpers/handlers';
 import supersetFetch from '../../../../../../services/superset';
 import { fetchGoals, Goal } from '../../../../../../store/ducks/goals';
 import { fetchJurisdictions, Jurisdiction } from '../../../../../../store/ducks/jurisdictions';
@@ -117,7 +117,7 @@ export const fetchData = async (
  * Build mapbox component event handlers
  * @param method  Event handler
  */
-export const buildHandlers = (method: any = popupHandler) => {
+export const buildHandlers = (method: PopHandler = popupHandler) => {
   return [
     {
       method,
--- a/src/containers/pages/FocusInvestigation/map/active/helpers/utils.tsx
+++ b/src/containers/pages/FocusInvestigation/map/active/helpers/utils.tsx
@@ -1,4 +1,5 @@
 import superset from '@onaio/superset-connector';
+import * as React from 'react';
 import {
   SUPERSET_GOALS_SLICE,
   SUPERSET_JURISDICTIONS_SLICE,
@@ -6,27 +7,28 @@ import {
   SUPERSET_PLANS_SLICE,
   SUPERSET_STRUCTURES_SLICE,
   SUPERSET_TASKS_SLICE,
-} from '../../../../../configs/env';
-import { AN_ERROR_OCCURED } from '../../../../../configs/lang';
-import { JURISDICTION_ID, PLAN_ID } from '../../../../../constants';
-import { displayError } from '../../../../../helpers/errors';
-import { popupHandler } from '../../../../../helpers/handlers';
-import supersetFetch from '../../../../../services/superset';
-import { fetchGoals, Goal } from '../../../../../store/ducks/goals';
-import { fetchJurisdictions, Jurisdiction } from '../../../../../store/ducks/jurisdictions';
-import { fetchPlans, Plan } from '../../../../../store/ducks/plans';
-import { setStructures, Structure } from '../../../../../store/ducks/structures';
-import { fetchTasks, Task } from '../../../../../store/ducks/tasks';
-
+} from '../../../../../../configs/env';
+import { AN_ERROR_OCCURED } from '../../../../../../configs/lang';
+import { CASE_CLASSIFICATION_LABEL, END_DATE, START_DATE } from '../../../../../../configs/lang';
+import { JURISDICTION_ID, PLAN_ID } from '../../../../../../constants';
+import { ROUTINE } from '../../../../../../constants';
+import { displayError } from '../../../../../../helpers/errors';
+import { popupHandler } from '../../../../../../helpers/handlers';
+import supersetFetch from '../../../../../../services/superset';
+import { fetchGoals, Goal } from '../../../../../../store/ducks/goals';
+import { fetchJurisdictions, Jurisdiction } from '../../../../../../store/ducks/jurisdictions';
+import { fetchPlans, Plan } from '../../../../../../store/ducks/plans';
+import { setStructures, Structure } from '../../../../../../store/ducks/structures';
+import { fetchTasks, Task } from '../../../../../../store/ducks/tasks';
 /**
  * Fetch data for the plan
- * @param fetchGoalsActionCreator // fetch goals action creator
- * @param fetchJurisdictionsActionCreator // fetch jurisdication action creator
- * @param fetchPlansActionCreator // fetch plans action creator
- * @param fetchStructuresActionCreator // fetch structures action creator
- * @param fetchTasksActionCreator // fetch tasks action creator
- * @param plan // plan object
- * @param supersetService // fetch superset promise
+ * @param fetchGoalsActionCreator Fetch goals action creator
+ * @param fetchJurisdictionsActionCreator Fetch jurisdication action creator
+ * @param fetchPlansActionCreator Fetch plans action creator
+ * @param fetchStructuresActionCreator Fetch structures action creator
+ * @param fetchTasksActionCreator Fetch tasks action creator
+ * @param plan Plan object
+ * @param supersetService Fetch superset promise
  */
 export const fetchData = async (
   fetchGoalsActionCreator: typeof fetchGoals,
@@ -113,7 +115,7 @@ export const fetchData = async (
 
 /**
  * Build mapbox component event handlers
- * @param method // Event handler
+ * @param method  Event handler
  */
 export const buildHandlers = (method: any = popupHandler) => {
   return [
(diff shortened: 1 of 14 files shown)


src/containers/pages/FocusInvestigation/map/active/index.tsx
```diff
@@ -153,319 +152,344 @@ export const defaultMapSingleFIProps: MapSingleFIProps = {
...
 
-/** Map View for Single Active Focus Investigation */
-class SingleActiveFIMap extends React.Component<
-  RouteComponentProps<RouteParams> & MapSingleFIProps,
-  FlexObject
-> {
-  public static defaultProps = defaultMapSingleFIProps;
-  constructor(props: RouteComponentProps<RouteParams> & MapSingleFIProps) {
-    super(props);
+const fetchData = async (
```
cKellyDesign: Should we export this function to make it testable?

src/containers/pages/FocusInvestigation/map/active/index.tsx
```diff
@@ -153,319 +152,344 @@ export const defaultMapSingleFIProps: MapSingleFIProps = {
...
+      .then((result4: Task[]) => {
+        fetchTasksActionCreator(result4);
+      })
+      .catch(error => {
+        throw error;
+      });
+  }
+};
+
+const buildHandlers = () => {
```
cKellyDesign: Ooh yeah let's export all of these for testability! Let's also include doc strings, and can we consider moving some of these into  an adjacent `helpers` file?

src/containers/pages/FocusInvestigation/map/active/index.tsx
```diff
@@ -153,319 +152,344 @@ export const defaultMapSingleFIProps: MapSingleFIProps = {
...
+      });
+    await supersetService(SUPERSET_TASKS_SLICE, supersetParams)
+      .then((result4: Task[]) => {
+        fetchTasksActionCreator(result4);
+      })
+      .catch(error => {
+        throw error;
+      });
+  }
+};
```
moshthepitt: 1. Please consider using [displayError](https://github.com/onaio/reveal-frontend/blob/acdc40f222c9a3bf281c5f2e90ffe180ba34f853/src/helpers/errors.ts#L9) instead of throwing.
2. Please also consider that the result inside the `then` might be false-y, in which case an HTTP error occurred (think 4xx or 5xx errors).  In this case, we also need to use `displayError`.  Look at [how we did it here](https://github.com/onaio/reveal-frontend/pull/692/files#diff-79ae51a7daeac08c863a2be206d150cdR58-R63) - and please suggest any improvements to that process in this PR.

kelvin-muchiri: > 2\. Please also consider that the result inside the `then` might be false-y, in which case an HTTP error occurred (think 4xx or 5xx errors).  In this case, we also need to use `displayError`.  Look at [how we did it here](https://github.com/onaio/reveal-frontend/pull/692/files#diff-79ae51a7daeac08c863a2be206d150cdR58-R63) - and please suggest any improvements to that process in this PR.

I also noticed that we display the raw error message. I think the error returned bycatch is useful for  a developer and can be logged to sentry but not useful to the user. I think display error should display generic error messages e.g **An error occurred while processing your request** etc

moshthepitt: @kelvin-muchiri you are absolutely right.  Its my bad - I had meant to create an issue to enhance displayError so that it logs the Error to somewhere, and then show to the user a nice friendly error.

moshthepitt: I have created the issue here https://github.com/onaio/reveal-frontend/issues/729

kelvin-muchiri: > 1. Please consider using [displayError](https://github.com/onaio/reveal-frontend/blob/acdc40f222c9a3bf281c5f2e90ffe180ba34f853/src/helpers/errors.ts#L9) instead of throwing.

The error was being displayed by the caller https://github.com/onaio/reveal-frontend/blob/561909bce59fddf4dd47336987c6819df92d4dec/src/containers/pages/FocusInvestigation/map/active/index.tsx#L326. Which one is best. Displaying the error in the caller or the callee?

moshthepitt: @kelvin-muchiri not sure what you mean, could you restate that?

kelvin-muchiri: > @kelvin-muchiri not sure what you mean, could you restate that?

I was  catching the errors when I call the method `fetchData`

moshthepitt: @kelvin-muchiri cool, thanks.  In that case I would leave it as it is.  Only thing left to do is deal with cases where within `fetchData` `supersetService` returns something false-y

src/containers/pages/FocusInvestigation/map/active/index.tsx
```diff
@@ -153,319 +143,272 @@ export const defaultMapSingleFIProps: MapSingleFIProps = {
...
+    );
   }
 
-  public componentDidMount() {
-    if (!this.props.plan) {
+  return detailViewPlanInvestigationContainer;
+};
+
+/** Map View for Single Active Focus Investigation */
+const SingleActiveFIMap = (props: MapSingleFIProps & RouteComponentProps<RouteParams>) => {
```
cKellyDesign: @kelvin-muchiri can we export all of these from this file?

src/containers/pages/FocusInvestigation/map/active/utils.ts
```diff
@@ -0,0 +1,126 @@
...
+ * Fetch data for the plan
+ * @param fetchGoalsActionCreator // fetch goals action creator
+ * @param fetchJurisdictionsActionCreator // fetch jurisdication action creator
+ * @param fetchPlansActionCreator // fetch plans action creator
+ * @param fetchStructuresActionCreator // fetch structures action creator
+ * @param fetchTasksActionCreator // fetch tasks action creator
+ * @param plan // plan object
+ * @param supersetService // fetch superset promise
+ */
+export const fetchData = async (
```
moshthepitt: What are your thoughts about using displayError here as opposed to throwing the errors and letting them be handled by whoever calls this function?

src/containers/pages/FocusInvestigation/map/active/index.tsx
```diff
@@ -153,319 +143,272 @@ export const defaultMapSingleFIProps: MapSingleFIProps = {
...
-  FlexObject
-> {
-  public static defaultProps = defaultMapSingleFIProps;
-  constructor(props: RouteComponentProps<RouteParams> & MapSingleFIProps) {
-    super(props);
+interface StatusBadgeProps {
+  plan: Plan;
+}
+
+const StatusBadge = (props: StatusBadgeProps) => {
```
moshthepitt: - What are your thoughts regarding moving this to the components directory (or a helper module)?
- Please add a docstring here

src/containers/pages/FocusInvestigation/map/active/index.tsx
```diff
@@ -153,319 +143,272 @@ export const defaultMapSingleFIProps: MapSingleFIProps = {
...
+      {plan && plan.plan_status}
+    </Badge>
+  );
+};
+
+interface MarkCompleteLinkProps {
+  plan: Plan;
+}
+
+const MarkCompleteLink = (props: MarkCompleteLinkProps) => {
```
moshthepitt: - What are your thoughts regarding moving this to the components directory (or a helper module)?
- Please add a docstring here

src/containers/pages/FocusInvestigation/map/active/index.tsx
```diff
@@ -153,319 +143,272 @@ export const defaultMapSingleFIProps: MapSingleFIProps = {
...
+      <Link className="btn btn-primary" to={`${PLAN_COMPLETION_URL}/${plan.id}`} color="primary">
+        {MARK_AS_COMPLETE}
+      </Link>
+    );
+  }
+
+  return null;
+};
+
+const getDetailViewPlanInvestigationContainer = (plan: Plan): React.ReactElement[] => {
```
moshthepitt: - What are your thoughts regarding moving this to the components directory (or a helper module)?
- Please add a docstring here

src/containers/pages/FocusInvestigation/map/active/index.tsx
```diff
@@ -153,319 +143,272 @@ export const defaultMapSingleFIProps: MapSingleFIProps = {
...
-       * the plans
+       * Fetch plans incase the plan is not available e.g when page is refreshed
        */
-      const { supersetService, fetchPlansActionCreator } = this.props;
+      const { supersetService, fetchPlansActionCreator } = props;
       const supersetParams = superset.getFormData(2000, [
         { comparator: InterventionType.FI, operator: '==', subject: PLAN_INTERVENTION_TYPE },
       ]);
       supersetService(SUPERSET_PLANS_SLICE, supersetParams)
         .then((result: Plan[]) => fetchPlansActionCreator(result))
```
moshthepitt: Let's please handle the case where result might be false-y

src/containers/pages/FocusInvestigation/map/active/index.tsx
moshthepitt: I am curious about your thoughts on this component as regards the [Open/Closed principle](https://stackify.com/solid-design-open-closed-principle/).  I am specifically wondering about its usage of sub components, and functions such as `fetchData` and whether it should take them in as props.  What do you think? (This is not a directive but a question)

kelvin-muchiri: @moshthepitt Mmmh...What do you think might change in future that would help if we passed it as a prop? Also, does it mean we extend the functionality of `fetchData` without affecting the render method of the component?

moshthepitt: @kelvin-muchiri that's a good question.  I'll leave it to you to decide if we need to do anythin regarding that.

kelvin-muchiri: @moshthepitt I haven't thought about it yet. Maybe we can proceed at it is for now.

moshthepitt: Cool @kelvin-muchiri

kelvin-muchiri: @moshthepitt Is the coveralls build not completing because the base branch is not master?

moshthepitt: Yeah we need @p-netm to update tiny-backend with master, and then you can do that for this branch as well.  But I am fine merging this before then.

src/containers/pages/FocusInvestigation/map/active/helpers/utils.tsx
```diff
@@ -0,0 +1,169 @@
...
+        throw error;
+      });
+  }
+};
+
+/**
+ * Build mapbox component event handlers
+ * @param method  Event handler
+ */
+export const buildHandlers = (method: any = popupHandler) => {
```
cKellyDesign: Can we be more specific than `any` with this typing?

src/containers/pages/FocusInvestigation/map/active/index.tsx
```diff
@@ -154,318 +142,204 @@ export const defaultMapSingleFIProps: MapSingleFIProps = {
...
-        </Link>
-      ) : null;
-    /** Array that holds keys of a plan object
-     * Will be used to check plan_effective_period_end or plan_effective_period_start to build the detailview
+    setCurrentGoalActionCreator(match.params.goalId ? match.params.goalId : null);
+    /**
+     * Only re-run effect if props.match.params.goalId changes
      */
-    const planKeysArray: string[] = Object.keys(plan);
+  }, [props.match.params.goalId]);
```
cKellyDesign: This feels like it's skipping the `mapStateToProps` step, would it make sense to reference the `currentGoal` being passed in as props? Would it make more sense to simply pass a `goalId` through the props?

@kelvin-muchiri @moshthepitt @p-netm - what are your thoughts?

PR #731, opened 2020-03-05
Update indicator color and thresholds for IRS Namibia

### LINKED ISSUE #646: RVL-734 Change Colour Gradient Range for Dashboards

kelvin-muchiri:
Closes #646

### DIFF
--- a/src/configs/settings.ts
+++ b/src/configs/settings.ts
@@ -67,7 +67,6 @@ import {
   IRS_GOAL_MEASURE,
   IRS_GREEN_THRESHOLD,
   IRS_GREY_THRESHOLD,
-  IRS_LIGHT_GREEN_THRESHOLD,
   IRS_RED_THRESHOLD,
   IRS_YELLOW_THRESHOLD,
   LARVAL_DIPPING_ACTIVITY,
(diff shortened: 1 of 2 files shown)


src/configs/settings.ts
```diff
@@ -882,21 +881,21 @@ export const indicatorThresholdsIRSNamibia: IndicatorThresholds = {
...
-    value: 0.85,
+  GREY_THRESHOLD: {
+    color: '#dddddd',
+    name: IRS_GREY_THRESHOLD,
+    value: 0.2,
   },
   RED_THRESHOLD: {
     color: '#FF4136',
     name: IRS_RED_THRESHOLD,
     orEquals: true,
```
moshthepitt: Should this still have orEquals as true?

kelvin-muchiri: @moshthepitt Not sure but it is still the same as the default thresholds above it. Also, doesn't affect the component `src/components/formatting/IRSIndicatorLegend/index.tsx`. The `orEquals` for the default threshold affects component `src/helpers/ProgressBar.tsx`

Issue #717, opened 2020-02-21
Express server should redirect user back to login if they do not authorize


p-netm:
**Current Behaviour:**

If a user denies the client(the express server) permission to access their resources on the opensrp authorization server. Then the user is stuck on a page with the error `{"message":"The resource owner or authorization server denied the request.","status":"error"}` 

**Expected Behaviour:**
the user should be redirected to the login page

**Probable solution**
I guess this would be an enhancement, the express server's callback url should be able to handle this kind of response and redirect accordingly

p-netm:
closed in #734

PR #734, opened 2020-03-06
722 test server

### LINKED ISSUE #688: Redirect User to a route after logging them in
__Current behaviour__:
if a user session does not exist and say they try visiting `/teams`. the app directs them to the authentication page, after which, if they are successfully authenticated they are then redirected to the home page

__expected behavior__:
if a user session does not exist and say they try visiting `/teams`. the app directs them to the authentication page, after which, if they are successfully authenticated they are then redirected to the page that they had initially tried to access i.e. `/teams`

### LINKED ISSUE #722: Test the server
currently server test coverage stands at 0% => we need to add tests for the same.

p-netm:
Adds tests for the server closes #722 
Redirects user to the path they were trying to access after authorizing them closes #688 
Redirect users to login url if they refuse authorizing the web app on the authorization server #717


server/.env.sample
```diff
@@ -17,4 +17,5 @@ EXPRESS_SESSION_FILESTORE_PATH=/tmp/express-sessions
 EXPRESS_PRELOADED_STATE_FILE=/tmp/revealState.json
 
 EXPRESS_SESSION_LOGIN_URL=/login
-FRONTEND_OPENSRP_CALLBACK_URL=http://localhost:3000/fe/oauth/callback/opensrp
\ No newline at end of file
+FRONTEND_OPENSRP_CALLBACK_URL=http://localhost:3000/fe/oauth/callback/opensrp
+FRONTEND_LOGIN_URL=/fe/login
```
moshthepitt: Can we append `EXPRESS_` to these please?

server/src/configs/envs.ts
```diff
@@ -62,3 +62,6 @@ export type EXPRESS_SESSION_PATH = typeof EXPRESS_SESSION_PATH;
 export const EXPRESS_REACT_BUILD_PATH =
   process.env.EXPRESS_REACT_BUILD_PATH || path.resolve(path.resolve(), '../build');
 export type EXPRESS_REACT_BUILD_PATH = typeof EXPRESS_REACT_BUILD_PATH;
+
+export const FRONTEND_LOGIN_URL = process.env.FRONTEND_LOGIN_URL || '/fe/login';
+export type FRONTEND_LOGIN_URL = typeof FRONTEND_LOGIN_URL;
```
moshthepitt: Lets append `EXPRESS_` to these please?

server/src/index.ts
```diff
@@ -133,6 +143,10 @@ const oauthCallback = (req: express.Request, res: express.Response, next: expres
             req.session.cookie.maxAge = expireAfterMs;
             // you have to save the session manually for POST requests like this one
             req.session.save(() => void 0);
+            if (nextPath) {
+              nextPath = undefined;
```
moshthepitt: why do we do this?

server/.env.sample
p-netm: yeap

server/src/configs/__mocks__/envs.ts
```diff
@@ -0,0 +1,51 @@
...
+export type EXPRESS_SESSION_FILESTORE_PATH = typeof EXPRESS_SESSION_FILESTORE_PATH;
+
+export const EXPRESS_PRELOADED_STATE_FILE = '/tmp/revealState.json';
+export type EXPRESS_PRELOADED_STATE_FILE = typeof EXPRESS_PRELOADED_STATE_FILE;
+
+export const EXPRESS_SESSION_LOGIN_URL = '/login';
+export type EXPRESS_SESSION_LOGIN_URL = typeof EXPRESS_SESSION_LOGIN_URL;
+
+export const FRONTEND_OPENSRP_CALLBACK_URL = '/fe/oauth/callback/opensrp';
+export type FRONTEND_OPENSRP_CALLBACK_URL = typeof FRONTEND_OPENSRP_CALLBACK_URL;
```
moshthepitt: Can we also prepend `EXPRESS_` here?

server/src/configs/envs.ts
moshthepitt: Don't forget about this

moshthepitt:
Please update this branch so that the coveralls tests are run

server/src/index.ts
p-netm: @moshthepitt I think this is variable represents a store that holds one-time-use data, therefore needing to reset it after where its used. Otherwise if i failed to reset the value, then all other logins (via the auth server) would redirect to the url in nextPath irregardless of whether the user first visited that url or not, before logging in.

moshthepitt: @p-netm cool, thanks.  Could we leave a comment in the code that says this please?

server/src/index.ts
```diff
@@ -104,6 +114,7 @@ const oauthLogin = (_: express.Request, res: express.Response) => {
 
 const oauthCallback = (req: express.Request, res: express.Response, next: express.NextFunction) => {
   const provider = opensrpAuth;
+  console.log("Original Url", req.originalUrl);
```
moshthepitt: Please remove this one when done.

moshthepitt: @p-netm curious why the lint test did not flag this one.  Do we need to create an issue to investigate this?

p-netm: Will create an issue.

server/src/index.ts
p-netm: yeah. on it

PR #743, opened 2020-03-13
Run frontend without backend

### LINKED ISSUE #735: Make it possible to run the frontend without a backend
Let us make it possible to run the frontend code without the requirement of having a backend server.

ciremusyoka:
Closes #735 
Makes it possible to run frontend without backend based on an environment variable `REACT_APP_BACKEND_ACTIVE`. When set to `true` backend configs are loaded and when set to `false` frontend configs are loaded.


src/App/App.tsx
```diff
@@ -77,6 +95,12 @@ toast.configure({
 });
 initGoogleAnalytics();
 
+const APP_CALLBACK_URL = BACKEND_ACTIVE ? OAUTH_CALLBACK_URL : REACT_LOGIN_URL;
+const { IMPLICIT, AUTHORIZATION_CODE } = AuthorizationGrantType;
+const AuthGrandType = BACKEND_ACTIVE ? AUTHORIZATION_CODE : IMPLICIT;
```
moshthepitt: This should say `AuthGrantType` - no?

src/App/App.tsx
```diff
@@ -77,6 +95,12 @@ toast.configure({
...
+const APP_LOGIN_URL = BACKEND_ACTIVE ? LOGIN_URL : REACT_LOGIN_URL;
+const APP_CALLBACK_PATH = BACKEND_ACTIVE ? OAUTH_CALLBACK_PATH : REACT_CALLBACK_PATH;
+
```
moshthepitt: Can we name these variables in a way that it is easier to differentiate between the REACT_x variables Vs the BACKEND_x variables?  For instance I think `OAUTH_CALLBACK_PATH` is not clear that it relates to the backend while `REACT_CALLBACK_PATH` is very clear that it relates to React/the frontend.

src/App/App.tsx
```diff
@@ -329,31 +353,56 @@ class App extends Component {
...
                   path={LOGOUT_URL}
                   // tslint:disable-next-line: jsx-no-lambda
                   component={() => {
-                    store.dispatch(logOutUser());
-                    window.location.href = EXPRESS_OAUTH_LOGOUT_URL;
-                    return <></>;
+                    if (BACKEND_ACTIVE) {
+                      store.dispatch(logOutUser());
+                      window.location.href = EXPRESS_OAUTH_LOGOUT_URL;
+                      return <></>;
```
moshthepitt: How would you rate this as opposed to returning `null` or `<div></div>`?

moshthepitt:
@ciremusyoka How do we test this?

ciremusyoka:
> @ciremusyoka How do we test this?

Maybe we find a way to change the status of the env variable `REACT_APP_BACKEND_ACTIVE`.

moshthepitt:
> Maybe we find a way to change the status of the env variable REACT_APP_BACKEND_ACTIVE in the tests.

Could you look into this @ciremusyoka  please?

Naima-Bashir:
Tested this with the help of @p-netm . Works as expected.

Issue #527, opened 2019-10-15
Use Proxies to define default values


cKellyDesign:
[Brought up in this PR](https://github.com/onaio/reveal-frontend/pull/503#discussion_r333813492), we should look into defining default object values (like reducer state and component props) using [Proxies](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy).

moshthepitt:
Nice one

PR #744, opened 2020-03-13
Use Proxies to define default values

kahummer:
fix: #527

### DIFF
--- a/src/store/ducks/opensrp/assignments/index.ts
+++ b/src/store/ducks/opensrp/assignments/index.ts
@@ -94,7 +94,15 @@ export const removeAssignmentsAction: RemoveAssignmentsAction = {
 export const fetchAssignments = (assignmentsList: Assignment[]): FetchAssignmentsAction => {
   // const assignmentsByPlanId: AssignmentsByPlanId = {};
   const defaultArrayAssignmentHandler: ProxyHandler<object> = {
-    get: (target: FlexObject, plan: string) => {
+    /**
+     * The methods inside the handler object are called traps for our case we use a get trap
+     * @param {AssignmentsByPlanId} target
+     * @param {string} plan planId the assignmentsByPlanId will be updated by empty arrays keyed by the planId
+     */
+    get: (target: AssignmentsByPlanId, plan: string) => {
+      /**
+       * Check if the planId already exists then return the target else add the key with an empty array value
+       */
       if (plan in target) {
         return target;
       } else {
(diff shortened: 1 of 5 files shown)


moshthepitt:
@kahummer this being new, could you document your thoughts about this, and how you propose we use it going forward?

Also, quick question: are we going to use this proxies in just the part of the code in this PR, or elsewhere as well?

kahummer:
@moshthepitt For sure the proxy concept is new to me too :baby: . The current implementation uses a proxy to implement default value for an object. Here is a sample of default value with **get trap** https://javascript.info/proxy#default-value-with-get-trap.  I think we can use proxies in similar scenarios and even in other use cases where proxies may be essential i.e validation, readonly properties among others. 

> Also, quick question: are we going to use this proxies in just the part of the code in this PR, or elsewhere as well?
> 
I created an issue to start a discussion on areas we might want to use this https://github.com/onaio/reveal-frontend/issues/748

src/store/ducks/opensrp/assignments/index.ts
```diff
@@ -91,15 +92,41 @@ export const removeAssignmentsAction: RemoveAssignmentsAction = {
...
+     */
+    get: (target: AssignmentsByPlanId, plan: string) => {
+      /**
+       * Check if the planId already exists then return the target else add the key with an empty array value
+       */
+      if (plan in target) {
+        return target;
+      } else {
+        target[plan] = [];
+        return target;
```
cKellyDesign: I think the first `return` negates the need for the `else` 😄

src/store/ducks/opensrp/assignments/index.ts
```diff
@@ -91,15 +92,41 @@ export const removeAssignmentsAction: RemoveAssignmentsAction = {
...
+      }
+    },
+  };
 
+  let assignmentsByPlanId: AssignmentsByPlanId = {};
+  /** Research on a solution  for Typescript can't infer types when using Proxy */
```
cKellyDesign: is this a todo item? If so please add `todo - ` at the start of the comment 🙏

moshthepitt:
Does need any extra testing?

src/store/ducks/opensrp/assignments/index.ts
kahummer: @cKellyDesign I think the return will only be called when plan exist in the target otherwise we add the plan as a key and assign an empty array.

cKellyDesign: yup totally, my nit pick was about syntax:
```ts
if (plan in target) {
  return target;
} // no else required

target[plan] = [];
return target;
```

kahummer:
@moshthepitt  Looking to add a test that asserts  what the trap returns

moshthepitt:
Nice

PR #745, opened 2020-03-13
Fix the application auto-logging out bug

p-netm:
**BUG** (There isn't an ticket for this.)
Currently visible on [staging](https://web.reveal-stage.smartregister.org/):
- login
- logout
- login
- Observe that you are not redirected to home, you are actually logget out.

**why**
the logout url is sometime appended to the express server login url as a next path that the express server should redirect to after logging in, therefor meaning that on the second login you actually got logged in and automatically logged out.

This Pr is intended to hopefully fix that issue.


p-netm:
Closing this since its being merged to the wrong branch, opening another soon

Issue #739, opened 2020-03-10
Use @onaio/utils


moshthepitt:
Use [@onaio/utils](https://github.com/onaio/js-tools/tree/master/packages/utils)

Remove the following

- [x] Remove FlexObject
- [x] Remove UpdateType
- [x] Remove percentage

Replace them with equivalent utils from @onaio/utils.

kahummer:
@moshthepitt am experiencing a type issue when updating percentage
 ```
Type 'Result<string>' is not assignable to type 'string'.
  Type 'Failure' is not assignable to type 'string'.```.
After going into the types dts file and adding the type param to failure it works. I think this can be updated on js tools published.

moshthepitt:
@kahummer could you share the problematic code - would like to see it?

Additionally, please create a js-tools issue that you'll then work on as part of closing this one

kahummer:
src/helpers/indicators.tsx
![Screenshot from 2020-03-16 19-58-25](https://user-images.githubusercontent.com/16018123/76781989-a520e680-67c0-11ea-930c-74f88d582338.png)

moshthepitt:
@kahummer I think you should look closer at what, exactly, the `percentage` function from @onaio/utils returns (it is not a string) and the [Result type](https://github.com/onaio/js-tools/blob/master/packages/utils/src/types.ts#L48). (read the blog post in the docstring).

PR #749, opened 2020-03-16
Use ona utils

kahummer:
fix: #739


src/components/AssignTeamPopover/index.tsx
```diff
@@ -41,6 +41,7 @@ const AssignTeamPopover = (props: AssignTeamPopoverProps) => {
 
   // define props for the actual popover component
   const popoverProps = {
+    flip: true,
```
moshthepitt: Why is this change in this PR?

src/gisida.d.ts
```diff
@@ -1,27 +1,28 @@
...
+  import { Dictionary } from '@onaio/utils';
   import { Map as mbMap } from 'mapbox-gl';
   import { ParseResult } from 'papaparse';
   import { AnyAction } from 'redux';
 
   /** interface for generic object */
-  interface FlexObject {
+  interface Dictionary {
     [key: string]: any;
   }
```
moshthepitt: This looks like it should be removed, thoughts?

kahummer: yeah for sure missed that

src/components/AssignTeamPopover/index.tsx
kahummer: My bad It shouldn't be here

Issue #407, opened 2019-08-23
Improvements to Focus Area Profile View


craigappl:
(I don't know the name of this view)
![image](https://user-images.githubusercontent.com/6067053/63601217-dbbfb700-c579-11e9-9c56-4cf6588bbc86.png)

Issues:
- [x] (Low) The search box doesn't work
~~- [ ] (High) We need to support pagination on all tables on this screen. It's currently limited to six~~
- [ ] (Medium) (Optional) Sort each table in reverse chronological order based on startDate

kahummer:
@moshthepitt @craigappl The current single FI view seems not to have the search box
![Screenshot from 2020-03-16 10-46-13](https://user-images.githubusercontent.com/16018123/76734331-ce675580-6773-11ea-9a88-19ce0eee1e7f.png).

moshthepitt:
@kahama94 I removed it in #728

moshthepitt:
I suggest you leave it removed, we had kind of agreed to remove all the search boxes that do not work for now. @kahummer

PR #750, opened 2020-03-17
Profile improvement

kahummer:
fix: #407


kahummer:
@moshthepitt sure let me close this

src/constants.tsx
```diff
@@ -117,3 +117,5 @@ export const MOSQUITO_COLLECTION_CODE = 'Mosquito Collection';
 export const PRACTITIONER_CODE = {
   text: 'Community Health Worker',
 };
+/** Field to sort plans by */
+export const SORT_FIELD = 'plan_effective_period_start';
```
moshthepitt: Do you think it makes sense to name this something more specific like to mean "sort by effective period start"?

src/helpers/utils.tsx
```diff
@@ -675,3 +675,13 @@ export const setDefaultValues = (target: Dictionary, prop: string) => {
   target[prop] = [];
   return target;
 };
+
+export function reverseChronologicalSort(arr: Plan[], sortField: string) {
```
moshthepitt: - Can we add a docstring here?
- I think this name isn't accurate - it seems this is not just a reverse chronological sort

src/store/ducks/plans.ts
```diff
@@ -657,6 +660,11 @@ export const makePlansArraySelector = (planKey?: string) => {
       getPlansArrayByParentJurisdictionId(planKey),
     ],
     (plans, plans2, plans3, plans4, plans5) =>
-      intersect([plans, plans2, plans3, plans4, plans5], JSON.stringify)
+      SORT_FIELD
+        ? reverseChronologicalSort(
+            intersect([plans, plans2, plans3, plans4, plans5], JSON.stringify),
+            SORT_FIELD
+          )
+        : intersect([plans, plans2, plans3, plans4, plans5], JSON.stringify)
```
moshthepitt: This violates the [closed/open principle](https://stackify.com/solid-design-open-closed-principle/).  That is, if we ever wanted to sort by some other field we would need to modify the code instead of just changing a parameter or something similar.

Is it possible to make the sort field an optional param to `makePlansArraySelector` and get it from there?

PR #753, opened 2020-03-19
Use explicit types for function params with any

### LINKED ISSUE #681: Fix type definitions in JurisdictionsReport/index.tsx
We should not have function params that are of type `any`

kelvin-muchiri:
Closes #681 

Replace **any** type for function params with explicit types for JurisdictionReport


moshthepitt:
Nice

PR #754, opened 2020-03-19
Enhancement/641 ga tracking

### LINKED ISSUE #641: Stop un-mounting containers when GA Tracking

kelvin-muchiri:
Closes #641 

Google Analytics tracking has been enhanced to use package [@onaio/google-analytics](https://www.npmjs.com/package/@onaio/google-analytics). The previous implementation was removed.


package.json
```diff
@@ -14,6 +14,7 @@
     "@onaio/drill-down-table": "^0.0.21",
     "@onaio/element-map": "^0.0.2",
     "@onaio/gatekeeper": "^0.0.16",
+    "@onaio/google-analytics": "^0.0.2",
```
cKellyDesign: 🎉

src/App/App.tsx
```diff
@@ -71,14 +74,47 @@ library.add(faExternalLinkSquareAlt);
...
   autoClose: TOAST_AUTO_CLOSE_DELAY /** defines how long a toast remains visible on screen */,
 });
-initGoogleAnalytics();
+
+const initializeOptions = {
+  testMode: GA_ENV_TEST === GA_ENV,
+};
+
+/** Initialize google analytics */
+initGoogleAnalytics(GA_CODE, initializeOptions);
```
cKellyDesign: @kelvin-muchiri just a nit-pick: please name-space this `initializeOptions` to be more specific to google analytics 🙏

src/App/App.tsx
```diff
@@ -71,14 +74,47 @@ library.add(faExternalLinkSquareAlt);
...
 import './App.css';
 
+/** Interface defining component state */
+export interface AppState {
+  username?: string;
+}
+
 /** Main App component */
-class App extends Component {
+class App extends Component<{}, AppState> {
```
cKellyDesign: @kelvin-muchiri @moshthepitt @p-netm - how do you guys feel about converting this to a functional component? Is that within the scope of this PR or should/could it be a part of a different ticket?

src/App/App.tsx
```diff
@@ -87,214 +123,215 @@ class App extends Component {
         <Container fluid={true}>
           <Row id="main-page-row">
             <Col>
+              {<RouteTracker />}
```
cKellyDesign: @kelvin-muchiri should we check that the GA code is provided here? If it's not then we should not be tracking

src/App/App.tsx
```diff
@@ -71,14 +74,47 @@ library.add(faExternalLinkSquareAlt);
...
+    super(props);
+    this.state = {
+      username: (getUser(store.getState()) || {}).username || '',
+    };
+  }
+
+  public componentDidUpdate(_: {}, prevState: AppState) {
+    const username = (getUser(store.getState()) || {}).username || '';
+
+    if (prevState.username !== username) {
```
cKellyDesign: should we check that the GA code is provided here? If it's not then we don't need to worry about the GA dimensions or who the user is (at this level).

src/App/App.tsx
```diff
@@ -87,214 +123,215 @@ class App extends Component {
...
             <Col>
+              {<RouteTracker />}
               <Switch>
                 {/* Home Page view */}
                 <ConnectedPrivateRoute
                   disableLoginProtection={DISABLE_LOGIN_PROTECTION}
                   exact={true}
                   path="/"
-                  component={WithGATracker(Home)}
+                  component={Home}
```
cKellyDesign: DRY at last! 👍

src/App/tests/App.test.tsx
```diff
@@ -41,6 +45,10 @@ describe('App', () => {
     expect(toJson(wrapper)).toMatchSnapshot();
     wrapper.unmount();
   });
+
+  /** @todo Tests that utilize window.open have been intentionally put at the bottom
+   * since they are causing some of the tests above to fail. Investiagte why this is happening
+   */
```
cKellyDesign: 👍

cKellyDesign: Should this have its own ticket?

src/App/App.tsx
moshthepitt: @cKellyDesign lets move it to a different issue.  But I like the idea!

moshthepitt:
@kelvin-muchiri why so many changes in the "server" directory?

kelvin-muchiri:
> @kelvin-muchiri why so many changes in the "server" directory?

I included server directory in "includes" section of `tsconfig.json` and had to fix the lint warnings

src/App/tests/App.test.tsx
```diff
@@ -64,6 +68,19 @@ describe('App', () => {
...
+      <Provider store={store}>
+        <Router history={history}>
+          <App />
+        </Router>
+      </Provider>
+    );
+    expect(GoogleAnalytics.pageview).toBeCalledWith('/');
+    wrapper.unmount();
+  });
+
```
moshthepitt: @kelvin-muchiri can we navigate to some other pages and check that those pages were tracked okay?  I suppose that we'd also want to check that the username was tracked okay.

kelvin-muchiri: @moshthepitt Yea sure. I am having trouble navigating to other pages or changing the value of the envs at the moment.

src/App/App.tsx
cKellyDesign: https://github.com/onaio/reveal-frontend/issues/756

src/App/tests/App.test.tsx
cKellyDesign: @kelvin-muchiri have you tried using the pushing on the history object?
```
import { createBrowserHistory } from 'history';
const history = createBrowserHistory();
...
history.push(PLAN_LIST_URL);
```

@moshthepitt @p-netm - is this the best way to navigate?

moshthepitt: @cKellyDesign @kelvin-muchiri I would first try to find the links in the enzyme wrapper and then click on them to go to a different page.   But it is also acceptable to do history.push

cKellyDesign: > I would first try to find the links in the enzyme wrapper and then click on them to go to a different page

Good call, this more closely resembles user interaction. Since we also have places which push history, maybe we should test both cases.

kelvin-muchiri: @moshthepitt @cKellyDesign Should we really worry about the implementation in this case?

kelvin-muchiri: I actually realized other tests made the tests fail. Commenting out other tests lead to the tests passing

moshthepitt: @kelvin-muchiri Commenting out?  That sounds .... weird :)

kelvin-muchiri: @moshthepitt I think it was to do with the history. Resetting the history before each test has fixed it.

cKellyDesign: > Commenting out? That sounds .... weird :)

@kelvin-muchiri I think this means the commented out code should be removed from the file and committed so that current git history does not include commented out tests. 😄

kelvin-muchiri: @cKellyDesign  Yea sure

kelvin-muchiri:
@moshthepitt How can I remock `GA_CODE` from `../../configs/env` to re-assign and test an empty string?

kelvin-muchiri:
> @moshthepitt How can I remock `GA_CODE` from `../../configs/env` to re-assign and test an empty string? Given that the module has already being mocked `jest.mock('../../configs/env')`

@moshthepitt  I have found a solution to this.

moshthepitt:
> @moshthepitt I have found a solution to this

Fantastic!  How did you solve it @kelvin-muchiri

kelvin-muchiri:
> Fantastic! How did you solve it @kelvin-muchiri

I re-imported the module
```
 const myModule = require('../../configs/env');
    myModule.GA_CODE = '';
```

src/App/tests/App.test.tsx
```diff
@@ -110,4 +128,53 @@ describe('App', () => {
...
+      </Provider>
+    );
+    await new Promise<unknown>(resolve => setImmediate(resolve));
+    expect(GoogleAnalytics.set).toBeCalledWith({ env: 'test', username: 'superset-user' });
+    wrapper.unmount();
+  });
+
+  it('does not track dimensions of google analytics code is not set', async () => {
+    fetch.mockResponse(JSON.stringify(expressAPIResponse));
+    const envModule = require('../../configs/env');
```
moshthepitt: very nice

src/App/tests/App.test.tsx
```diff
@@ -13,12 +15,13 @@ jest.mock('../../configs/env');
...
 // tslint:disable-next-line: no-var-requires
 const fetch = require('jest-fetch-mock');
-
 const history = createBrowserHistory();
 
 describe('App', () => {
   beforeEach(() => {
     jest.resetAllMocks();
+    // Reset history
+    history.push('/');
```
cKellyDesign: 👍

Naima-Bashir:
@craigappl kindly give me access to GA account to see if I can visit pages and test it. Thanks.

PR #759, opened 2020-03-24
Fix loadPlans service call in IRS plan page

p-netm:
Bug: service call to get a plan in `IrsPlan` returns an array but the component assumes it returns a single object.
As a result the component would be unable to extract the planRecord. As consequence the plan is not added to the store.
The component just keeps spinning obliviously.
This is evident when you refresh the page, otherwise if the plan is already in the store,the page loads as expected

### DIFF
--- a/src/containers/pages/InterventionPlan/IRS/plan/index.tsx
+++ b/src/containers/pages/InterventionPlan/IRS/plan/index.tsx
@@ -260,8 +260,8 @@ class IrsPlan extends React.Component<
     // GET PLAN
     if (planId && !planById) {
       await OpenSrpPlanService.read(planId)
-        .then((plan: PlanPayload) => {
-          const planRecord = extractPlanRecordResponseFromPlanPayload(plan);
+        .then((plan: PlanPayload[]) => {
+          const planRecord = extractPlanRecordResponseFromPlanPayload(plan[0]);
           if (planRecord) {
             return fetchPlansActionCreator([planRecord]);
           }
--- a/src/containers/pages/InterventionPlan/IRS/plan/index.tsx
+++ b/src/containers/pages/InterventionPlan/IRS/plan/index.tsx
@@ -115,7 +115,6 @@ import plansReducer, {
   extractPlanRecordResponseFromPlanPayload,
   fetchPlanRecords,
   getPlanRecordById,
-  PlanPayload,
   PlanRecord,
   PlanStatus,
   reducerName as plansReducerName,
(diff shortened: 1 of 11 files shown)


src/containers/pages/InterventionPlan/IRS/plan/serviceCalls.ts
```diff
@@ -0,0 +1,31 @@
...
+  planId: UUID,
+  actionCreator: typeof fetchPlanRecords
+) {
+  const OpenSrpPlanService = new service(OPENSRP_PLANS);
+  await OpenSrpPlanService.read(planId)
+    .then((plan: PlanPayload[]) => {
+      const planRecord = extractPlanRecordResponseFromPlanPayload(plan[0]);
+      if (planRecord) {
+        actionCreator([planRecord]);
+      }
```
moshthepitt: What happens when `planRecord` is false-ey?  Can we call displayError with some error message?

src/containers/pages/InterventionPlan/IRS/plan/serviceCalls.ts
```diff
@@ -0,0 +1,31 @@
+import { OPENSRP_PLANS } from '../../../../../constants';
+import { displayError } from '../../../../../helpers/errors';
+import { OpenSRPService } from '../../../../../services/opensrp';
+import {
+  extractPlanRecordResponseFromPlanPayload,
+  fetchPlanRecords,
+  PlanPayload,
+} from '../../../../../store/ducks/plans';
+
+type UUID = string;
```
moshthepitt: Nice :)

src/store/ducks/plans.ts
```diff
@@ -245,6 +256,7 @@ export const extractPlanRecordResponseFromPlanPayload = (
     }
     return planRecordResponse;
   }
+  displayError(new Error(FAILED_TO_EXTRACT_PLAN_RECORD));
```
moshthepitt: I have mixed feelings about this.

- Does extractPlanRecordResponseFromPlanPayload belong in the dux module - it does not seem to be a reducer function, an action creator or a selector?
- Can we add docstrings to extractPlanRecordResponseFromPlanPayload?

p-netm: - I could move them to the root utils file.

src/containers/pages/InterventionPlan/IRS/plan/index.tsx
```diff
@@ -259,14 +259,9 @@ class IrsPlan extends React.Component<
...
-        .then((plan: PlanPayload) => {
-          const planRecord = extractPlanRecordResponseFromPlanPayload(plan);
-          if (planRecord) {
-            return fetchPlansActionCreator([planRecord]);
-          }
-        })
-        .catch(err => err);
+      await loadPlan(OpenSRPService, planId, fetchPlansActionCreator).catch(err =>
+        displayError(err)
+      );
```
cKellyDesign: ❤️

Issue #777, opened 2020-03-31
Monitor Dashboards on Reveal Web Loads with Grayed out Structures


samkanga:
Access to completed FI plans under Reveal Web > Monitor > Focus Investigation> Select plan

- [x] Loads a map with grayed structures. The map loads on the android client with tasks on it. See sample below:
![image](https://user-images.githubusercontent.com/5908630/78005241-f34af380-7343-11ea-85f3-548de16bd940.png)

- [x] Clicking on any of the indicators e.g. Family Registration, results into a spinner. See screenshot below:
![image](https://user-images.githubusercontent.com/5908630/78005699-9d2a8000-7344-11ea-9995-c72dbb6f93e3.png)

samkanga:
@moshthepitt and @kahummer just learnt that the color coding on the structures is gray by default when the map loads. The color coding occurs when the indicators on the right are clicked.

kahummer:
@samkanga  That then means the
The second issue is the one to be fixed ?

samkanga:
Yes @kahummer

kahummer:
@samkanga I think we should close this

moshthepitt:
@kahummer is there not an open PR for this?  If yes then we can't close it until the PR is closed.  No?

moshthepitt:
It seems that this is no longer an issue.

Naima-Bashir:
![image](https://user-images.githubusercontent.com/61143692/78764283-00dd2a80-798f-11ea-8ac8-f0ceb4b5b548.png)
Yes this is no longer an issue. Tested; it has passed QA.

PR #778, opened 2020-03-31
Fix Task selection Bug

kahummer:
fix: #777

### DIFF
--- a/src/components/GisidaWrapper/index.tsx
+++ b/src/components/GisidaWrapper/index.tsx
@@ -275,8 +275,7 @@ class GisidaWrapper extends React.Component<GisidaProps, GisidaState> {
      */
     if (
       (some(pointFeatures) || some(polygonFeatures)) &&
-      nextProps.currentGoal !== this.props.currentGoal &&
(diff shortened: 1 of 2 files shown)


kahummer:
Findings: 

- The parent component had been refactored to use react hooks

- On task selection GisidaWrapper state takes default state and before we set locations from incoming props doesn't get to call initmap (this was the main issue)

- I removed the condition which relied on locations being set to state and it worked.

The tests have not reflected the change made to this component this is because I couldn't get the component to run certain methods and set state accordingly and had to force them to do so.

kahummer:
@moshthepitt @p-netm I haven't been able to spot the root issue that caused this bug. In addition to the comments [above](https://github.com/onaio/reveal-frontend/pull/778#issuecomment-607101056). The release prior to this fix had this issue but seems to be working now. This  may require more investigation. I created an issue based on a convo we had with @p-netm  https://github.com/onaio/reveal-frontend/issues/792 in relation to this issue

moshthepitt:
Seems like this is not an issue any longer.

PR #783, opened 2020-04-03
Test config types

### LINKED ISSUE #721: Investigate and add tests for config/typings
- investigate and add tests for `src/configs/types.ts`

p-netm:
close #721
- Adds the folder test-d/ to contain the types
- we are using tsd to check the types
- Added a test:types script to package.json
- Apparently we need a root index.d.ts for tsd to work
  SamVerschueren/tsd#32 (comment)


src/test-d/types.test-d.ts
```diff
@@ -0,0 +1,48 @@
+import * as React from 'react';
+import { expectAssignable, expectType } from 'tsd';
+import {
+  adminLayerColorsType,
+  ADMN0_PCODE,
+  EmptyObject,
+  FlexComponent,
```
moshthepitt, line 7: Should we use this one istead? https://github.com/onaio/js-tools/blob/master/packages/react-utils/src/types.ts#L7

p-netm: yeah will remove that redundancy

p-netm:
@moshthepitt @cKellyDesign this might require re-review before merge.

PR #790, opened 2020-04-03
Refactor IRS reporting-legend's text

### LINKED ISSUE #774: Edit the percentages on the IRS Reporting Color Codes
Edit the percentage ranges on the IRS Reporting color codes. Presently it appears as follows:
image

They should read:
- [ ] Grey < 20%
- [ ] Red  20% - <75%
- [ ] Yellow 75% - < 90%
- [ ] Green 90% -  100%

This is to avoid the overlap of the figures

p-netm:
The range symbols( < - >) or operators will now also respect the `orEquals` IndicatorItem property. for example an indicatorItem with {orEquals: true, value: 20} will now be shown as `- 20%` and not `< 20%`

close #774

### DIFF
--- a/src/components/formatting/IRSIndicatorLegend/index.tsx
+++ b/src/components/formatting/IRSIndicatorLegend/index.tsx
@@ -10,7 +10,7 @@ import {
   IndicatorThresholdsLookUp,
   indicatorThresholdsLookUpIRS,
 } from '../../../configs/settings';
-import { percentage } from '../../../helpers/utils';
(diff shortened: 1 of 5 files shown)


p-netm:
updatedScreenshortReporting

moshthepitt:
> Coverage decreased (-6.07%) to 69.589%

WOW.  Why?

Naima-Bashir:
image
Passed QA.

Issue #782, opened 2020-04-02
Session expires while users is logged in(This is not gracefully handled)


p-netm:
once the session expires while a user is logged in, opensrp service calls begin failing.
However this is a hypothesis albeit one that is very likely.
The tasks for this:
-  confirm the hypothesis
- enhance code to gracefully handle the issue.

moshthepitt:
@p-netm should we also create this issue on the expres-server repo?

Issue #776, opened 2020-03-31
RVL-806 Logout of OAuth when you hit logout button in WebUI


samkanga:
https://smartregister.atlassian.net/browse/RVL-806
Signing out of the WebUI does not sign the user out of OpenSRP  OAuth session, and the user can log back in without re-entering their password

**Current behaviour**

1. Click the user name in top right corner of the WebUI and Click Sign out
2. 
![image](https://user-images.githubusercontent.com/5908630/77999010-aa427180-733a-11ea-854a-5882c048a410.png)

3. User is redirected to 
![image](https://user-images.githubusercontent.com/5908630/77999029-b29aac80-733a-11ea-8cb9-53773efdad61.png)

4. Click OpenSRP, user redirected to 
![image](https://user-images.githubusercontent.com/5908630/77999054-bdedd800-733a-11ea-8846-aadd1109e09f.png)

5. Click Authorize

6. User is logged in to the WebUI 

**Expected/Desired behaviour**

Step 1 should send the user to https://reveal-stage.smartregister.org/opensrp/logout.do. When the user reaccesses openSRP, s/he should have to reenter their password.

Try and make the web app also log out of the OpenSRP server. Perhaps when the web app is logging you out it can make an asynchronous request to https://reveal-stage.smartregister.org/opensrp/logout.do and thus we can assume you’re logged out of both systems. Although we must note that this 2nd option is not standard at all and is not part of the OAuth spec. We can only make it work due to the (lucky?) fact that we control both systems.

kahummer:
@moshthepitt @p-netm 
To close this issue we need to logout the user from our reveal and opensrp (this is working well without the backend server using the connectedlogout component.
When using the backend server we have to load the connectedlogout component and move execution to the express server to destroy sessions and clear cookies. Running both of these 
```
<ConnectedLogout
                      {...{
                        logoutURL: 'https://reveal-stage.smartregister.org/opensrp/logout.do',
                        redirectPath: null,
                      }}
                    />;
                    window.location.href = EXPRESS_OAUTH_LOGOUT_URL;
```
Loads the connected logout and fails to go into the express server. I also tried running the redirect on js-tools but it doesn't work either.
I'll try create a custom component that implements what we have on js-tools but also redirects to express as a workaround

moshthepitt:
@kahummer I think right now it roughly works like so:

1. Click logout
2. It sends you to express which clears its session
3. It brings you back to the frontend (I forget which specific component)

[...]

Issue #770, opened 2020-03-27
Test Reaveal-th login using a cooking jar


p-netm:
related to #768 .
and using this link: https://stackoverflow.com/a/19937138

Issue #768, opened 2020-03-26
Reveal Thailand Production session 500 error


craigappl:
We're getting the following error code when trying to login to the thailand production environment through the web UI:

```
{
message: "The provided authorization grant (e.g., authorization code, resource owner credentials) or refresh token is invalid, expired, revoked, does not match the redirection URI used in the authorization request, or was issued to another client.",
status: "error",
statusCode: 500
}
```

### Triaging the problem
1 Starting with just the google analytics cookies on web.reveal-th
![image](https://user-images.githubusercontent.com/6067053/77685431-adafb480-6f58-11ea-8fef-489dc986605a.png)

2 Click OpenSRP button which redirects to https://reveal-th.smartregister.org/opensrp/login.jsp
Five cookies in use
Two for google analytics
reveal-th.smartregister.org
- AWSALB: lQ6z+3mOJ0Rxx4aMJHe/Aan6MlGNfqdr9Ingn0c
- AWSALBCORS: lQ6z+3mOJ0Rxx4aMJHe/Aan6MlGNfqdr9Ingn0c
- JSESSIONID: 85302A2412074B69EDC3F61635E4EFD7

3 I enter the username and password and click login. I'm redirected to the OAuth Approval view
https://reveal-th.smartregister.org/opensrp/oauth/authorize?client_id=52vjwavfeLkMxZArGorgAoExwicfDGb49rVy&redirect_uri=https%3A%2F%2Fweb.reveal-th.smartregister.org%2Foauth%2Fcallback%2FOpenSRP%2F&scope=read%20write&response_type=code&state=opensrp
Five cookies in use
Two for google analytics
reveal-th.smartregister.org
- AWSALB: g+D58fHD44XeKLLh+J0qxR6PWu1eSNqm2cY/Fum7Z2OWkF1pPl6DpnsKb5rJWemMhw+eiT/0PcvR69QZ0m7Y1xzDeeq0cEWzH99HiT0IC8wk7q0F5y2wUQVtSV8e
- AWSALBCORS: g+D58fHD44XeKLLh+J0qxR6PWu1eSNqm2cY/Fum7Z2OWkF1pPl6DpnsKb5rJWemMhw+eiT/0PcvR69QZ0m7Y1xzDeeq0cEWzH99HiT0IC8wk7q0F5y2wUQVtSV8e
- JSESSIONID: 3319974A520925A41DED81C3899B242F
(Note that the JSESSIONID changed

[...]

PR #793, opened 2020-04-07
Remove server from reveal-frontend repo

### LINKED ISSUE #738: Add server tests to CI process

### LINKED ISSUE #742: Move server out of reveal FE
- This is a continuation of work that's taking place here #735. the next steps; steps that would close this issue is to consider the server as a repo in itself.

The 2 schools of thought around this:
- to create a new repo for the server. This would make deployment and setting up CI/CD eaiser. For the latter it would really not be that significant of a factor. The cons for this would be one more manual thing to check; i.e. that versions of server and version of FE can actually work together. Making changes to either server or FE for cross cutting issues would need one to also be vigilant to make corresponding changes to the other.
- To have this as a monorepo: The con expressed above becomes a non-issue. some extra negligible effort in deployment.
- Other that i dont know.

Some of other things that might be affected by this(This is not exhaustive):
- #738 
- [github/onaio/ansible-express](https://github.com/onaio/ansible-express)

p-netm:
close #738
close #742

Affects:
 - #782
 - #776
 - #768
 - #770
 - #716
 - #635

when the `REACT_APP_BACKEND_ACTIVE` env variable is set to true.
There is need to also setup the backend(a tiney express server).
One option for this is located at github.com/onaio/express-server.


p-netm:
I suppose, will close this alongside onaio/reveal-express#2

p-netm:
> Are there any packages in the main package.json that should be removed?

Doesn't seem like there are any.

Issue #798, opened 2020-04-14
UI/UX Differentiating between errored response and empty response


p-netm:
The logic of several pages is to show the loader based on the presence of some data that is supposed to be rendered on the respective page. This is not clear on whether any async operations to fetch the data were erred or we just got an empty response. I suppose it would lead for better UX if we could inform the user in cases where the web app completed the data fetching operations successfully but just didn't get any response data to render.

PR #802, opened 2020-04-15
Feature/772 search

### LINKED ISSUE #772: RVL-760 Enable WebUI Search for Plans in Plan and Monitor screens
Enable search for the Plan entity in the Plan and Monitor screens as follows:

- [x] Plans > Manage Plans
- [x] Monitor > Focused Investigations
- [x] Monitor > IRS Reporting

**Acceptance Criteria:**

1. Allow a user to enter text within the search box
2. The results returned should filter results based on text entered within the search box

kelvin-muchiri:
Closes #772 

Added a reusable search form that is used for search. The search is done by calling `Array.filter` on the list received from props and is accompanied by a delayed search (debounce) of 1s to make sure the filtering is not done too frequently. The search filter is currently by name/title only


moshthepitt:
**Question**: why not do this using a [reselect](https://github.com/reduxjs/reselect) selector at the store level?  It should be possible to make that re-useable as well, right?

I'm leaning towards that because:

1. What you are doing here (i.e. computing/deriving data from state) is traditionally what selectors are used for in Redux-land
2. Using reselect may have performance benefits

Let me know what you think.

kelvin-muchiri:
@moshthepitt Let me look into the libabry never worked with it before. It may be of help

src/containers/pages/FocusInvestigation/active/index.tsx
```diff
@@ -124,8 +141,33 @@ class ActiveFocusInvestigation extends React.Component<
       .then((result: Plan[]) => fetchPlansActionCreator(result))
       .catch(err => displayError(err));
   }
-  public handleSubmit(event: React.FormEvent<HTMLFormElement>) {
-    event.preventDefault();
+
+  public debouncedSearch(event: React.ChangeEvent<HTMLInputElement>) {
```
cKellyDesign, line 145: How / where is this being debounced? Looks like it fires every time the input value is updated (which should be fine I think) 🤔

src/helpers/hooks.tsx
```diff
@@ -17,3 +17,35 @@ export function useConfirmOnBrowserUnload(hasUnsavedChanges: boolean = false) {
...
   });
 }
+
+/**
+ * Debounce calls to make sure they do not execute too frequently such as when
+ * a user is typing and a call needs to be made to the API
+ * @param value Value to be debounced
+ * @param delay Time in ms to wait for since the last call
+ */
+export function useDebounce(value: string, delay: number = 1000) {
```
cKellyDesign, line 27: Is this being used anywhere?

src/containers/pages/FocusInvestigation/active/index.tsx
```diff
@@ -108,11 +116,20 @@ export const defaultActiveFIProps: ActiveFIProps = {
 /** Reporting for Active Focus Investigations */
 class ActiveFocusInvestigation extends React.Component<
   ActiveFIProps & RouteComponentProps<RouteParams>,
-  {}
+  ActiveFIState
 > {
   public static defaultProps: ActiveFIProps = defaultActiveFIProps;
+
+  public debouncedSearchCallback = _.debounce(this.debouncedSearch, 1000);
```
cKellyDesign, line 123: Oh I see 👍

src/helpers/hooks.tsx
kelvin-muchiri: Yes:

1. `src/containers/pages/InterventionPlan/PlanDefinitionList/index.tsx
2. `src/containers/pages/IRS/plans/index.tsx`

kelvin-muchiri:
@moshthepitt I was able to use reselect for `src/containers/pages/FocusInvestigation/active/index.tsx`.  For the other components, the enhancement to use reselect might not meet this PR.

kelvin-muchiri:
@moshthepitt I was also partially getting confused due to this part of code

```
 */
export function getPlansArray(
  state: Partial<Store>,
  intervention: InterventionType | null = null,
  statusList: string[] = [],
  reason: string | null = null,
  jurisdictions: string[] = [],
  parentJurisdictionId: string | null = null
): Plan[] {
  return values((state as any)[reducerName].plansById).filter(
    (plan: Plan) =>
      (intervention ? plan.plan_intervention_type === intervention : true) &&
      (statusList.length ? statusList.includes(plan.plan_status) : true) &&
      (reason ? plan.plan_fi_reason === reason : true) &&
      (jurisdictions.length ? jurisdictions.includes(plan.jurisdiction_id) : true) &&
      (parentJurisdictionId && !plan.jurisdiction_path ? false : true) &&
      (parentJurisdictionId && plan.jurisdiction_path
        ? plan.jurisdiction_path.includes(parentJurisdictionId)
        : true)
  );
}
```
This should use selectors right?

moshthepitt:
> @moshthepitt Shouldn't this part use selectors?

@kelvin-muchiri that's correct.  The context is that reselect is being added slowly.  Whatever needs https://github.com/onaio/reveal-frontend/blob/master/src/store/ducks/plans.ts#L294 can probably be refactored to use https://github.com/onaio/reveal-frontend/blob/master/src/store/ducks/plans.ts#L523.

src/containers/pages/FocusInvestigation/active/index.tsx
```diff
@@ -124,8 +145,35 @@ class ActiveFocusInvestigation extends React.Component<
...
+        searchedRoutinePlans: makePlansArraySelector()(store.getState(), {
+          interventionType: InterventionType.FI,
+          parentJurisdictionId: this.props.jurisdictionParentId,
+          reason: ROUTINE,
+          statusList: [PlanStatus.ACTIVE, PlanStatus.COMPLETE],
+          title: event.target.value,
+        }),
+      });
+    }
+  }
```
moshthepitt, line 172: I think that this should probably be done in `mapStateToProps` - no?  I think that is the standard approach.

src/containers/pages/FocusInvestigation/active/index.tsx
```diff
@@ -401,7 +441,7 @@ class ActiveFocusInvestigation extends React.Component<
               columns: emptyPlansColumns,
             };
             routineReactivePlans.push(
-              <NullDataTable tableProps={tableProps} reasonType={header} key={`${'current'}`} />
+              <NullDataTable tableProps={tableProps} reasonType={header} key={i} />
```
moshthepitt, line 444: why was this done?

src/containers/pages/FocusInvestigation/active/index.tsx
```diff
@@ -421,6 +461,7 @@ export { ActiveFocusInvestigation };
 
 /** interface to describe props from mapStateToProps */
 interface DispatchedStateProps {
+  jurisdictionParentId: string;
```
moshthepitt, line 464: What caused the need for this?

src/containers/pages/FocusInvestigation/active/index.tsx
```diff
@@ -453,6 +494,7 @@ const mapStateToProps = (state: Partial<Store>, ownProps: any): DispatchedStateP
   );
   return {
     caseTriggeredPlans,
+    jurisdictionParentId,
```
moshthepitt, line 497: What caused the need for this?

src/containers/pages/FocusInvestigation/active/index.tsx
kelvin-muchiri: @moshthepitt Ok...but how to we get the input value from `mapStateToProps`?

src/containers/pages/FocusInvestigation/active/index.tsx
kelvin-muchiri: If you have more than two nullable tables they both have the same key

src/containers/pages/FocusInvestigation/active/index.tsx
kelvin-muchiri: I need to use it in the component when making the call to `makePlansArraySelector()`

src/containers/pages/FocusInvestigation/active/index.tsx
moshthepitt: I think that state is accessible from mapStateToProps.  Right?  If so then if the search value is in state then it can work.

src/containers/pages/IRS/plans/index.tsx
```diff
@@ -73,16 +79,26 @@ const IRSPlansList = (props: PlanListProps) => {
...
-  }
+  useEffect(() => {
+    if (plans && debouncedSearchQuery) {
+      setSearchedPlans(
+        plans.filter((plan: IRSPlan) =>
+          plan.plan_title.toLowerCase().includes(debouncedSearchQuery.toLowerCase())
+        )
+      );
+    }
+  }, [debouncedSearchQuery]);
```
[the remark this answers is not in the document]
kelvin-muchiri: @moshthepitt  Ok..Wanted to have it working for the release first on Friday but I can move it to a selector

src/containers/pages/FocusInvestigation/active/index.tsx
moshthepitt: What value of `jurisdictionParentId` are you going to pass to it?  Also, are we 100% certain that `makePlansArraySelector()` requires `jurisdictionParentId`?

src/containers/pages/FocusInvestigation/active/index.tsx
kelvin-muchiri: > I think that state is accessible from mapStateToProps. Right? If so then if the search value is in state then it can work.

Ok..Let me try

kelvin-muchiri: @moshthepitt I think not unless we move the search value to redux store? @p-netm What to you think?

p-netm: yeah I don't hink we can access state from mapStateToProps. but we can still use the selectors inline inside the component, I think.

src/containers/pages/IRS/plans/index.tsx
```diff
@@ -6,6 +6,7 @@ import { connect } from 'react-redux';
 import { Link } from 'react-router-dom';
```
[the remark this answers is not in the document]
p-netm: We can probably have another pr that addresses this. The conversation at #806 is a nice start but I think we do not yet have a definitive standard. The PR would close both #806 and #798

src/containers/pages/InterventionPlan/PlanDefinitionList/index.tsx
```diff
@@ -6,6 +6,7 @@ import { connect } from 'react-redux';
 import { Link } from 'react-router-dom';
```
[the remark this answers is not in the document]
p-netm: We can probably have another pr that addresses this. The conversation at #806 is a nice start but I think we do not yet have a definitive standard. The PR would close both #806 and #798

src/containers/pages/FocusInvestigation/active/index.tsx
p-netm: I have been thinking about this, and i think if we have to supply all the prop options to the inlined selectors like done in the PR, I would go with filtering through the plans within the component.

moshthepitt: Basically I feel extremely uncomfortable doing the filtering within the component.  I am yet to see anything convincing about why we would want that.

src/containers/pages/FocusInvestigation/active/index.tsx
kelvin-muchiri: > What caused the need for this?

I need to use it in the component when making the call to makePlansArraySelector()

src/containers/pages/FocusInvestigation/active/index.tsx
kelvin-muchiri: > What value of `jurisdictionParentId` are you going to pass to it? Also, are we 100% certain that `makePlansArraySelector()` requires `jurisdictionParentId`?

Was trying to achieve what this function is doing but with an added filter of `title`

```
getPlansArray(
    state,
    InterventionType.FI,
    [PlanStatus.ACTIVE, PlanStatus.COMPLETE],
    CASE_TRIGGERED,
    [],
    jurisdictionParentId
  )
```

src/containers/pages/FocusInvestigation/active/index.tsx
p-netm: @kelvin-muchiri @moshthepitt  hmm,.. then seems as intended, the searchText will also have to be stored in the redux store, the debounced search functions implementation should only dispatch an action to modify this value. My only worry with this approach is that it feels to me like abusing the plans ducks module.

p-netm: I also think we need a final decision on this, like now. or we can create an issue.

moshthepitt: > abusing the plans ducks module

@p-netm how so?

p-netm: @moshthepitt taking the plans ducks for instance. it stores plans data fetched from some data source. In this regard it can be re-used anywhere where such data is in needed without assumptions as to how the component maintains its own UI state(the search string would be one this). If anything its almost nitpickish, but wanted to bring it up nevertheless.

moshthepitt: It is not yet clear to me whether we 100% have to store the search text in Redux.  

Additionally, keep in mind that we are going to have to support search in a lot of our components.  We should come up with something that is clean and re-usable.  In this very PR we are doing the same search functionality in three different components, for instance. 

cc: @cKellyDesign @kelvin-muchiri @kahummer

cKellyDesign: > It is not yet clear to me whether we 100% have to store the search text in Redux.

Interesting, could we simply push it to a URL query param, which the containers would reference via the router props?

This could allow us to grab the string and pass it to our existing selectors as another filter parameter all in `mapStateToProps` (which I agree is how/where this filtering should be happening).

@moshthepitt @kelvin-muchiri @p-netm

p-netm: @moshthepitt If I understand you correctly then we are not going to store the searchString in the store, then that leaves us with atleast these 2 options:

- filter the plans inline in the component using the canonical `Array.filter`.
  - which is less than ideal
- filter the plans inline using the selectors
  - Introduce another gateway for data coming in from the store, which is even lesser as far as the ideal spectrum goes.

This is because we only have access to the searchString inside the component. AFAIK, we can only use props in mapStateToProps not the component's state

moshthepitt: @p-netm @kelvin-muchiri I think @cKellyDesign's input [here](https://github.com/onaio/reveal-frontend/pull/802#discussion_r411467588) is brilliant. It seems if the search terms is part of the URL we can get at least two benefits that I care a lot about:

1. We can access the URL param in mapStateToProps
2. We can now have a URL that will lead you to the filtered/searched page i.e. user's can share this URL to render already filtered pages.

Unless there is an objection, I vote that we go with that.

Thanks @cKellyDesign

p-netm: > Interesting, could we simply push it to a URL query param, which the containers would reference via the router props?

I like this option

moshthepitt: >  It seems if the search terms is part of the URL 

@kelvin-muchiri @p-netm my suggestion on how to best do this would be to have the search form submit a GET request to the URL in which it exists.  Doing this keeps us honest (we are using forms in the exact way they were intended) and makes it trivial to add more filter fields to the same form that work the same way.

moshthepitt: This is actually very similar to how we did it in Kaznet (and I bet elsewhere): https://github.com/onaio/kaznet-frontend/blob/master/src/containers/tasks/TasksList.js#L45

kelvin-muchiri: @moshthepitt @cKellyDesign @p-netm I like the idea. But I think we cannot avoid using the store to store the search string. Cause we need to use the update string in `mapStateToProps`

cKellyDesign: Updated string?

p-netm: @kelvin-muchiri any updates on the search string should be reflected in an updated url

kelvin-muchiri: @cKellyDesign @p-netm Just to be on the same page. Are we going to store the searched value in redux store?

cKellyDesign: No I don't think that will be necessary

kelvin-muchiri: Ok...but `mapStateToProps` will not fire if the state hasn't changed so if we move the selectors to `mapStateToProps` and update the URL, will it really work?

cKellyDesign: Isn't one of the stores is the state of the router, which should update upon `history.push()`?

kelvin-muchiri: @cKellyDesign Yea its working

moshthepitt: @kelvin-muchiri How are you implementing this?  Just checking, because as I mentioned above, I think the heavy lifting should be done in the form, both because this is a natural use of HTML forms, and it makes it trivial to re-use the code.

kelvin-muchiri: I am changing the search to happen on submit.

kelvin-muchiri: I am retroducing the submit button

kelvin-muchiri: @moshthepitt @cKellyDesign @p-netm I also realized that for most views (if not all) absence of data renders a loader rather. That means even if a call to the API was a success but  empty response was returned, the loader will still show. Which I think from a user experience point of view its confusing

kelvin-muchiri: Once a call to the API is successful we should stop the loader and display the data if not display no data message

moshthepitt: @kelvin-muchiri 100% agreed.  That is also the subject of this discussion issue https://github.com/onaio/reveal-frontend/issues/806 - please contribute!

cKellyDesign: > That means even if a call to the API was a success but an empty response was returned, the loader will still show.

We should very much have a standard way of knowing when a request is finished and that the UI should update accordingly. (#806)

In this case, (if we know our request has completed, and) if we know that our selector returns no plans, then we can implicitly know that there are not going to be any plans to show (at this time), so we can at that point remove the spinner, and then indicate that there are no plans matching the search parameters.

If at this point, by some asynchronous function a plan matching the filter criteria is added to the store, the container props will update, which will trigger component to re-render/return with the fresh plan showing in the table (replacing our message of sincerest apologies that no plans were found 😉).

kelvin-muchiri:
I also think we should have discussions on ways to preserve the user's search. Not sure if it's on this PR or a later issue. If a user searches a plan, then click on a plan from the results returned, if they navigate back either using the breadcrumbs or the browser back button, shouldn't they find the results of their previous search?

src/store/ducks/opensrp/PlanDefinition/index.ts
```diff
@@ -166,3 +168,35 @@ export function getPlanDefinitionsArray(
...
+  title?: string /** plan object title */;
+}
+
+/** planDefinitionsArrayBaseSelector select an array of all plans
+ * @param state - the redux store
+ */
+export const planDefinitionsArrayBaseSelector = (planKey?: string) => (
+  state: Registry
+): PlanDefinition[] =>
+  values((state as any)[reducerName][planKey ? planKey : 'planDefinitionsById']);
```
p-netm, line 185: Do we still need to pass planKey as an argument here. It seems to me that it is not needed in this module. The reason it is required in `src/store/plans` ducks, is because the store state in that ducks has 2 top level properties, and thus one has to choose which of them they will use when creating the selector.

cKellyDesign:
@kelvin-muchiri I think hitting the back button should navigate back to the url which includes the search string in the URL, and I think clicking on the breadcrumb should not preserve the search filter.

kelvin-muchiri:
> I think hitting the back button should navigate back to the url which includes the search string in the URL

True. I've Confirmed this

src/store/ducks/generic/plans.ts
```diff
@@ -172,3 +174,35 @@ export function getIRSPlansArray(
...
+/** This interface represents the structure of IRS plan filter options/params */
+export interface IRSPlanFilters {
+  plan_title?: string /** IRS plan title */;
+}
+
+/** IRSPlansArrayBaseSelector select an array of all plans
+ * @param state - the redux store
+ */
+export const IRSPlansArrayBaseSelector = (planKey?: string) => (state: Registry): IRSPlan[] =>
+  values((state as any)[reducerName][planKey ? planKey : 'IRSPlansById']);
```
p-netm, line 189: Do we still need to pass planKey as an argument here. It seems to me that it is not needed in this module. The reason it is required in `src/store/plans` ducks, is because the store state in that ducks has 2 top level properties, and thus one has to choose which of them they will use when creating the selector.

src/containers/pages/InterventionPlan/PlanDefinitionList/index.tsx
```diff
@@ -143,8 +150,13 @@ interface DispatchedStateProps {
...
 /** map state to props */
-const mapStateToProps = (state: Partial<Store>): DispatchedStateProps => {
-  const planDefinitionsArray = getPlanDefinitionsArray(state);
+const mapStateToProps = (state: Partial<Store>, ownProps: any): DispatchedStateProps => {
+  const searchString = trimStart(ownProps.location.search, '?');
+  const queryParams = querystring.parse(searchString);
+  const searchedTitle = queryParams.search as string;
+  const planDefinitionsArray = !searchedTitle
+    ? getPlanDefinitionsArray(state)
+    : getPlanDefinitionsArrayByTitle()(store.getState(), { title: searchedTitle });
```
p-netm, line 159: Doesn't seem like this condition is required. IMO i.e. seems like it's already handled in the selector

src/containers/pages/InterventionPlan/PlanDefinitionList/index.tsx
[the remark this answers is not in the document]
p-netm: @moshthepitt @cKellyDesign is [this](https://github.com/onaio/reveal-frontend/pull/802/files#r411319729) ok,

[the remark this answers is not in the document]
cKellyDesign: Ok with me if this gets started and standardized outside of this PR

src/components/forms/Search/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+  const handleSearchChange: Change = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setSearchQuery(event.target.value);
+  };
+
+  const handleSubmit: Submit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    props.history.push({
+      search: `?search=${searchQuery}`,
+    });
```
p-netm, line 36: @kelvin-muchiri Do you think it would be a good Idea to only push a new url if the searchQuery is not an empty string, or maybe activate the search button only when there is some input.

kelvin-muchiri: Since we reintroduced the search button, it is also used to clear the search

src/store/ducks/generic/plans.ts
kelvin-muchiri: @p-netm I see what you mean. But can we leave it for scalability?

kelvin-muchiri: Or we are sure multiple level properties will not occur in future for IRSPlans?

src/components/forms/Search/index.tsx
kelvin-muchiri: Basically to clear the search results, clear the search input then submit.

src/store/ducks/generic/plans.ts
p-netm: > @p-netm I see what you mean. But can we leave it for scalability?

ok, cool

src/components/forms/Search/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+ * Interface for SearchForm props
+ */
+export interface SearchFormProps {
+  history: History;
+  location: Location;
+  placeholder?: string;
+}
+
+/** Search Form component */
+export const SearchForm = (props: SearchFormProps) => {
```
moshthepitt, line 28: I think that eventually we want this to be a [formik form](https://jaredpalmer.com/formik/).  But I am fine with us doing that in the future.  cc: @p-netm

src/components/forms/Search/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+ */
+export type Submit = (event: React.FormEvent<HTMLFormElement>) => void;
+
+/**
+ * Interface for SearchForm props
+ */
+export interface SearchFormProps {
+  history: History;
+  location: Location;
+  placeholder?: string;
```
moshthepitt, line 24: Can we make this not optional?  It seems we already have a desired default, which can be supplied to this component via **defaultProps**.  [Parse, don't validate](https://news.ycombinator.com/item?id=21476261).

src/components/forms/Search/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+  }, []);
+
+  const handleSearchChange: Change = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setSearchQuery(event.target.value);
+  };
+
+  const handleSubmit: Submit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    props.history.push({
+      search: `?search=${searchQuery}`,
```
moshthepitt, line 45: Could we please have the string "search" come from our constants module?

src/containers/pages/FocusInvestigation/active/index.tsx
```diff
@@ -95,14 +98,18 @@ export interface ActiveFIProps {
...
   plan: Plan | null;
+  jurisdictionParentId: string;
+  searchedTitle: string | null;
 }
 
 /** default props for ActiveFI component */
 export const defaultActiveFIProps: ActiveFIProps = {
   caseTriggeredPlans: null,
   fetchPlansActionCreator: fetchPlans,
+  jurisdictionParentId: '',
```
moshthepitt, line 109: I think that `makePlansArraySelector` does not require `jurisdictionParentId` so I believe it is not necessary for us to do this.

src/containers/pages/FocusInvestigation/active/index.tsx
```diff
@@ -436,26 +432,29 @@ const mapStateToProps = (state: Partial<Store>, ownProps: any): DispatchedStateP
...
-    InterventionType.FI,
-    [PlanStatus.ACTIVE, PlanStatus.COMPLETE],
-    ROUTINE,
-    [],
-    jurisdictionParentId
-  );
+  const searchString = trimStart(ownProps.location.search, '?');
+  const queryParams = querystring.parse(searchString);
+  const searchedTitle = queryParams.search as string;
+  const caseTriggeredPlans = makePlansArraySelector()(store.getState(), {
```
moshthepitt, line 438: Why are we doing `store.getState()`, do we not already have `state` available here?

src/containers/pages/FocusInvestigation/active/index.tsx
```diff
@@ -436,26 +432,29 @@ const mapStateToProps = (state: Partial<Store>, ownProps: any): DispatchedStateP
...
+    interventionType: InterventionType.FI,
+    parentJurisdictionId: jurisdictionParentId,
+    reason: CASE_TRIGGERED,
+    statusList: [PlanStatus.ACTIVE, PlanStatus.COMPLETE],
+    title: searchedTitle,
+  });
+  const routinePlans = makePlansArraySelector()(store.getState(), {
```
moshthepitt, line 445: Why are we doing `store.getState()`, do we not already have `state` available here?

src/containers/pages/IRS/plans/index.tsx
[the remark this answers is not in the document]
moshthepitt: @p-netm @kelvin-muchiri can we please create an issue to address this ^^ please?

src/components/forms/Search/index.tsx
p-netm: what I meant  was that you would have submit disabled untill we had some input in the text field, I do not think it collides with this.

> Basically to clear the search results, clear the search input then submit.

Anyway I do not have any strong feelings around this, so I gues its still fine, the way its is.

src/containers/pages/InterventionPlan/PlanDefinitionList/index.tsx
```diff
@@ -22,12 +26,12 @@ import { PlanDefinition, planStatusDisplay } from '../../../../configs/settings'
 import { HOME_URL, OPENSRP_PLANS, PLAN_LIST_URL, PLAN_UPDATE_URL } from '../../../../constants';
 import { displayError } from '../../../../helpers/errors';
 import { OpenSRPService } from '../../../../services/opensrp';
+import store from '../../../../store';
 import planDefinitionReducer, {
   fetchPlanDefinitions,
-  getPlanDefinitionsArray,
+  getPlanDefinitionsArrayByTitle,
```
moshthepitt, line 32: I wonder if it is possible to use `makePlansArraySelector` here?

src/containers/pages/IRS/plans/index.tsx
```diff
@@ -22,9 +26,10 @@ import { planStatusDisplay } from '../../../../configs/settings';
 import { HOME_URL, REPORT_IRS_PLAN_URL } from '../../../../constants';
 import { displayError } from '../../../../helpers/errors';
 import supersetFetch from '../../../../services/superset';
+import store from '../../../../store';
 import IRSPlansReducer, {
   fetchIRSPlans,
-  getIRSPlansArray,
+  getIRSPlansArrayByTitle,
```
moshthepitt, line 32: I wonder if it is possible to use `makePlansArraySelector` here?

src/containers/pages/InterventionPlan/PlanDefinitionList/index.tsx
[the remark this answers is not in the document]
moshthepitt: @p-netm @kelvin-muchiri can we please create an issue to address this? ^^

src/containers/pages/InterventionPlan/PlanDefinitionList/tests/__snapshots__/index.test.tsx.snap
```diff
@@ -1,7 +1,8 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
```
moshthepitt, line 1: @kelvin-muchiri unfortnately/fortunately you've run into one of those "poisoned chalice" components which still has a huge snapshot test :) - We need to make this snapshot smaller.

src/components/forms/Search/index.tsx
cKellyDesign: @kelvin-muchiri @p-netm if we allow the user to submit with an empty string, could we just navigate to the current url without the search params (effectively just showing unfiltered data)?

src/components/forms/Search/index.tsx
cKellyDesign: @kelvin-muchiri will you create a ticket for this before merging? 🙏

src/containers/pages/FocusInvestigation/active/index.tsx
```diff
@@ -4,17 +4,21 @@ import DrillDownTable from '@onaio/drill-down-table';
 import reducerRegistry from '@onaio/redux-reducer-registry';
 import superset from '@onaio/superset-connector';
 import { Dictionary } from '@onaio/utils';
+import _ from 'lodash';
+import { trimStart } from 'lodash';
```
cKellyDesign, line 8: @kelvin-muchiri should grab what we need from lodash instead of bringing in the whole library?

kelvin-muchiri: @cKellyDesign This is actually now unused. Seems like the pre-commit hook ignored it maybe cause its a dash?

src/containers/pages/InterventionPlan/PlanDefinitionList/tests/__snapshots__/index.test.tsx.snap
kelvin-muchiri: @moshthepitt The snapshot is huge the fixtures are an array of plans which are then rendered as a table. Looks good to me. What do you think?

src/containers/pages/InterventionPlan/PlanDefinitionList/index.tsx
kelvin-muchiri: > can we please create an issue to address this ^^ please?

@moshthepitt What issue is this?

src/containers/pages/InterventionPlan/PlanDefinitionList/tests/__snapshots__/index.test.tsx.snap
kelvin-muchiri: @moshthepitt But yea..its possible to reduce it since we are not interested in testing the `ListView` but the data passed to it as prop

moshthepitt: @kelvin-muchiri we do not need the fixtures to appear in the Snapshot.  We can use [property matchers](https://jestjs.io/docs/en/snapshot-testing#property-matchers) for this.

- Just posting this for reference: [Effective snapshot testing](https://kentcdodds.com/blog/effective-snapshot-testing/)

src/containers/pages/InterventionPlan/PlanDefinitionList/index.tsx
moshthepitt: @kelvin-muchiri the fact that when we get an empty list of plans we will continue showing a Loading indicator

kelvin-muchiri:
> can we please create an issue to address this? ^^

What issue is this?

moshthepitt:
> @moshthepitt What issue is this?

@kelvin-muchiri the fact that when we get an empty list of plans we will continue showing a Loading indicator

src/containers/pages/InterventionPlan/PlanDefinitionList/tests/__snapshots__/index.test.tsx.snap
kelvin-muchiri: @moshthepitt Just looked at the component and remembered we do not need to about worry structure of the components since that should be handled by those components in their module tests. We should only be concerned about the props.

src/components/forms/Search/index.tsx
moshthepitt: @kelvin-muchiri can we make this not optional? i.e. remove the `?`

src/helpers/utils.tsx
```diff
@@ -871,3 +873,11 @@ export const IndicatorThresholdItemPercentage = (item: number, decimalPoints?: n
 };
 
 export const reactSelectNoOptionsText = () => NO_OPTIONS;
+
+/**
+ * Get query params from URL
+ * @param {Location} location location object from props
```
moshthepitt, line 879: Seems this is repeated `location location`

moshthepitt: @kelvin-muchiri ^^

src/store/ducks/plans.ts
```diff
@@ -389,53 +388,62 @@ export interface PlanFilters {
...
   reason?: FIReasonType /** plan FI reason */;
   statusList?: string[] /** array of plan statuses */;
+  title?: string /** plan title */;
 }
 
 /** plansArrayBaseSelector select an array of all plans
  * @param state - the redux store
  */
-export const plansArrayBaseSelector = (planKey?: string) => (state: Registry): Plan[] =>
+export const plansArrayBaseSelector = (planKey?: string) => (state: Partial<Store>): Plan[] =>
```
moshthepitt, line 397: @kelvin-muchiri what led to use changing from `Registry` to `Partial<Store>` in this file?

src/components/forms/Search/index.tsx
moshthepitt: @kelvin-muchiri has this ticket been created?

kelvin-muchiri: @moshthepitt  Yes

src/store/ducks/plans.ts
kelvin-muchiri: @moshthepitt  `state` from `mapStateToProps` is not of type `Registry` and will raise lint warning if passed as an argument. Or which is the best way to solve this?

Naima-Bashir:
This has passed QA. Please note that only the following search functionalities have been tested: 
 - home> monitor > 
 - home > manage plans

PR #804, opened 2020-04-16
Fix merge issues

kahummer:
Use ona utils pr added the GA tracker module which should be loaded from js-tools. I removed it and removed FlexObjects that I didn't capture when fixing conflicts.


src/store/ducks/opensrp/locations/index.ts
```diff
@@ -48,7 +48,7 @@ export type LocationsActionTypes = FetchLocationsAction | RemoveLocationsAction
 
 /** interface to describe location state */
 export interface LocationsState {
-  locationsByPlanId: FlexObject<Location[]> | {};
+  locationsByPlanId: Dictionary<Location[]> | {};
```
moshthepitt, line 51: @kahummer what does this object look like?

kahummer: any key value pair object the FlexObject is pointing to '../../../../helpers/utils' instead of using the Dictionary equivalent added by js-tools ona@utils

moshthepitt: I mean this `Dictionary<Location[]>` - what data fits this type def?

kahummer: I think it's array with location data keyed by id
maybe 
```
{
'ed2b4b7c-3388-53d9-b9f6-6a19d1ffde1f': [
identifier: string,
name: string
]
}
```

moshthepitt: cool

PR #805, opened 2020-04-16
opensrp logout

kahummer:
fix: #776 
The fix is currently working awaiting tests and review

### DIFF
--- a/src/App/App.tsx
+++ b/src/App/App.tsx
@@ -17,12 +17,12 @@ import { Route, Switch } from 'react-router';
 import { Redirect } from 'react-router-dom';
 import { toast } from 'react-toastify';
 import { Col, Container, Row } from 'reactstrap';
+import Logout from '../components/Logout';
 import CustomConnectedAPICallBack from '../components/page/CustomCallback';
 import Loading from '../components/page/Loading';
 import {
   BACKEND_ACTIVE,
   DISABLE_LOGIN_PROTECTION,
-  EXPRESS_OAUTH_LOGOUT_URL,
   GA_CODE,
   GA_ENV,
   OPENSRP_LOGOUT_URL,
@@ -411,7 +411,10 @@ const App = () => {
                 component={() => {
                   if (BACKEND_ACTIVE) {
                     store.dispatch(logOutUser());
-                    window.location.href = EXPRESS_OAUTH_LOGOUT_URL;
+                    return (
+                      <Logout logoutURL="https://reveal-stage.smartregister.org/opensrp/logout.do" />
+                    );
+                    // window.location.href = EXPRESS_OAUTH_LOGOUT_URL;
                     return null;
                   }
                   const state = getOauthProviderState(store.getState());
--- a/src/App/App.tsx
+++ b/src/App/App.tsx
@@ -111,8 +111,9 @@ if (GA_CODE) {
 export interface AppState {
   username?: string;
 }
+/** Interface defining component props */
 interface AppProps {
-  logoutComponent?: (props: LogoutProps) => null;
+  logoutComponent: (props: LogoutProps) => null;
 }
 
 const APP_CALLBACK_URL = BACKEND_ACTIVE ? BACKEND_CALLBACK_URL : REACT_LOGIN_URL;
@@ -413,6 +414,7 @@ const App = (props: AppProps) => {
                 component={() => {
                   if (BACKEND_ACTIVE && props.logoutComponent) {
                     store.dispatch(logOutUser());
+                    /** returns logout component responsible for opensrp logot and moving execution to express server */
                     return <props.logoutComponent logoutURL={OPENSRP_LOGOUT_URL} />;
                   }
                   const state = getOauthProviderState(store.getState());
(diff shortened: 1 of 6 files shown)


src/App/App.tsx
```diff
@@ -411,7 +411,10 @@ const App = () => {
                 component={() => {
                   if (BACKEND_ACTIVE) {
                     store.dispatch(logOutUser());
-                    window.location.href = EXPRESS_OAUTH_LOGOUT_URL;
+                    return (
+                      <Logout logoutURL="https://reveal-stage.smartregister.org/opensrp/logout.do" />
```
cKellyDesign, line 415: This string feels magical, perhaps it should live in `constants.ts`?

cKellyDesign: Also let's utilize the URL bases from `.env` 🙏

src/components/Logout/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+export function logoutFromAuthServer(logoutURL: string) {
+  const logoutWindow: Window | null = window.open(logoutURL);
+  const timer: NodeJS.Timeout = setInterval(() => {
+    if (logoutWindow) {
+      logoutWindow.close();
+      window.location.href = EXPRESS_OAUTH_LOGOUT_URL;
+    }
+    clearInterval(timer);
+  }, 20);
+}
```
cKellyDesign, line 19: Doc strings?

cKellyDesign: Should this interval have a maximum number of tries? Or maybe like a timeout? (So it doesn't fire ever 20ms forever if something goes wrong?)

src/components/Logout/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+/** Logout component */
+const Logout = (props: LogoutProps) => {
+  logoutFromAuthServer(props.logoutURL);
+  return null;
+};
```
cKellyDesign, line 25: If we're doing this via a simple functional component, could / should this component return a `<Redirect />` component (instead of null) so we don't have to manually set `window.location.href`? 🤔

src/components/Logout/index.tsx
kahummer: @cKellyDesign I feel like we have no risk of this running on since we are opening and closing the window in a short time, then moving the execution to the server

src/components/Logout/index.tsx
kahummer: Cool will give this a shot

p-netm: @kahummer @cKellyDesign the logout action here needs to be served by express server. Using Redirect will limit the logout to the react-apps' scope and the express-server will never know of it.

kahummer: @p-netm I've actually confirmed this

p-netm: cool

src/App/App.tsx
```diff
@@ -111,6 +111,9 @@ if (GA_CODE) {
 export interface AppState {
   username?: string;
 }
+interface AppProps {
+  logoutComponent?: (props: LogoutProps) => null;
```
moshthepitt, line 115: - Does this need to be optional?
- Can we add docstrings to this file please?

p-netm: @kahummer I think the idea was having a defualt value defined on APP for `logoutComponent`. On the other hand and this is my solo opinion, we don't have to supply the logout component as a prop to `APP`. I would argue its not like we are calling `APP` in more than a single place such that we would need to customize its props.

src/components/Logout/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+ * @param {string} logoutURL URL string representing the auth server logout URL endpoint.
+ * This function takes the approach of opening a new window and navigating to the logout
+ * url of the authentication server in order to go around the browser's CORS policy.
+ */
+export function logoutFromAuthServer(logoutURL: string) {
+  const logoutWindow: Window | null = window.open(logoutURL);
+  const timer: NodeJS.Timeout = setInterval(() => {
+    if (logoutWindow) {
+      logoutWindow.close();
+      window.location.href = EXPRESS_OAUTH_LOGOUT_URL;
```
p-netm, line 19: would also be cool, to add a test for this line, I think we already have an example [here](https://github.com/onaio/reveal-frontend/blob/619a65e2b22742eabeb3481c7bff69ba685c92cf/src/components/page/CustomCallback/tests/index.test.tsx#L70)

src/App/App.tsx
kahummer: @p-netm I think we can have it as a default value. It eased mocking the component when running test related to logout on App

p-netm: ok

src/components/Logout/index.tsx
kahummer: @p-netm due to time constraints I added an issue to attend to this issue later https://github.com/onaio/reveal-frontend/issues/808

Naima-Bashir:
@p-netm this has passed QA.

PR #825, opened 2020-04-22
Fix IRS reporting legend

### LINKED ISSUE #820: The Conditional formatting rules percentage table shows double
Login to the system
Select the IRS reporting
Select one of the IRS plans
Observe the Conditional formatting rules percentage table.
Observe if it shows double data. It should not show two sets of data.

image

p-netm:
close #820 
Seems the error was a result of an erred rebase or merge.


Naima-Bashir:
Tested this. It works as expected.

PR #826, opened 2020-04-23
Use OpenSrpService from OpenSrp web for CRUD operations

### LINKED ISSUE #682: Use OpenSRP service from OpenSRP web
Use https://github.com/OpenSRP/opensrp-web/tree/master/packages/opensrp-server-service instead of the OpenSRP service class defined in this repo.

ciremusyoka:
Closes #682 
Start using `OpenSrpService` from `@opensrp/server-service`  library located at [OpenSrp Web](https://github.com/OpenSRP/opensrp-web/tree/master/packages/opensrp-server-service) to perform all `CRUD` operations.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "@onaio/session-reducer": "^0.0.10",
     "@onaio/superset-connector": "^0.0.13",
     "@onaio/utils": "^0.0.1",
+    "@opensrp/server-service": "^0.0.1",
     "@testing-library/react": "^10.0.1",
     "@types/bootstrap": "^4.2.1",
     "@types/fetch-mock": "^7.2.3",
--- a/src/services/opensrp/index.ts
+++ b/src/services/opensrp/index.ts
@@ -25,19 +25,6 @@ export interface URLParams {
   [key: string]: string | number | boolean;
 }
 
-/** params option type */
-type paramsType = URLParams | null;
-
-/** converts URL params object to string
- * @param {URLParams} obj - the object representing URL params
- * @returns {string} URL params as a string
- */
-export function getURLParams(obj: URLParams | {}): string {
-  return Object.entries(obj)
-    .map(([key, val]) => `${key}=${val}`)
-    .join('&');
-}
-
 /** converts filter params object to string
  * @param {URLParams} obj - the object representing filter params
  * @returns {string} filter params as a string
@@ -46,17 +33,6 @@ export function getFilterParams(obj: URLParams | {}): string {
   return Object.entries(obj)
     .map(([key, val]) => `${key}:${val}`)
     .join(',');
-}
-
-/** get payload for fetch
- * @param {HTTPMethod} method - the HTTP method
- * @returns the payload
- */
-export function getPayload(method: HTTPMethod) {
-  return {
-    headers: getDefaultHeaders() as HeadersInit,
-    method,
-  };
 }
 
 /** get payload for fetch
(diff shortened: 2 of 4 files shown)


src/services/opensrp/index.ts
```diff
@@ -58,6 +59,19 @@ export function getPayload(method: HTTPMethod) {
...
+ * @param {HTTPMethod} method - the HTTP method
+ * @returns the payload
+ */
+/* tslint:disable-next-line: no-unused-variable */
+export function getPayloadOptions(signal: AbortSignal, method: HTTPMethod) {
+  return {
+    headers: getDefaultHeaders() as HeadersInit,
+    method,
+  };
+}
```
cKellyDesign, line 73: I think we can remove the tslint disable and simple change unused variables to `_` like: 

```ts
getPayloadOptions(_: AbortSignal, method: HTTPMethod) { ... }
```

package.json
```diff
@@ -22,6 +22,7 @@
     "@onaio/session-reducer": "^0.0.10",
     "@onaio/superset-connector": "^0.0.13",
     "@onaio/utils": "^0.0.1",
+    "@opensrp/server-service": "^0.0.1",
```
cKellyDesign, line 25: ❤️

src/services/opensrp/index.ts
```diff
@@ -105,22 +119,8 @@ export class OpenSRPService {
...
-    if (!response.ok || response.status !== 201) {
-      throw new Error(
-        `OpenSRPService create on ${this.endpoint} failed, HTTP status ${response.status}`
-      );
-    }
-
-    return {};
+    const service = new OpenSRPServiceWeb(this.baseURL, this.endpoint, getPayloadOptions);
+    return await service.create(data, params, method);
   }
```
cKellyDesign, line 124: Nice! :D

ciremusyoka:
> @ciremusyoka looks good to me once we get tests passing 
> 
> (Looks like [c7ce32a](https://github.com/onaio/reveal-frontend/commit/c7ce32a60657e4cc47030f8280f46e0a07c9dbab) broke tests)

@cKellyDesign am out of options here because [src/services/opensrp/index.ts](https://github.com/onaio/reveal-frontend/blob/openSRP-service/src/services/opensrp/index.ts) is the only file i changed and it has a test coverage of `100%`.

p-netm:
@cKellyDesign please rereview.

- I think either we don't yet quite understand how coveralls interprets the jest-generated coverage reports or there is something wrong with the actual reports.

src/services/opensrp/tests/index.test.ts
```diff
@@ -26,20 +26,26 @@ describe('services/OpenSRP', () => {
...
+    const output = {
+      headers: {
+        accept: 'application/json',
+        authorization: 'Bearer hunter2',
+        'content-type': 'application/json;charset=UTF-8',
+      },
+      method: 'POST',
+    };
+    const signal = new AbortController().signal;
+    expect(getPayloadOptions(signal, 'POST')).toEqual({ ...output });
```
cKellyDesign, line 39: @ciremusyoka what happens to the report if you add another `expect()` here? (because before there were two?)

cKellyDesign: > I think either we don't yet quite understand how coveralls interprets the jest-generated coverage reports or there is something wrong with the actual reports

Agreed, we should do some research here to see how it's interpreted. Can you have over 100% coverage of a file?

ciremusyoka: @cKellyDesign one of the assertions was for the `getPayload` method which was being used before we started using `openSrpService` class. With the current implementation the method is no longer used so i removed it plus it's tests

p-netm: > Can you have over 100% coverage of a file?

:rofl: it would break some mathematical and physical laws. I however do not think that a drop coverage should have risen from the changes in this pr.

PR #846, opened 2020-05-07
890 refactoring tables

p-netm:
Refactor Table components in the following pages:
- planning pages,
- monitor pages.


p-netm:
image

p-netm:
image

p-netm:
image

p-netm:
image

src/containers/pages/FocusInvestigation/active/index.tsx
```diff
@@ -175,19 +196,122 @@ class ActiveFocusInvestigation extends React.Component<
...
+        Header: END_DATE,
+        accessor: 'plan_effective_period_end',
+      },
+    ];
+
+    const createTableProps = <D extends object>(
+      columns: Array<Column<D>>,
+      data: Plan[] | null,
+      queryParam: string
+    ) => {
```
cKellyDesign, line 294: @p-netm would it make sense to move this out of the render function so we don't re-declare it every render? Also, docstrings pls 🙏

src/helpers/utils.tsx
```diff
@@ -94,7 +94,7 @@ export interface GeoJSON {
 export function getLocationColumns(
   locations: LocationItem[] = locationHierarchy,
   padHeader: boolean = false
-): Column[] {
+) {
```
cKellyDesign, line 97: If not `Column[]`, what is being returned?  Can we add params and returns to the docstring?

p-netm: ok, this function is being used by 2 versions of React-table with different type declarations for Column, a union intersection will also not suffice as a solution here.

src/containers/pages/FocusInvestigation/active/index.tsx
p-netm: cool

src/helpers/utils.tsx
cKellyDesign: Ok great, that makes sense 👍

Issue #817, opened 2020-04-21
RVL-890 - Web UI Review from Thailand


craigappl:
@moshthepitt or @cKellyDesign , please complete this issue. We need to implement the changes recommended in the Web Ui Review from Thailand.

moshthepitt:
More: https://smartregister.atlassian.net/browse/RVL-890

moshthepitt:
@p-netm as noted in my last comment [here](https://smartregister.atlassian.net/browse/RVL-890), please note that we will be doing the search client side, while simultaneously sending an API request to the server with the search/filter parameters

p-netm:
closed by #911

PR #831, opened 2020-05-04
Web ui Thailand feedback

p-netm:
For #817 ([RVL-890](https://smartregister.atlassian.net/browse/RVL-890))
- Adds a button to customize the row Height
- Refactors searchinput filter
- Adds a table component with sort indicators as requested in mockup
- style tweaks for table layout

PR #850, opened 2020-05-11
{DNM} 890 web ui

p-netm:
base branch for changes leading to resolution of #817


moshthepitt:
@p-netm why do we need to use both v6 and v7?

p-netm:
> @p-netm why do we need to use both v6 and v7?

this was a safeguard incase I was not able to replace all drillDownTables.I will remove all react-table-v6 traces once I eventually refactor all the tables and make sure it will not be used in the codebase. 
@moshthepitt

moshthepitt:
Thanks @p-netm - could you please point me to where the drop down tables work was done?

p-netm:
https://github.com/onaio/reveal-frontend/pull/831

moshthepitt:
@p-netm is there also going to a corresponding PR here https://github.com/onaio/js-tools ?

p-netm:
> @p-netm is there also going to a corresponding PR here https://github.com/onaio/js-tools ?

@moshthepitt yes.

moshthepitt:
Cool.  Thanks.  It seems that we need to break down #831 and move its constituent parts around to where they more naturally belong.

PR #852, opened 2020-05-11
817 drill down table

p-netm:
- Adds a generic react-table based on the react-table package version v7:
 - can sort
 - can paginate
 - can change the row height.


moshthepitt:
@p-netm is this repo the best place to have this code live?

PR #858, opened 2020-05-18
webui version

### LINKED ISSUE #844: RVL-714 Add a version to Reveal Web UI homepage screen
We need to add the Web UI version number to the homepage so users know which version of the web UI they're using. This could be a footer,

https://smartregister.atlassian.net/browse/RVL-714

p-netm:
close #844 

Add a minimal footer section that shows the web apps version from package.json


p-netm:
Screenshot from 2020-05-18 11-01-00

src/App/App.tsx
```diff
@@ -136,300 +137,315 @@ const App = (props: AppProps) => {
...
+                {/* tslint:disable jsx-no-lambda */}
+                <Route
+                  exact={true}
+                  path={APP_LOGIN_URL}
+                  render={routeProps => (
+                    <OauthLogin
                       providers={providers}
-                      oAuthUserInfoGetter={oAuthUserInfoGetter}
+                      authorizationGrantType={AuthGrantType}
+                      OAuthLoginPromptMessage={LOGIN_PROMPT}
```
moshthepitt, line 391: What caused this change?

src/App/tests/App.test.tsx
```diff
@@ -62,7 +62,7 @@ describe('App', () => {
       </Provider>
     );
     // before resolving get oauth state request, the user is logged out
-    expect(wrapper.text()).toMatchInlineSnapshot(`"HomeLogin"`);
+    expect(wrapper.text()).toMatchInlineSnapshot(`"HomeLoginreveal-frontend: 0.4.2-rc7"`);
```
moshthepitt, line 65: Do you think it is worthwhile to have an additional snapshot test just for the footer/other sections?  I find the current test to be a bit strange i.e. this test just shows that this text appears on this page, right?

src/App/App.tsx
p-netm: non-intuitive breakpoint for the diff,

src/App/tests/App.test.tsx
p-netm: yeah, sure

src/App/App.tsx
p-netm: The bigger diff is around nesting the header and body sections, giving that a min-height so that it push the footer section to the bottom of the page.

moshthepitt:
@p-netm nice work.

Naima-Bashir:
nice. This has passed QA

Issue #848, opened 2020-05-11
There's a large search icon on the FI page, IRS plans page and Manage plans page


Naima-Bashir:
Login to the system
Go to the FI page: https://web.reveal-stage.smartregister.org/focus-investigation
There's a There's a large search icon on the FI page
![image](https://user-images.githubusercontent.com/61143692/81554124-e849a100-938e-11ea-8152-27c1b464de0d.png)
![image](https://user-images.githubusercontent.com/61143692/81556559-3496e000-9393-11ea-9c00-e58babb9f010.png)

Naima-Bashir:
![image](https://user-images.githubusercontent.com/61143692/81556499-103b0380-9393-11ea-8fa2-3d0839ecaf8a.png)
On the Manage plans as well

p-netm:
I have made a small change as a temporary fix for this, a permanent fix will be included in the next deployment.

moshthepitt:
> I have made a small change as a temporary fix for this, a permanent fix will be included in the next deployment.

@p-netm could you provide details on what constitutes the small temp fix?

p-netm:
> > I have made a small change as a temporary fix for this, a permanent fix will be included in the next deployment.
> 
> @p-netm could you provide details on what constitutes the small temp fix?

@moshthepitt I have removed the icon, I will be adding it back with updated styles in the coming deployment

p-netm:
closed by #862

PR #862, opened 2020-05-20
Filter plans by user

### LINKED ISSUE #840: RVL-952 - WebUI should allow filtering of plans by user
**Description**
In Thailand it is imperative that users of the WebUI can filter plans by user. This will be implemented using a drop down/auto complete control where the user can select a username from a list. Ideally, the filter should have a way for the user to easily click a 'me' button . This will become important as Thailand adds all staff as 'users' of the WebUI. It is expected that role based access will NOT be implemented, but filtering of plans by user (and ideally logged in user )

p-netm:
close #840 

Adds a filter section in manage plans page, from which selecting a user will filter out plans that are not accessible the respective user.


moshthepitt:
@p-netm once you filter by user, there is no indication on the page that a filter has been upload (even if you click on "filter" again).

Oo

Was this covered in the mockups?  If not please reach out to @rowo for guidance

moshthepitt:
@p-netm also noticed that if you press the search button before typing anything, you will end up with `?title=undefined` and an empty result page with no way to recover.  Can we prevent this?

src/helpers/dataLoading/plans.ts
```diff
@@ -0,0 +1,25 @@
+import { toast } from 'react-toastify';
+import { ActionCreator, AnyAction } from 'redux';
+import { OPENSRP_PLANS_BY_USER_FILTER } from '../../constants';
+import { OpenSRPService } from '../../services/opensrp';
+import store from '../../store';
+import { growl } from '../utils';
+
+/** find plans that the given OpenMRS userName has access to
+ * @param {string} userName - openMRSUserName
```
moshthepitt, line 9: We should not mention `openMRS` at all anywhere in my view.

p-netm: sawa

p-netm:
> Seems you cannot filter by user on the FI page http://localhost:3000/focus-investigation

@moshthepitt  will this be a applied globally or on a per table basis, i.e. since we have 2 tables in ActiveFIs page.

moshthepitt:
> @moshthepitt will this be a applied globally or on a per table basis, i.e. since we have 2 tables in ActiveFIs page.

@p-netm I think it was globally, but please check the mockups

p-netm:
@moshthepitt Please re-review

src/components/forms/PractitionerForm/UserIdSelect/index.tsx
```diff
@@ -91,28 +105,42 @@ export const UserIdSelect: React.FC<Props> = props => {
     return allOpenMRSUsers;
   };
 
-  /** filters out openMRs User objects that have already been mapped to an existing
-   * practitioner, this is an effort towards ensuring a 1-1 mapping between an openMRS user
-   * and a practitioner entity
+  /** depending on the value of showPracitioners; it filters out openMRs User objects that have
+   * already been mapped to an existing practitioner, this is an effort towards ensuring a 1-1
+   * mapping between an openMRS user and a practitioner entity
```
moshthepitt, line 110: Please let us avoid mentioning OpenMRS anywhere.

src/store/ducks/opensrp/planIdsByUser/index.ts
```diff
@@ -0,0 +1,157 @@
+import { Dictionary } from '@onaio/utils';
+import { AnyAction, Store } from 'redux';
+import { createSelector } from 'reselect';
+import SeamlessImmutable from 'seamless-immutable';
+import { PlanDefinition } from '../../../../configs/settings';
+
+/** the reducer name */
+export const reducerName = 'planIdsByUser';
```
moshthepitt, line 8: What's preventing us from using the already existing plans reducer module?

p-netm: the plans reducer module stores plans with the `PlanRecord` type, but the plans that we get from the filter endpoint are of `PlanDefinitionList` type and are handled by the opensrp/PlanDefinition reducer.

moshthepitt: Cool thanks, does that mean that we can use this reducer https://github.com/onaio/reveal-frontend/blob/89e8ef36c9a6f2fa036d598630bb6082ad3bfc56/src/store/ducks/opensrp/PlanDefinition/index.ts

moshthepitt:
> We should not allow submitting the empty search box to result in http://localhost:3000/focus-investigation?title=undefined or http://localhost:3000/plans/list?title=undefined

Do we have a test to prevent a regression on this?

p-netm:
> > We should not allow submitting the empty search box to result in http://localhost:3000/focus-investigation?title=undefined or http://localhost:3000/plans/list?title=undefined
> 
> Do we have a test to prevent a regression on this?

@moshthepitt I do not think we need one:
- There is no way to set the title value `undefined` from the frontend.
- Also there is just no feasible way to simulate such an action given that the input in the searchform is auto-consumed as soon as input starts, you either have some text or don't(an empty string).

p-netm:
I also think it would be best to remove the search feature form the admin module pages, and add them in another PR.

moshthepitt:
> I also think it would be best to remove the search feature form the admin module pages, and add them in another PR.

@p-netm I agree.  Please go ahead and do this.

p-netm:
close #848

Issue #869, opened 2020-05-25
RVL-526 - Restrict planning to only plan types available for the region


samkanga:
The planning menu at the top of the WebUI has drop down options of 'manage plans' and 'irs'.

Manage plans takes you to a list of all plans for IRS and FI. From here you can create a new plan for IRS or FI

IRS takes you to a list of draft IRS plans (cannot see active or retired plans). You can create a new plan here for IRS or FI.

There needs to be distinct purpose for each option in this drop down menu. I suggest:

- [x] Adding a FI drop down option under the plan Menu that shows only FI draft plans and allows you to add a FI plan. 
- [x] Remove the IRS menu under Plan for Thailand

**Acceptance Criteria**

1. When I go to Plan menu on the WebUI, I should see the three options. Manage Plans, IRS and Focus Investigation.
2. Selection of IRS plan should only allow viewing of draft IRS plans and creation of new IRS plans.
3. Selection of Focus Investigation should only allow viewing of draft FI plans and creation of new FI plans.
4. For Thailand, the IRS option under plan should not be available.

moshthepitt:
> When I go to Plan menu on the WebUI, I should see the three options. Manage Plans, IRS and Focus Investigation.
Selection of IRS plan should only allow viewing of draft IRS plans and creation of new IRS plans.
Selection of Focus Investigation should only allow viewing of draft FI plans and creation of new FI plans.

@samkanga we likely will not do it this way.

I would rather have one "Manage Plans" page, and when you view it you only see the plans that are enabled for you.

---

@ciremusyoka  ^^

moshthepitt:
Closed by #893

samkanga:
@moshthepitt I would like to re-open this ticket based on your last comment. Its OK to have a common manage page but there was additional requirement to have additional menu for Focused Investigation, similar to the IRS one under plan that would provide access to all the FI draft plans.

moshthepitt:
> @moshthepitt I would like to re-open this ticket based on your last comment. Its OK to have a common manage page but there was additional requirement to have additional menu for Focused Investigation, similar to the IRS one under plan that would provide access to all the FI draft plans and be able to create new FI plans.

@samkanga I don't agree with this.  For us to consider it further, could we please have a plan, with mockups, on how this would work with the looming dynamic plans.  

> Separately, could you confirm if the IRS option for the Thailand WebUI was removed?

We have not deployed to Thailand.

moshthepitt:
[...]

Issue #865, opened 2020-05-25
RVL-643 - Hide from Web UI all functionality that does not work - Thailand


samkanga:
For the Thailand implementation, hide from Web UI, functionality that does not work, i.e:

1. Ability to create IRS plans 
3. Reactive plan creation through Web UI 

**Acceptance Criteria:**

1. Ability to create IRS plans  on Thailand implementation removed. Go to Plans > Manage Plans > Add Plan > Intervention Type
3.  “Case Triggered” option under “Focus Investigation Reason” when creating new plans for the Thailand implementation, removed. Go to Plans > Manage Plans > Add Plan > Focus Investigation Reason

moshthepitt:
> Ability to create IRS plans

Looks like this will be done in https://github.com/onaio/reveal-frontend/issues/869

We'll try and get this done:  

- Reactive plan creation through Web UI

moshthepitt:
Closed by #893

PR #893, opened 2020-06-04
Manage Intervention types with env variable

ciremusyoka:
Adresses #869 and #865 

- Intervention types options on `Plans > Manage Plans > Add Plan > Intervention Type` configurable with an env variable.
- Limit list of plans on `assign`  and `plans >  Manage Plans` pages to enabled plan types.
- Limit search functionality to only enabled plan types.
-  FI reasons options on `Plans > Manage Plans > Add Plan > Focus Investigation Reason` configurable  with env variable


src/components/forms/PlanForm/index.tsx
```diff
@@ -163,6 +165,11 @@ const PlanForm = (props: PlanFormProps) => {
   const irsActivities = getFormActivities(IRSActivities);
   const fiActivities = getFormActivities(FIActivities);
 
+  let filteredFIReasons: FIReasonType[] = [...FIReasons];
+  if (LANGUAGE === 'th') {
+    filteredFIReasons = FIReasons.filter(reason => reason !== CASE_TRIGGERED);
+  }
+
```
moshthepitt, line 172: I would vote for using an env variable (something like enabled_fi_reasons) instead of relying on the language.  It may be possible that we have an English Thai deployment, for example.

ciremusyoka:
> We also need to change the list of plans in the manage plans page (https://web.reveal-stage.smartregister.org/plans/list) and the assign page (https://web.reveal-stage.smartregister.org/assign) to filter out plans that are not enabled.
> 

Noted

src/components/forms/PlanForm/index.tsx
```diff
@@ -303,8 +321,12 @@ const PlanForm = (props: PlanFormProps) => {
...
                 className={errors.interventionType ? 'form-control is-invalid' : 'form-control'}
               >
-                <option value={InterventionType.FI}>{FOCUS_INVESTIGATION}</option>
-                <option value={InterventionType.IRS}>{IRS_TITLE}</option>
+                {isPlanTypeEnabled(InterventionType.FI) && (
+                  <option value={InterventionType.FI}>{FOCUS_INVESTIGATION}</option>
+                )}
+                {isPlanTypeEnabled(InterventionType.IRS) && (
+                  <option value={InterventionType.IRS}>{IRS_TITLE}</option>
+                )}
```
moshthepitt, line 329: Do we need to cater for other plan types here?

src/components/forms/PlanForm/index.tsx
```diff
@@ -289,7 +305,9 @@ const PlanForm = (props: PlanFormProps) => {
                 component="select"
                 name="interventionType"
                 id="interventionType"
-                disabled={disabledFields.includes('interventionType')}
+                disabled={
+                  disabledFields.includes('interventionType') || ENABLED_PLAN_TYPES.length < 2
```
moshthepitt, line 309: Can we please remove `ENABLED_PLAN_TYPES.length < 2` for now?

src/components/forms/PlanForm/index.tsx
ciremusyoka: `MDA Point` plan type added

PR #894, opened 2020-06-04
Fix web log out bug

### LINKED ISSUE #755: RVL-944 - Web logging out bug
Noticed primarily on web thailand. Sometimes logging out freezes and before redirecting to https://web.reveal-th.smartregister.org/undefined.

p-netm:
close #755

### DIFF
--- a/src/App/App.tsx
+++ b/src/App/App.tsx
@@ -125,7 +125,7 @@ export interface AppState {
 }
 /** Interface defining component props */
 interface AppProps {
-  logoutComponent: (props: LogoutProps) => null;
+  logoutComponent: (props: LogoutProps) => JSX.Element;
 }
 
 const APP_CALLBACK_URL = BACKEND_ACTIVE ? BACKEND_CALLBACK_URL : REACT_LOGIN_URL;
--- a/src/App/App.tsx
+++ b/src/App/App.tsx
@@ -124,7 +124,7 @@ export interface AppState {
 }
 /** Interface defining component props */
 interface AppProps {
-  logoutComponent: (props: LogoutProps) => JSX.Element;
+  logoutComponent: (props: LogoutProps) => null;
 }
 
 const APP_CALLBACK_URL = BACKEND_ACTIVE ? BACKEND_CALLBACK_URL : REACT_LOGIN_URL;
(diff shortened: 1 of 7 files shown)


src/App/tests/App.test.tsx
```diff
@@ -18,6 +18,8 @@ jest.mock('../../configs/env');
 const fetch = require('jest-fetch-mock');
 const history = createBrowserHistory();
 
+// jest.useFakeTimers();
```
moshthepitt, line 21: Can we remove is not used?

src/components/Logout/index.tsx
```diff
@@ -12,14 +13,20 @@ export interface LogoutProps {
...
-      logoutWindow.close();
-      window.location.href = EXPRESS_OAUTH_LOGOUT_URL;
-    }
-    clearInterval(timer);
-  }, 20);
+  const logoutWindow = window.open(logoutURL);
+  /** I am not sure the optimum value for this, mozilla firefox seems to take some
+   * time loading up the logoutWindow and actually making the request.
+   */
+  const RedirectAfter = 1000;
```
moshthepitt, line 20: Should we make it configurable?  (it should at least be a constant)

p-netm: yeah

src/App/App.tsx
```diff
@@ -462,8 +461,7 @@ const App = (props: AppProps) => {
                   // tslint:disable-next-line: jsx-no-lambda
                   component={() => {
                     if (BACKEND_ACTIVE) {
-                      store.dispatch(logOutUser());
```
moshthepitt, line 465: Hmm, does removing this completely make things work better?

p-netm: I added some information at the time of making this change; on why I thought it should be made
> I believe the dispatch action causes any views dependent on the gatekeeper authenticated state to rerender, which is I think the whole app. this introduces an unforseen and uncotrolled rerender that was not initially accounted for when writing the logout component for authorization code flow. 

this rerender actually makes things a little harder to think about, but it seems when I removed the dispatch action, the logout component was able to execute to completion. 
The things that I am not quite certain of is how the gatekeeper module `authenticated` variable resets to false,.. My initial thought was that after logout; when `fetchState` is invoked while attempting to access any of the connected private routes, it finds no session info and rehydrates the store to reflect that.

I also think if that if we are going to treat this in the SSO line of thought, the opensrp server should be able to redirect the user to a predefined url, in this case the express logout url.

p-netm: @moshthepitt :point_up:

faith-mutua:
@p-netm users can now log out successfully from https://web.reveal-stage.smartregister.org/ on both Chrome and Firefox.

PR #907, opened 2020-06-15
Add country configurations for zambia

### LINKED ISSUE #908: Assign Plans Option on Zambia Production WebUI keeps loading
- [ ] When I select a plan for team assignment under the assign menu on the Zambia production instance, it keep loading and does not open the plan assignment page. See screenshot below:

image
 The browser does not return any error on the console

p-netm:
close #908

### DIFF
--- a/src/configs/settings.ts
+++ b/src/configs/settings.ts
@@ -1065,7 +1065,7 @@ export interface JurisdictionsByCountry extends ADMN0 {
   // list of tilesets used for administrative boundaries
   tilesets?: Tileset[];
 }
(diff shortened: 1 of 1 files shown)


moshthepitt:
Maybe this is better for the bounds https://gist.github.com/graydon/11198540#gistcomment-2702247

p-netm:
> Maybe this is better for the bounds https://gist.github.com/graydon/11198540#gistcomment-2702247

cool, rechecked and the bounds should be ok.

Issue #910, opened 2020-06-15
Fix css for modal in clients upload csv page


p-netm:
Actual:

![Screenshot from 2020-06-15 19-26-32](https://user-images.githubusercontent.com/28119869/84682365-3594d180-af3e-11ea-8a5d-57f0de68485a.png)

Expected:
TBD

moshthepitt:
fixed by #914

PR #914, opened 2020-06-16
Fix client modal style

ciremusyoka:
Addresses #910 
Fixes css for modal in clients csv upload page


Naima-Bashir:
This has passed QA.

Issue #913, opened 2020-06-16
Improve performance of IRS Plan Page


moshthepitt:
Right now we can potentially make hundreds or thousands of calls to the OpenSRP API when trying to [fetch assignments](https://github.com/onaio/reveal-frontend/blob/master/src/containers/pages/InterventionPlan/IRS/plan/index.tsx#L283).  It seems that we can make this better performing by replacing the many call with just one i.e. fetching all assignments for the plan using something like `assignedLocationsAndPlans?plan=identifier`.

moshthepitt:
Right now we are calling this endpoint https://reveal-zm.smartregister.org/opensrp/rest/organization/assignedLocationsAndPlans/23ee9e55-8824-44b1-a826-352dfe2d5c96?plan=f1c3aedf-2c7b-563c-b1e2-5f0af070002b which in Zambia results is hundreds of calls to OpenSRP.

Naima-Bashir:
This has passed QA on stage

moshthepitt:
Fixed by #929 and #983

Issue #892, opened 2020-06-03
RVL-157 - CSV import for risk files


samkanga:
This document provides details on how to upload the risk and structure count csv and how to use the uploaded information. It includes mock-ups for the risk profile  and the number of structure upload alongside the risk evaluation workflow.

Some points to note are:
- The MVP will focus on one risk profile
- The Upload of the risk profile and number of structures document will be manual for this step

Below is the desired workflow:

1. A user will be provided with a template pre-filled with jurisdiction IDs and names and with provision to capture the risk profile and the structure count. See sample template mock-up below: 
2.  ![image](https://user-images.githubusercontent.com/5908630/83618189-c230ae80-a592-11ea-8bf9-3131034ff3c8.png)
3. The user will fill in the risk profile (the MVP will provide one variable) and provide an option to provide the structure count. Where the structure count already exists in the system, the structure count uploaded will override what exists in the system.
4. Ona will facilitate the manual upload of the completed file into the system from the backend.
5. On the WebUI, the user will select the create plan button and create an IRS plan and have an option to move to the risk threshold page.
6. The page will load with a risk threshold slider that automatically calculates the number of structures that meet the set threshold. See mock-up below:
7. ![image](https://user-images.githubusercontent.com/5908630/83618265-e2f90400-a592-11ea-8bba-74e83ad7d18f.png)
8. On clicking on the “continue to next step” button, the user is provided with a map with the jurisdictions affected by the risk threshold selection in the earlier step. The map allows the user to drill down the maps until they get to the hierarchy level that is the immediate parent of the individual focus areas. Note, you cannot go back once the process has started, but you can start afresh. Selection of the risk check-box allows one to toggle on and off a risk layer allowing one to see the jurisdictions affected by the risk profile. The step further provides a breakdown of the number of structures in each of the jurisdiction. Each of the pages will also provide the total number of structures in the district. See mock-up below:
9. ![image](https://user-images.githubusercontent.com/5908630/83762267-08b00700-a680-11ea-93b6-be6d488390b7.png)
10. On clicking on the “continue to next step” button, the user gets the function allowing editing of the targeted jurisdictions based on the available resources. This is possible through the “change target status” button. It allows one to manually remove areas that may have been automatically selected based on the risk profile but that can not be supported by the available resources. The status change is made from targeted to Not targeted and vice versa. A note is also made on whether the status update was through auto-selection or by a user-change. See sample mock-up below:
[...]

Issue #788, opened 2020-04-03
IRS Plan Map Page Issues


moshthepitt:
- [x] IRS Plan-Jurisdiction Assignment Cleanup #537 
- [x] Load map progressively #286 
- [x] IRS Plan Page: Don't Load Jurisdictions From Superset #785 #404
- [x] Remove deprecated component lifecycle methods or convert to functional component #439 
- [x] IRS Plan Page Sometimes Does Not Load #786
- [x] IRS Plan Page Increase test coverage #669
- [x] IRS Assign Page Sometimes Does Not Load
- [x] IRS Plan Map Page DrillDown Back Button Issue #787

moshthepitt:
Addressed partly by #977 and #929

moshthepitt:
Finally closed by #1038

PR #916, opened 2020-06-17
Get plan assignments all at once

moshthepitt:
This commit changes how the assign page works but fetching all plan
assignments from OpenSRP in one API call instead of possibly hundreds/thousands.

I consider this a workaround to stabilize the assign page by making things more stable,
more work will be done in: https://github.com/onaio/reveal-frontend/issues/892
and https://github.com/onaio/reveal-frontend/issues/788

Fixes: https://github.com/onaio/reveal-frontend/issues/913


Naima-Bashir:
image
This has passed on stage. To be retested on zambia production

moshthepitt:
Overtaken by #929

PR #922, opened 2020-06-17
Jurisdiction metadata import

### LINKED ISSUE #909: CSV import of arbitrary jurisdiction metadata
We want to create a simple [Formik](https://jaredpalmer.com/formik/) form that takes a CSV file upload.

The CSV file will contain these standard fields:

- jurisdiction_id
- jurisdiction_name

Any other field that the CSV contains will be considered to be metadata for the jurisdiction that it is attached to.

On submitting the form, we will use [papaparse](https://www.papaparse.com/) to convert the CSV to JSON and then submit it using the [OpenSRP server service](https://github.com/OpenSRP/opensrp-web/tree/master/packages/opensrp-server-service) so the OpenSRP server settings API endpoint.

This is part of https://github.com/onaio/reveal-frontend/issues/892

gmwasi:
Resolves #909

Bulk upload of Jurisdiction Metadata Files and saving in the settings endpoint

- CSV headers take the format: jurisdiction_id, jurisdiction_id,<metadata_1>,< metadata_2>,...


src/components/forms/JurisdictionMetadata/index.tsx
```diff
@@ -0,0 +1,230 @@
...
+                value,
+              };
+              settings.push(setting);
+            }
+          }
+        }
+        const payload: SettingConfiguration = {
+          identifier: `jurisdiction_metadata-${header}`,
+          locationId: '',
+          providerId: 'demo',
```
moshthepitt, line 101: Let us use the currently logged in user for this.  There is a `getUser` selector from '@onaio/session-reducer';

src/components/forms/JurisdictionMetadata/index.tsx
```diff
@@ -0,0 +1,230 @@
...
+          settings,
+          teamId: '',
+          type: 'SettingConfiguration',
```
moshthepitt, line 104: Could we please move such magic strings to the constants module?

src/components/forms/JurisdictionMetadata/index.tsx
```diff
@@ -0,0 +1,230 @@
...
+import { growl } from '../../../helpers/utils';
+import { OpenSRPService } from '../../../services/opensrp';
+
+const SUPPORTED_FORMATS = ['text/csv'];
+
+/** yup validation schema for Jurisdiction Metadata Form input */
+export const JurisdictionSchema = Yup.object().shape({
+  file: Yup.mixed()
+    .required(REQUIRED)
+    .test('fileFormat', 'CSV Files only', value => value && SUPPORTED_FORMATS.includes(value.type)),
```
moshthepitt, line 31: Should this be come from the lang module?

src/components/forms/JurisdictionMetadata/tests/index.test.tsx
```diff
@@ -0,0 +1,32 @@
+import { render } from '@testing-library/react';
```
moshthepitt, line 1: I think we should add tests that explore how the form works:

- when showing errors
    - what errors are actually supposed to be shown to the user
- when submitting
    - is the payload built successfully based on what was uploaded
    - is the API endpoint called with the right params

src/configs/env.ts
```diff
@@ -24,6 +24,11 @@ export type GA_ENV = typeof GA_ENV;
 export const ENABLE_ASSIGN = process.env.REACT_APP_ENABLE_ASSIGN === 'true';
 export type ENABLE_ASSIGN = typeof ENABLE_ASSIGN;
 
+/** Do you want to enable Jurisdiction Metadata Upload */
+export const ENABLE_JURISDICTION_METADATA_UPLOAD =
+  process.env.REACT_APP_ENABLE_JURISDICTION_METADATA_UPLOAD === 'true';
```
moshthepitt, line 29: Can we add this to the sample.env file please?

src/components/page/Header/index.tsx
```diff
@@ -162,6 +165,18 @@ export class HeaderComponent extends React.Component<HeaderProps, State> {
                 </NavItem>
               )}
 
+              {ENABLE_JURISDICTION_METADATA_UPLOAD && (
```
moshthepitt, line 168: Cold we move this to fall under the `Admin` header group?

src/components/forms/JurisdictionMetadata/index.tsx
```diff
@@ -0,0 +1,230 @@
...
+  identifier: string;
+  providerId: string;
+  locationId: string;
+  settings: Setting[];
+  teamId: string;
+}
+
+export interface JurisdictionMetadataFormProps {
+  disabledFields: string[];
+  OpenSRPService: new (...args: any[]) => any;
```
moshthepitt, line 56: Why do we need `any` here?  I think we should use an actual type.  We already know the parameters that this class takes, and if all else fails we don't have to make the prop be an instantiated object, it could just be the class that is then instantiated later.

src/components/forms/JurisdictionMetadata/index.tsx
```diff
@@ -0,0 +1,230 @@
...
+  initialValues: JurisdictionMetadataFormFields;
+  redirectAfterAction: string;
+}
+
+export const defaultInitialValues: JurisdictionMetadataFormFields = {
+  file: new File([], ''),
+};
+
+// create payloads for sending
+const createPayloads = (result: any): SettingConfiguration[] => {
```
moshthepitt, line 73: Does this have to be any?

src/components/forms/JurisdictionMetadata/index.tsx
```diff
@@ -0,0 +1,230 @@
...
+        };
+        payloads.push(payload);
+      }
+    }
+  }
+  return payloads;
+};
+
+// read csv and convert to json
+const handleFile = (file: File, complete: (results: any) => any) => {
```
moshthepitt, line 114: Why any?

src/components/forms/JurisdictionMetadata/index.tsx
```diff
@@ -0,0 +1,230 @@
...
+                type: toast.TYPE.SUCCESS,
+              });
+            })
+            .catch((e: Error) => {
+              setGlobalError(e.message);
+            });
+        }
+        setSubmitting(false);
+      } else {
+        setGlobalError('Invalid CSV');
```
moshthepitt, line 152: This needs to go into the lang file

moshthepitt, line 152: Maybe we should consider using the `displayError` utility?

src/components/forms/JurisdictionMetadata/index.tsx
moshthepitt: I think that we should move `createPayloads` to something like a helpers.ts or utils.ts and then make it more general purpose of re-usable.  I think that its inputs do not necessary have to come from `JurisdictionMetadataForm` i.e. it is conceivable that some other component may have need to create a payload for OpenSRP settings, and we should make that possible.

moshthepitt:
@gmwasi could we please deal with the complaint from coveralls?

src/components/forms/JurisdictionMetadata/index.tsx
```diff
@@ -0,0 +1,230 @@
...
+const createPayloads = (result: any): SettingConfiguration[] => {
+  const payloads: SettingConfiguration[] = [];
+  const { data } = result;
+  // check if jurisdiction_id exists
+  if (data.length > 0 && data[0].jurisdiction_id) {
+    // get the metadata items
+    const headers = Object.keys(data[0]);
+    for (const header of headers) {
+      const settings: Setting[] = [];
+      if (header !== JURISDICTION_ID && header !== JURISDICTION_NAME) {
```
moshthepitt, line 82: Maybe we can instead use an array of headers to ignore that would then be a param that this function takes, with the default being [JURISDICTION_ID, JURISDICTION_NAME]

src/components/forms/JurisdictionMetadata/index.tsx
```diff
@@ -0,0 +1,230 @@
...
+        // add the metadata values with jurisdiction as the key
+        for (const item of data) {
+          const entries = Object.entries(item);
+          for (const [key, value] of entries) {
+            if (key === header) {
+              const setting: Setting = {
+                description: `${JURISDICTION_METADATA} for ${item.jurisdiction_name} id ${item.jurisdiction_id}`,
+                key: item.jurisdiction_id,
+                label: `${item.jurisdiction_name} metadata`,
```
moshthepitt, line 91: In case `jurisdiction_name` is not present, could we use `jurisdiction_id`?  I think there may be cases where the name is not provided or is empty.

src/components/forms/JurisdictionMetadata/index.tsx
gmwasi: Maybe not because this validation happens locally

moshthepitt: Okay.  In that case then we should probably unset setGlobalError when you try to resubmit a form so that the error is not stuck on the screen, yes?

src/components/forms/JurisdictionMetadata/index.tsx
gmwasi: We are not passing the parameters here, It's just a placeholder for the parameters in the super. Also I'm doing it here for ease of testing

src/containers/pages/JurisdictionMetadata/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+  };
+  const homePage = {
+    label: `${HOME}`,
+    url: `${HOME_URL}`,
+  };
+  breadcrumbProps.pages = [homePage];
+
+  /** props for the Jurisdiction Metadata form */
+  const jurisdictionMetadataFormProps: JurisdictionMetadataFormProps = {
+    OpenSRPService,
```
moshthepitt, line 53: I think that in this case we want to use `serviceClass` from props instead of `OpenSRPService` - right?

src/components/forms/JurisdictionMetadata/index.tsx
moshthepitt: I still don't agree with the weakened typing.  I mean, OpenSRP service is already fairly well typed - we are removing all the goodness that comes with that in this move.  Is there a problem with it being typed as `OpenSRPService: typeof OpenSRPService`?

moshthepitt: Can we please change the name of this prop to something like `serviceClass`.  Basically I prefer that we avoid re-using `OpenSRPService` - which is the name of the import at the top.  It makes for less easier to read code, and higher chances of using the wrong variable.

src/components/forms/JurisdictionMetadata/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+              onClose: () => setIfDoneHere(true),
+              type: toast.TYPE.SUCCESS,
+            });
+          })
+          .catch((e: Error) => {
+            setGlobalError(e.message);
+          });
+        setSubmitting(false);
+      } else {
+        setGlobalError(INVALID_CSV);
```
moshthepitt, line 92: what error


We need to unset the error once the user selects another CSV file and/or clicks submit.  This is so that the previous error message is not stuck on the screen, with the user not knowing whether it represents a new error or not.

src/components/forms/JurisdictionMetadata/index.tsx
gmwasi: I've unset it when a new file is selected

Issue #900, opened 2020-06-09
Add Web UI to Upload Configurable Forms


craigappl:
We need to create the following web UI pages for the following functionalities 

1.View all releases page #903 
2. View release related files #904
3. View JSON widget validators #905
4. Client form upload, edit #901 
5. Release making #902 

This [document](https://docs.google.com/document/d/1vRv3GpivpHQ8wmGDV_2PEmK5tiz4hvTeMkBX3WUSoNQ/edit?pli=1#) will help in case more information is required.

moshthepitt:
Fixed by #949

Naima-Bashir:
1.View all releases page #903    --passed QA
2. View release related files #904     --passed QA
3. View JSON widget validators #905    --passed QA
4. Client form upload, edit #901   --passed QA

PR #929, opened 2020-06-21
Improve plan assignments page

moshthepitt:
This PR attempts to improve the current plan assignments page by rewriting to be better performing, using the ideas in #916 and #925

Fix: #537
Fix: #870 
Fix: #882
Fix: #908 
Fix: #927
Fix: #916
Fix: #935 
Fix: #936
Fix: #937
Fix: #951
Fix: #952
Part of: #913
Invalidates, and thus closes: #529

### Approach

To load, the assignment page for a particular plan we need to fetch a number of things:

- Active organisations (aka teams) - this is loaded from the OpenSRP API
- Active assignments for the plan - this is loaded from the OpenSRP API
- The plan jurisdiction hierarchy - this is loaded from Superset but should be [loaded from the OpenSRP API](https://github.com/OpenSRP/opensrp-server-web/issues/396).  This is important to ensure that we do not try to traverse all jurisdictions, but just those that the plan cares about

#### Important Note

Loading from Superset is problematic because for new plans, the plan hierarchy may not yet be created in the Canopy database.  We should replace this ASAP.

If the any of the above fail to load, the best thing to do in my view is to show the user a helpful error message, and abort.  Or ask them to refresh the page.

#### Traversing the jurisdiction hierarchy

Once the above are loaded, we use the [TreeWalker component](https://github.com/onaio/reveal-frontend/pulls) to allow the user to drill down (and up) the plan's jurisdiction tree.  At each level the user can see current assignments for the jurisdiction, as well as edit them.

#### Saving assignments

When a user selects organisations to assign to a jurisdiction, an API call is made to OpenSRP to save the selected organisations.

#### Editing/removing existing assignments

This is not yet handled.  The [previous way of handling this](https://github.com/onaio/reveal-frontend/blob/871739f4a25b144d0beb31ed32733e85327ff579/src/containers/pages/InterventionPlan/IRS/plan/index.tsx#L2189) seems to be setting a start date far in the past and a stop date far in the future to "retire" assignments.  I felt this was too hacky for my liking and will be implementing it differently shortly.

#### Map component

This page does not currently load a map of the jurisdictions.  There was not enough time to do that.  Will be done shortly.

### New envs added

REACT_APP_SUPERSET_PLAN_HIERARCHY_SLICE=577  (for Stage)

### TODO

- [x] Handle editing/removing existing assignments
- [x] Improve error handling
- [x] Add tests
[...]

PR #949, opened 2020-06-24
Form config

ciremusyoka:
Use openSrp form config package to:
- Upload config form
- Edit config form
- Download uploaded config forms
- Publish manifest files

issue #900


src/configs/lang.ts
```diff
@@ -573,6 +573,36 @@ export const ALB_TABLETS_DISTRIBUTED = translate(
...
+export const FORM_REQUIRED_LABEL = translate('FORM_REQUIRED_LABEL', 'Form is required');
+export const FORM_NAME_REQUIRED_LABEL = translate(
+  'FORM_NAME_REQUIRED_LABEL',
+  'Form name is required'
+);
+export const FIND_RELEASE_LABEL = translate('FIND_RELEASE_LABEL', 'Find Releases');
+export const FIND_DRAFT_FILE_LABEL = translate('FIND_DRAFT_FILE_LABEL', 'Find Draft Files');
+export const FIND_RELEASE_FILES = translate('FIND_RELEASE_FILES', 'Find Release Files');
+export const FIND_VALIDATOR_FILES = translate('FIND_VALIDATOR_FILES', 'Find Validator Files');
+
```
moshthepitt, line 605: These need to be added to en.json

src/containers/pages/ConfigForm/manifest/filesList/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+ */
+const mapStateToProps = (
+  _: Partial<Store>,
+  ownProps: RouteComponentProps<RouteParams>
+): ManifestFilesProps => {
+  const formVersion = ownProps.match.params.id || '';
+  return {
+    formVersion,
+  };
+};
```
moshthepitt, line 103: What's the use of getting this formVersion in mapStateToProps?  If its coming from props, just get it in the component?  Ama there is some other reason?

ciremusyoka: @moshthepitt from my understanding `RouterParams` are passed only to the top level component  so it is either i pass the formVersion or i pass all router params. Just choose to pass the form version alone.

moshthepitt: I think the component has access to these router props.  See https://github.com/onaio/reveal-frontend/blob/72abd6d96d181d2b2c4d8d722fdf49355096322d/src/containers/pages/IRS/Map/index.tsx#L126 and https://github.com/onaio/reveal-frontend/pull/929/files#diff-0a00081136a35b6f7d1a2672959bfbedR67 but you may need to extend the props like this https://github.com/onaio/reveal-frontend/pull/929/files#diff-8825a081fc6f0bc30cb3fd21223bbceeR29

src/configs/lang.ts
ciremusyoka: done

p-netm:
@ciremusyoka can one pass tableProps to the containers through the exposed package interfaces, Or have you found a way to address the default no Data component look

ciremusyoka:
> @ciremusyoka can one pass tableProps to the containers through the exposed package interfaces, Or have you found a way to address the default no Data component look

@p-netm currently it is not possible but will address that on issue #958

Naima-Bashir:
Upload config form   --passed QA
Edit config form    --passed QA
Download uploaded config forms    --passed QA
Publish manifest files      --passed QA

Issue #967, opened 2020-06-26
Change the word APP into small letter


Naima-Bashir:
![image](https://user-images.githubusercontent.com/61143692/85853444-30941580-b7bb-11ea-8289-438b3d1ce4d4.png)

moshthepitt:
Closed by https://github.com/onaio/reveal-frontend/pull/971

Issue #958, opened 2020-06-25
Feedback on the new Form Configuration Web UI features


craigappl:
Craig spoke with Matt and recommended the following changes to the form config page:
1. Add a timestamp to the table on the manifest releases page so users have a sense of when these were updated
![image](https://user-images.githubusercontent.com/6067053/85649412-fcecaa80-b657-11ea-9ac7-fd134d71c615.png)
2. Add a timestamp for each file on the manifest releases page so people know when that file was last updated (Assuming they can be different. If they can't be different, maybe add it to the title next to the version number.)
![image](https://user-images.githubusercontent.com/6067053/85649493-24dc0e00-b658-11ea-9d3b-1250175adc9f.png)
3. Add an upload button to the draft files page The upload new file from the manifest page may need to be removed, we need to discuss why that button is there.
![image](https://user-images.githubusercontent.com/6067053/85649523-358c8400-b658-11ea-893d-d2788c091163.png)
4. Add a timestamp to the draft files page that shows when that file was uploaded.
![image](https://user-images.githubusercontent.com/6067053/85649573-5523ac80-b658-11ea-8134-ab1336c48990.png)
5. Change the name `manifest releases` to `releases`

dubdabasoduba:
> Add an upload button to the draft files page The upload new file from the manifest page may need to be removed, we need to discuss why that button is there.
@craigappl This button is here to enable users to easily upload any new files they are sure were not sent to the server in the initial files upload or any other upload.

dubdabasoduba:
> Add an upload button to the draft files page The upload new file from the manifest page may need to be removed, we need to discuss why that button is there.

@ciremusyoka You can go ahead and add the upload button as we discuss whether we still need it on the releases page

ciremusyoka:
@dubdabasoduba  @craigappl what is the decision on the `Upload New File` button in `Manifest Releases` page?

moshthepitt:
Closed by https://github.com/onaio/reveal-frontend/pull/971

Naima-Bashir:
1- Add a timestamp to the table on the manifest releases page so users have a sense of when these were updated --passed QA
2- Add a timestamp for each file on the manifest releases page so people know when that file was last updated (Assuming they can be different. If they can't be different, maybe add it to the title next to the version number.) --Passed QA
3- Add an upload button to the draft files page The upload new file from the manifest page may need to be removed, we need to discuss why that button is there.  --Passed QA
[...]

PR #971, opened 2020-06-30
Update form config tables

ciremusyoka:
Part of issue #958 and #967
Imports additional features of [form config](https://github.com/OpenSRP/opensrp-web/pull/262) i.e
- Time stamp columns on table data
- `Upload new file` button to drafts component
- Rename Table column header


Naima-Bashir:
1- Add a timestamp to the table on the manifest releases page so users have a sense of when these were updated --passed QA
2- Add a timestamp for each file on the manifest releases page so people know when that file was last updated (Assuming they can be different. If they can't be different, maybe add it to the title next to the version number.) --Passed QA
3- Add an upload button to the draft files page The upload new file from the manifest page may need to be removed, we need to discuss why that button is there.  --Passed QA
4- Add a timestamp to the draft files page that shows when that file was uploaded.  --Passed QA
5- Change the name manifest releases to releases  --Passed QA
6- Change the word APP into small letter --removed (Passed QA)

PR #985, opened 2020-07-03
AutoSelecting top node should cascade the selection to child nodes

p-netm:
Bug Description

Earlier autoselecting a node worked fine for just that one node, hence if you autoselected a node that wasn't a leaf node you would have a selected parent node with unselected child nodes, now autoselecting any node makes sure the selection cascades to the children of that node.


Naima-Bashir:
How to test: planning tool
compare with IRS plan, no map yet, just a table, Test the checkbox and auto selection

Naima-Bashir:
Tested this. It has passed QA.

PR #995, opened 2020-07-07
Jurisdictions assignment page jurisdiction count

### LINKED ISSUE #989: Show the number of currently selected jurisdictions on Jurisdiction Assignment Page
From the ui ideas [here](https://github.com/onaio/reveal-frontend/blob/master/src/containers/pages/PlanAssignment/helpers/AssignedOrgs/index.tsx), we should show the number of selected jurisdictions at any time

gmwasi:
- Added 'Jurisdictions Selected' column showing the count of jurisdictions selected and a tooltip showing their labels
resolves #989


src/containers/pages/JurisdictionAssignment/JurisdictionTable/tests/__snapshots__/index.test.tsx.snap
```diff
@@ -2,7 +2,7 @@
 
 exports[`src/containers/pages/jurisdictionView/jurisdictionTable selects and deselect nodes: should be now checked 1`] = `
 <input
-  checked={true}
+  checked={false}
```
moshthepitt, line 5: Any reason why these are changing?

src/containers/pages/JurisdictionAssignment/helpers/SelectedJurisdictionsCount/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+  tree.walk(node => {
+    selectedLeafNodes.forEach(leaf => {
+      if (leaf.model.id === node.model.id) {
+        nodesList.push(leaf);
+      }
+    });
+    return true;
+  });
+  return nodesList;
+};
```
moshthepitt, line 44: Is it possible to replace this with the selector? i.e. https://github.com/onaio/reveal-frontend/blob/master/src/store/ducks/opensrp/hierarchies/index.ts#L331

src/containers/pages/JurisdictionAssignment/JurisdictionTable/tests/__snapshots__/index.test.tsx.snap
```diff
@@ -29,15 +29,15 @@ exports[`src/containers/pages/jurisdictionView/jurisdictionTable selects and des
```
moshthepitt, line 29: Could we add a small snapshot that shows a the actual count rendered?

src/containers/pages/JurisdictionAssignment/helpers/SelectedJurisdictionsCount/index.tsx
gmwasi: I'm not sure it would work... i'm using https://github.com/onaio/reveal-frontend/blob/master/src/store/ducks/opensrp/hierarchies/index.ts#L331 and filtering out the items depending on the e drill down level

src/containers/pages/JurisdictionAssignment/JurisdictionTable/tests/__snapshots__/index.test.tsx.snap
```diff
@@ -48,3 +48,164 @@ exports[`src/containers/pages/jurisdictionView/jurisdictionTable works correctly
   Akros_1
 </span>
 `;
+
+exports[`src/containers/pages/jurisdictionView/jurisdictionTable works correctly through a full render cycle: structure count rendered 1`] = `
```
moshthepitt, line 52: This snapshot is too large for merely showing us that we get a structure count.  Please consult: https://kentcdodds.com/blog/effective-snapshot-testing

Idea: maybe just have a snapshot of the text, or maybe just assert the props match some expected value

src/containers/pages/JurisdictionAssignment/JurisdictionTable/tests/index.test.tsx
```diff
@@ -130,13 +130,32 @@ describe('src/containers/pages/jurisdictionView/jurisdictionTable', () => {
...
+    expect(toJson(tbodyRow.find('SelectedJurisdictionsCount'))).toMatchSnapshot(
+      'structure count rendered'
+    );
+
+    expect(
+      tbodyRow
+        .find('SelectedJurisdictionsCount')
+        .at(0)
+        .text()
+        .includes('')
```
moshthepitt, line 142: maybe we should use data that does not result in an empty string here?

src/containers/pages/JurisdictionAssignment/helpers/SelectedJurisdictionsCount/index.tsx
```diff
@@ -0,0 +1,84 @@
+import React, { Fragment, useState } from 'react';
+import { Tooltip } from 'reactstrap';
+import { TreeNode } from '../../../../../store/ducks/opensrp/hierarchies/types';
+
+/** Props for SelectedJurisdictionsCount  */
+export interface SelectedJurisdictionsCountProps {
+  id: string;
+  jurisdictions: TreeNode[] /** array of jurisdictions */;
+  parentNode: TreeNode | undefined;
```
moshthepitt, line 9: this is equivalent to `parentNode?` - why not do that?  Or `TreeNode | null`?

src/containers/pages/JurisdictionAssignment/helpers/SelectedJurisdictionsCount/index.tsx
moshthepitt: in that case then this function needs to be a selector function in the redux module no?  My reasoning is that 1) the 
purpose of selectors is to get/derive/calculate data from the store and 2) if we have a [memoized selector](https://github.com/reduxjs/reselect) then we improve performance - we might need that because your function here is traversing the whole tree for each parent's descendants.

src/containers/pages/JurisdictionAssignment/JurisdictionTable/index.tsx
```diff
@@ -204,7 +204,7 @@ const JurisdictionTable = (props: JurisdictionSelectorTableProps) => {
       />,
       <NodeCell key={`${node.model.id}-jurisdiction`} node={node} baseUrl={baseUrl} />,
       node.model.node.attributes.structureCount,
-      <SelectedJurisdictionsCount
+      <ConnectedSelectedJurisdictionsCount
```
moshthepitt, line 207: Just curious why you chose to connect `SelectedJurisdictionsCount` to the redux store?  It seems like its parent component is already connected and could have used the selector and fed it into `SelectedJurisdictionsCount` as a prop.

src/containers/pages/JurisdictionAssignment/helpers/SelectedJurisdictionsCount/index.tsx
gmwasi: With the selector implementation it has to be define that way

src/containers/pages/JurisdictionAssignment/JurisdictionTable/index.tsx
gmwasi: I figured it was cleaner to pass the filters this way since the result if the `selectedLeafNodes` as one of the filters to pass to the `getSelectedNodesUnderParentNode`

Naima-Bashir:
image
Tested this. When one selects a plan eg IRS and clicks on a certain jurisdiction, it shows the number of Selected Jurisdictions.
However, when no jurisdiction is selected, the column is blank. Perhaps as an improvement, we can have 0 as the initial number. 
What do you think? @p-netm @samkanga 
https://github.com/onaio/reveal-frontend/issues/1036

PR #1010, opened 2020-07-10
Add Error Boundary

ciremusyoka:


ciremusyoka:
@Naima-Bashir currently when the app encounters an error (this are errors that makes the app to crash) a blank page is displayed.
Adding Error Boundary enables us to render a specific page indicating to the user an error occurred.
Sample screenshot
Screenshot from 2020-07-10 15-36-20

Naima-Bashir:
@moshthepitt @ciremusyoka can you give me replicatable steps so that I can test it? For example, are there known areas the error is seen.
I cannot test this since I don't think we have any known page that breaks the app.

PR #1018, opened 2020-07-10
Upgrade @opensrp/population-characteristics package

ciremusyoka:
Upgrades @opensrp/population-characteristics package to the latest version


Naima-Bashir:
image

I've tested this. The UI does not look like the rest of the pages. Here's the issue I raised: https://github.com/onaio/reveal-frontend/issues/1035

Issue #930, opened 2020-06-22
RVL-1002 - Active FI Map Issues


p-netm:
- Tracking issues around using the map in the active FI map page
  - Fix map jankiness(probably because it gets mounted with each of the component's rerenders.)
      - takes a while to show up, 
      - rerenders the whole map when props change
  - Render a tool tip upon click of the index case point, include: Case number, Case Classification, and Date of Diagnosis

moshthepitt:
This is blocked because it needs Case number, Case Classification, and Date of Diagnosis added to the database view that the web UI uses.

moshthepitt:
Closed by #1096

PR #1021, opened 2020-07-10
Fix map jankiness

kelvin-muchiri:
Partially address #930 

Reduce map re-render by wrapping `GisidaLite` with HOC `React.memo` and passing a custom function as the second argument. Moving `ReactMapboxGl` initialization out of the function component, so that we have only one instance instead of recreating a new instance when props change resulted in the error `There is already a source with this ID`. This was solved by using `item.key` instead of `index` when rendering the `React.fragment`. The issue seemed resolved but some layers were being rendered incorrectly, for instance, structures that were to be rendered as yellow were now being rendered green.


kelvin-muchiri:
The map jankiness is resolved, but the fill layers are not rendering correctly. This needs to be resolved

moshthepitt:
> The map jankiness is resolved, but the fill layers are not rendering correctly. This needs to be resolved

@kelvin-muchiri please address that in a new issue/PR

PR #1029, opened 2020-07-13
Prevent infinite loading

### LINKED ISSUE #890: Loader  showing even when API does not return any files
The loader is being rendered in all cases when the store does not return the expected response. The loader should be rendered when waiting for data from the API and not all times that the data is missing.

ciremusyoka:
Closes #890
Disables loader when API doesn't return data or store doesn't have data for `Admin > Clients` page.

### DIFF
--- a/src/containers/pages/MDAPoint/ClientListView/helpers/serviceHooks.ts
+++ b/src/containers/pages/MDAPoint/ClientListView/helpers/serviceHooks.ts
@@ -12,12 +12,19 @@ import store from '../../../../../store';
 import { fetchFiles, File } from '../../../../../store/ducks/opensrp/clientfiles';
 import { getAccessToken } from '../../../../../store/selectors';
 
+/** useState function type for setting loader on and off */
+type SetIsLoading = ((status: boolean) => void) | null;
+
 /** loads and persists to store files data from upload/history endpoint
  */
 export const loadFiles = async (
+  setLoading: SetIsLoading = null,
   fetchFileAction = fetchFiles,
   serviceClass: any = OpenSRPService
 ) => {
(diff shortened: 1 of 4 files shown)


src/containers/pages/MDAPoint/ClientListView/helpers/serviceHooks.ts
```diff
@@ -12,12 +12,19 @@ import store from '../../../../../store';
 import { fetchFiles, File } from '../../../../../store/ducks/opensrp/clientfiles';
 import { getAccessToken } from '../../../../../store/selectors';
 
+/** useState function type for setting loader on and off */
+type SetIsLoading = ((status: boolean) => void) | null;
+
 /** loads and persists to store files data from upload/history endpoint
  */
 export const loadFiles = async (
+  setLoading: SetIsLoading = null,
```
p-netm, line 21: @ciremusyoka lets make this totally optional, in the sense that we can call loadFiles and not need to have to pass this.

src/containers/pages/MDAPoint/ClientListView/index.tsx
```diff
@@ -80,25 +81,31 @@ export const buildListViewData: (rowData: File[]) => ReactNode[][] | undefined =
 
 export const ClientListView = (props: ClientListViewProps & RouteComponentProps) => {
   const { location, files, clientLabel } = props;
+
+  const [loading, setLoading] = useState(false);
+
+  // update loading status
+  const isLoading = (status: boolean) => setLoading(status);
```
p-netm, line 88: Just curious here; why not pass `setLoading` itself as the `isLoading` param, without having to wrap it in another function

PR #1032, opened 2020-07-14
Fix intermittent errors when plan Jurisdiction page is loading

### LINKED ISSUE #1015: Plan Assignment Page Intermittent Error
Both the plan jurisdiction assignment and the plan team assignment pages show an error message briefly while loading.  We need to ensure that the error message is only shown when a definite error has occurred

- [x] Fix for plan jurisdiction assignment
- [ ] Fix for plan team assignment

p-netm:
fix #1015 

introduces a hook that one can use to track the loading state of a component when:
  - promises are spread across hooks or code sections where their loading state cannot be tracked in a unified way like when using `promise.all`


src/helpers/useLoadingReducer.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+  /** sets loading to true for loading sequence with the  passed in key
+   * @param key - target this load entry
+   * @param condition - by default start loading will start the loading sequence to true, this overrides that
+   *
+   * @return {string} the key
+   */
+  const startLoading = (key: string, condition: boolean = true) => {
+    return changeLoading(key, condition);
+  };
```
ciremusyoka, line 86: should we allow passing of `condition` here. I think passing false here can mess this.

p-netm: yeah that's the intention, so you can say something like `startLoading('plansData', !plansArray.length)`. as a replacement of

`setLoading(!plansArray.length)`

ciremusyoka: nice

PR #1046, opened 2020-07-16
Child Level  MDA Point  reporting

### LINKED ISSUE #979: Add child level view to the MDA Point Reporting
We need to add one last reporting view to the current MDA Point reporting. This layer will be the lowest level aka you will view it by clicking through the school view. 

Each row will represent an individual child and the columns will reflect the information below: 

**Column Headers:** 
Child's name
National ID
Enrolled in school --> (yes or no)
MMA drug administered --> (yes or no)
SAC refused --> (yes or no)
SAC sick/pregnant/contraindicated --> (yes or no)
ADR reported --> (yes or no)
PZQ tablets distributed --> (total number)
Alb tablets distributed -- (total number) 

[Data Dictionary found here](https://docs.google.com/spreadsheets/d/14NyA7b67arS_oy0uG3NOY0KcaTjK36kkobGRi02A0xk/edit#gid=1336356138)

**UI:** 
Because the view before this view has rows for children split into age categories instead of just a row for the school, we need a solution for how to click through to the full list of children (see image below). What we will do is add a row to the bottom of the table on the school view (view below) that has clickable text in the first column that says "See all children" - when they click it, the user will be taken to the view described above.  

Screen Shot 2020-07-01 at 9 28 14 PM

gmwasi:
Adds the lowest level view in MDA point reporting which shows individual rows 
resolves #979


src/containers/pages/MDAPoint/ChildReports/ChildSupersetDataTable/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+    loadData().catch(error => displayError(error));
+  }, []);
+
+  const listViewProps = {
+    data,
+    headerItems,
+    tableClass: tableClass || 'table table-striped table-bordered plans-list',
+  };
+
+  if (loading === true) {
```
p-netm, line 50: nitpick: `if (loading){}`

src/containers/pages/MDAPoint/ChildReports/ChildSupersetDataTable/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  /** async function to load the data */
+  async function loadData() {
+    try {
+      setLoading(data.length < 1);
+      await service(supersetSliceId).then((result: ChildReport[]) => {
+        fetchItems(result);
+      });
+    } catch (e) {
+      // handle error
```
p-netm, line 34: Lets do something with this error.

src/containers/pages/MDAPoint/ChildReports/index.tsx
```diff
@@ -0,0 +1,221 @@
...
+  data: React.ReactNode[][];
+  pageUrl: string;
+  pageTitle: string | null;
+  prevPage: Page[];
+}
+
+/** map state to props */
+const mapStateToProps = (
+  state: Partial<Store>,
+  ownProps: RouteComponentProps<RouteParams>
```
p-netm, line 163: lets use the full type even if we are only interested in the routeComponent props

src/store/ducks/generic/MDAChildReport.ts
```diff
@@ -0,0 +1,167 @@
...
+ */
+export const FetchMDAPointChildReportAction = (
+  childReportList: ChildReport[] = []
+): FetchMDAPointChildReport => {
+  const childReportByPlanId = {};
+  childReportList.forEach((report: ChildReport) => {
+    (childReportByPlanId as any)[report.plan_id]
+      ? (childReportByPlanId as any)[report.plan_id].push(report)
+      : ((childReportByPlanId as any)[report.plan_id] = [report]);
+  });
```
p-netm, line 58: lets refactor the use of any here, seems we already have a type for `childReportByPlanId`, ` { [key: string]: ChildReport[] }`

PR #1048, opened 2020-07-16
Add new/updated plans to store

### LINKED ISSUE #1020: Optimistically store new and updated plans in Redux store
When a plan is updated or created from:

- The manage plans page
- The planing tool
- The plan team assignment page

Then this plan should be optimistically added to the Plan Definition redux store i.e. the Redux store used for [plan definition list pages](https://github.com/onaio/reveal-frontend/issues/1019).

### Suggested implementation plan

- Wait object payload is sent to OpenSRP and we receive a success status
- Call a reusable function (or something like it) that adds the new/updated plan definition object to the redux store

This is so that the Plan Definition redux store always contains the most up to date versions of plans.  We are optimistically adding the new/updated plans to the redux store before we retrieve them from OpenSRP but we are fairly certain that they were successfully saved to OpenSRP.

kelvin-muchiri:
Closes #1020 

Add a new plan/updated plan to the redux store once it's successfully submitted to opensrp


src/components/forms/PlanForm/index.tsx
```diff
@@ -244,6 +248,15 @@ const PlanForm = (props: PlanFormProps) => {
     return <Redirect to={redirectAfterAction} />;
   }
 
+  const onSubmitSuccess = (setSubmitting: any, payload: PlanDefinition) => {
```
p-netm, line 251: @kelvin-muchiri  can we move this to a util file. change any to a setState type its usually sth like `React.Dispatch<React.setStateAction<boolean>>`, ooh and a docstring would be nice

src/containers/forms/ConnectedPlanForm/index.tsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
```
p-netm, line 1: why not use `store.dispatch` in the `onSubmitSuccess` callback as an alternative to having to connect the form.

kelvin-muchiri: @p-netm Ok....What are the pros of doing it like that over connect?

kelvin-muchiri: @p-netm https://stackoverflow.com/a/41107592/3247914

kelvin-muchiri:
I have managed to add a plan to store for a new/updated plan. As in the issue description, I am yet to do for 

- The planing tool
- The plan team assignment page

Wanted to add more tests for `ConnectedPlanForm` to test that `addPlan` prop is called when submit is successful, but had trouble with that test, so I removed it

kelvin-muchiri:
I have managed to [add the tests](https://github.com/onaio/reveal-frontend/pull/1048/commits/a4caab756111b24b9aae8d17b2c955abb94a3d7c) that I had trouble adding

src/containers/forms/ConnectedPlanForm/index.tsx
moshthepitt: @kelvin-muchiri with respect to https://stackoverflow.com/a/41107592/3247914 if PlanForm is called from an already redux-connected component (which is the case 100% of the time) then it already has access to a dispatch function provided via `mapDispatchToProps` of the component that calls PlanForm, no?  If so I am uncertain whether we need `ConnectedPlanForm`

kelvin-muchiri:
@moshthepitt Ok...In that case `addPlanDefinition` action will be dispatched by the parent component. Does that mean  for instance UpdatePlan and NewPlan will each implement its own prop function to handle dispatching `addPlanDefinition` action?

moshthepitt:
@kelvin-muchiri Lets say the parent component is called `UpdatePlan`.  We should have this component receive `addPlanDefinition` via `mapDispatchToProps`.  And then pass it to PlanForm as a prop.  `PlanForm` should then call this function when the form is successfully submitted.  In all cases, the parent component only received the function via `mapDispatchToProps` and then passes it to `PlanForm`.  Makes sense?

kelvin-muchiri:
@moshthepitt Yea..now I see, sure...

kelvin-muchiri:
@moshthepitt If there some components that are not connected and using `PlanForm` such as `BaseNewPlan`, should we connect them?

moshthepitt:
@kelvin-muchiri I think that chances are `BaseNewPlan` by itself is not used directly, but used inside other components that are connected.

@p-netm thoughts?

p-netm:
> @kelvin-muchiri I think that chances are `BaseNewPlan` by itself is not used directly, but used inside other components that are connected.
> 
> @p-netm thoughts?

The create/update Plan Pages are actually not connected; for instance `BaseNewPlan` isn't connected. 

 I however do not think we necessarily have to connect a component to be able to dispatch an action.  I do not think it would lead to coupling the action creators to the store singleton, because this would be a callback(`onSubmitSuccess`) defined uniquely according to the functionality of a respective view that is rendering the planform. That being said @kelvin-muchiri This is just matter of preference btw. I have no strong feelings against connecting these components if it works.

kelvin-muchiri:
@p-netm You are suggesting we use `stored.dispatch` instead of connecting a component?

p-netm:
> @p-netm You are suggesting we use `stored.dispatch` instead of connecting a component?

@kelvin-muchiri lets use connect, as a wrapper for the HOC that makes use of the planform

kelvin-muchiri:
@p-netm Ok...I'll do the same for the other components. Just like https://github.com/onaio/reveal-frontend/pull/1048/commits/712e7ccb201204b9be2bb1bf0fc70a86d1db597a

kelvin-muchiri:
@moshthepitt @p-netm  Please re-review

PR #1057, opened 2020-07-21
Disallow large snapshots

### LINKED ISSUE #1025: Add eslint rule for no large snapshots
Add [disallow large snapshots](https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/no-large-snapshots.md)

Might need tweaking for max allowed size.  Please discuss with the team.

kelvin-muchiri:
Closes #1025

Add eslint rule to disallow large snapshots. The largest snapshot file has a size of 3225. This is the value that has been set as the limit until the large snapshots are reduced below the ideal value we would like to set as the limit


kelvin-muchiri:
@p-netm @ciremusyoka @moshthepitt I was thinking of setting the max size of a snapshot as 1000 but the large snapshot is 3225. So I presume for now shall have it as 3225 until the large snapshot is reduced.

To test this functionality run `yarn lint-snap` while on this branch

p-netm:
> @p-netm @ciremusyoka @moshthepitt I was thinking of setting the max size of a snapshot as 1000 but the large snapshot is 3225. So I presume for now shall have it as 3225 until the large snapshots are reduced to sizes below the limit we want
> 
> To test this functionality run `yarn lint-snap` while on this branch

@kelvin-muchiri  a 1000 is magnitudes larger than what IMO we want to set as the max size. I think a good figure for external snapshots, i.e. snapshots inside .snap files should be about ~100 ~150 at most. while for an inline snapshot it should not be larger than 5 lines. 

also:
 - It would really be nice if the lint was applied during commit see [lint-staged](https://github.com/onaio/reveal-frontend/blob/7ed485c2b0d44e4634ffc3642cc7795d4e4eb9da/package.json#L122) 
 - if there is a lot work involved in reducing the snapshot size you can [whitelist the snapshots](https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/no-large-snapshots.md#options), and create an issue to reduce their sizes.

kelvin-muchiri:
@p-netm There many snap files larger than 150? The whitelist will contain many files

kelvin-muchiri:
@p-netm I tried adding the precommit hook in lint-staged but it did not work, do you know the exact command to add?

kelvin-muchiri:
@p-netm With a limit of 150, we are going to whitelist 11 snapshots should we proceed with the whitelisting?

kelvin-muchiri:
On my part I think 300 is not bad. I find 150 too low

kelvin-muchiri:
I removed the `inlineMaxSize` rule because it results in the error
Screenshot from 2020-07-24 18-14-24

p-netm:
> @p-netm I tried adding the precommit hook in lint-staged but it did not work, do you know the exact command to add?

I think something along the lines of 
```
"*.{snap, test.tsx}": [
      "eslint"
    ]
```

p-netm:
> @p-netm With a limit of 150, we are going to whitelist 11 snapshots should we proceed with the whitelisting?

yes, please. if its not too much trouble

p-netm:
> On my part I think 300 is not bad. I find 150 too low

Am curious what @ciremusyoka @KipSigei @gmwasi @moshthepitt  think!

kelvin-muchiri:
> I think something along the lines of
I tried this but it doesn't seem to work. Once can still proceed to commit

kelvin-muchiri:
@p-netm Added the snapshots above 150 to the whitelist

p-netm:
@kelvin-muchiri 
```
    "*.{snap, test.tsx}":[
      "eslint"
    ],
```

kelvin-muchiri:
@p-netm Added

PR #1063, opened 2020-07-22
Optional plan assignment map

### LINKED ISSUE #1055: PlanAssignment map should be optional
The PlanAssignment component the `ConnectedAssignmentMapWrapper` should be rendered optionally.

- [x] Add an env var of comma-separated plan intervention types which will have the map displayed
- [x] Don't display the map if current plan intervention type not in env var above
- [x] The default case is for all plan intervention types to show the map

gmwasi:
resolves #1055


src/configs/env.ts
```diff
@@ -359,3 +359,10 @@ export type JURISDICTION_METADATA_RISK = typeof JURISDICTION_METADATA_RISK;
 export const JURISDICTION_METADATA_RISK_PERCENTAGE =
   process.env.REACT_APP_JURISDICTION_METADATA_RISK_PERCENTAGE || '90';
 export type JURISDICTION_METADATA_RISK_PERCENTAGE = typeof JURISDICTION_METADATA_RISK_PERCENTAGE;
+
+/** plan types with map enabled during plan assignment */
+export const MAP_ENABLED_PLAN_TYPES = String(
+  process.env.REACT_APP_MAP_ENABLED_PLAN_TYPES ||
```
moshthepitt, line 365: I think that we should switch to using a blacklist i.e. disabled plan types.  This is so that when we add new plan types we don't have to remember to change this

.env.sample
```diff
@@ -111,5 +111,5 @@ REACT_APP_HIDDEN_MAP_LEGEND_ITEMS='Partially Sprayed'
 REACT_APP_JURISDICTION_METADATA_RISK=jurisdiction_metadata-risk
 REACT_APP_JURISDICTION_METADATA_RISK_PERCENTAGE=90
 
-# plan types with map enabled during plan assignment
-REACT_APP_MAP_ENABLED_PLAN_TYPES=FI,IRS,MDA,MDA-Point,Dynamic-FI,Dynamic-IRS,Dynamic-MDA
+# plan types with map disabled during plan assignment
+REACT_APP_MAP_DISABLED_PLAN_TYPES=FI,IRS,MDA,MDA-Point,Dynamic-FI,Dynamic-IRS,Dynamic-MDA
```
moshthepitt, line 117: Lets just have `MDA-Point` as default

src/containers/pages/PlanAssignment/index.tsx
```diff
@@ -248,18 +248,18 @@ const PlanAssignment = (props: PlanAssignmentProps) => {
...
       type ? (interventionType = type.valueCodableConcept) : (interventionType = '');
     }
-    return MAP_ENABLED_PLAN_TYPES.includes(interventionType);
+    return MAP_DISABLED_PLAN_TYPES.includes(interventionType);
   };
 
   return (
     <>
-      {isMapEnabled ? <ConnectedAssignmentMapWrapper {...AssignmentWraperProps} /> : ''}
+      {isMapDisabled ? '' : <ConnectedAssignmentMapWrapper {...AssignmentWraperProps} />}
```
moshthepitt: Should we maybe change this to `!isMapDisabled && <ConnectedAssignmentMapWrapper {...AssignmentWraperProps} />`?  This is so that we avoid the weird empty string output

src/containers/pages/PlanAssignment/index.tsx
```diff
@@ -245,7 +259,7 @@ const PlanAssignment = (props: PlanAssignmentProps) => {
 
   return (
     <>
-      <ConnectedAssignmentMapWrapper {...AssignmentWraperProps} />
+      {isMapDisabled(plan) ? '' : <ConnectedAssignmentMapWrapper {...AssignmentWraperProps} />}
```
moshthepitt, line 262: Should we maybe change this to !isMapDisabled && <ConnectedAssignmentMapWrapper {...AssignmentWraperProps} />? This is so that we avoid the weird empty string output

src/containers/pages/PlanAssignment/index.tsx
```diff
@@ -72,6 +74,18 @@ interface PlanAssignmentProps extends JurisdictionTableProps {
...
+ * Check if map should be visible based on plan type
+ */
+export const isMapDisabled = (plan: PlanDefinition | null): boolean => {
+  let interventionType: string | undefined = '';
+  if (plan) {
+    const type = plan.useContext.find(element => element.code === INTERVENTION_TYPE_CODE);
+    type ? (interventionType = type.valueCodableConcept) : (interventionType = '');
+  }
+  return MAP_DISABLED_PLAN_TYPES.includes(interventionType);
+};
```
moshthepitt, line 87: This looks weird to me.  Questions:

- Do we have to account for plan being null here?

Suggestions:

- If plan is null just return false
- We may be able to avoid declaring `interventionType` as a variable

src/containers/pages/PlanAssignment/index.tsx
```diff
@@ -72,6 +74,21 @@ interface PlanAssignmentProps extends JurisdictionTableProps {
...
+ * Check if map should be visible based on plan type
+ */
+export const isMapDisabled = (plan: PlanDefinition | null): boolean => {
+  if (plan) {
+    const type = plan.useContext.find(element => element.code === INTERVENTION_TYPE_CODE);
+    if (type) {
+      return MAP_DISABLED_PLAN_TYPES.includes(type.valueCodableConcept);
+    } else {
+      return false;
+    }
```
moshthepitt, line 87: I think that this else is not necessary

Issue #1043, opened 2020-07-16
The Plan Title shows an error when one creates a new plan


Naima-Bashir:
![image](https://user-images.githubusercontent.com/61143692/87659682-98100600-c766-11ea-892b-3fcd1c173a4e.png)
This relates to : https://github.com/onaio/reveal-frontend/pull/1001
https://web.reveal-stage.smartregister.org/plans/planning/new    FI
When one tries to create a new plan, the plan title already shows an error before clicking on it.

PR #1071, opened 2020-07-23
Show Plan Title error when the field is touched

ciremusyoka:
Fixes `Plan Title` field showing error all the time. 
Error will be displayed only when the field has been touched or submit button is clicked.
Fix: #1043


ciremusyoka:
ScreenShot
Screenshot from 2020-07-23 15-48-24

PR #1073, opened 2020-07-24
Stop showing loader if user has no superset permissions

### LINKED ISSUE #1013: RVL-1019 - Set friendly message when trying to access the Monitor page on WebUI without Superset Permissions
- [ ] Presently when a user tries to access the Monitor Page on the WebUI without the superset permissions, the page goes into an endless spinner and does not load. There is need to give a friendly message such as "This user does not have permissions to access this page" instead of the spinner.

kelvin-muchiri:
Closes #1013 

Stop showing loader if result is falsy


src/containers/pages/FocusInvestigation/active/tests/index.test.tsx
```diff
@@ -434,4 +435,36 @@ describe('containers/pages/ActiveFocusInvestigation', () => {
         .prop('data')
     ).toEqual([selectedPlan1]);
   });
+
+  it('should handle falsy result correctly', async () => {
+    const supersetServiceMock: any = jest
+      .fn(() => Promise.resolve(fixtures.plans))
+      .mockImplementationOnce(async () => null);
```
ciremusyoka, line 442: Why this two and `supersetService` is a function?
Can't we have `.fn(() => Promise.resolve(null))` only?

kelvin-muchiri: The intention was to have the first call (Called in component did mount) to return null and the [default](https://jestjs.io/docs/en/mock-function-api#mockfnmockimplementationoncefn) implementation to return correctly. Your suggested implementation would still be ok

src/containers/pages/FocusInvestigation/active/index.tsx
```diff
@@ -144,16 +145,19 @@ class ActiveFocusInvestigation extends React.Component<
     ]);
     supersetService(SUPERSET_PLANS_SLICE, supersetParams)
       .then((result: Plan[]) => {
```
p-netm, line 147: @kelvin-muchiri should we also change this type. what do we expect to recieve when there is an error?

p-netm:
@kelvin-muchiri nice work so far, I think there is something missing
>  There is need to give a friendly message such as "This user does not have permissions to access this page"

I think we have response message from a failed request with 401.
```{"message":"Access is Denied","severity":"danger"}
```
Maybe we can use this to know if we should display the above message or if its any other error, yes?

kelvin-muchiri:
@p-netm Can we get the status from the error object returned in catch block?

kelvin-muchiri:
@p-netm I am thinking this check on permissions should be done be done here. We should do it where all HTTP requests are being intercepted and have the 401 or 403 errors handled from there

src/containers/pages/FocusInvestigation/active/index.tsx
kelvin-muchiri: When the result is falsy it returns false

kelvin-muchiri: When 401 error is encountered the result is **undefined**

p-netm: @kelvin-muchiri then the the type should be `Plan[] | undefined`, yes? because otherwise `Plan[]` is a type that is forever truthy.

p-netm:
> @p-netm Can we get the status from the error object returned in catch block?

after looking at the code, it seems we can access the returned response, be it an error message from [here](https://github.com/onaio/reveal-frontend/blob/e131f99ed90189dc4eae57beebe8739de6850c5e/src/services/superset/index.ts#L17)

p-netm:
> @p-netm I am thinking this check on permissions should not be done here. We should do it where all HTTP requests are being intercepted and have the 401 or 403 errors handled from there

@kelvin-muchiri where in the code do you feel this best belongs?

kelvin-muchiri:
@kelvin-muchiri Maybe in supersetFetch, I presume. I have created an issue to follow up on this https://github.com/onaio/reveal-frontend/issues/1084. Otherwise, the loader now stops showing if the user has no superset permissions

PR #1074, opened 2020-07-24
Revert to initialize ReactMapboxGl inside GisidaLite component

### LINKED ISSUE #1056: Some fill layers on the map are not being rendered correctly
The map jankiness addressed by the PR https://github.com/onaio/reveal-frontend/pull/1021 was fixed by having the `Mapbox` instance created outside the component to prevent new instances from being created when props change. However, some fill layers stopped being rendered correctly. For instance, some structures that were to be rendered with a fill of yellow are now being rendered with a fill of green

kelvin-muchiri:
Closes #1056

The map jankiness addressed by the PR https://github.com/onaio/reveal-frontend/pull/1021 was fixed by having the Mapbox instance created outside the component to prevent new instances from being created when props change. However, some fill layers stopped being rendered correctly. For instance, some structures that were to be rendered with a fill of yellow are now being rendered with a fill of green. The FI superset data is currently not available and so a fix to this bug has been flagged as blocked. A revert to have `ReactMapboxGl` component initialization in component `GisidaLite` is essential so as to not carry forward this bug to the next release


p-netm:
@kelvin-muchiri please update this branch

PR #1076, opened 2020-07-27
Add structure summary table

p-netm:
Add the structure summary Table and tests.
This is the table component that is used on the second step of auto selection of jurisdictions.
it pulls a tree hierarchy structure that only has nodes that are in a path that has a selected leaf node.
It shows the structures count at any level of the drill down for only the selected leaf nodes.
leaf nodes will be already selected in the auto selection slider view which is the step that preceeds this one.


moshthepitt:
Lets get tests passing

moshthepitt:
Lets also get some screenshots

p-netm:
ss


*mock* autoselect `Lunga`  somwhere in Zambia hierarchy

sss

PR #1083, opened 2020-07-28
Add assignment page map handlers

### LINKED ISSUE #1050: Adding Handlers to Assignment page maps
jurisdiction assignment View:
  - can click on a jurisdiction fill to drill down 
both Jurisdiction assignment and plan assignment view:
  - show jurisdiction name on hover.

### LINKED ISSUE #1054: AssigmentMapWrapper changes
- [x] On the JurisdictionAssignment page map, only selected jurisdictions should have a blue border around them
- [x] filterGeom in `JurisdictionGeomFilters` should be true to ensure only jurisdictions with geometry are returned
- [x] Lets add docstrings to `AssignmentMapWrapper`
- [x] `getJurisdictions` should only be called for jurisdictions that we do not yet have in the redux store or jurisdictions in the redux store whose geometry we do not have.  The intention is to limit the requests to the API for jurisdictions.  Please see https://github.com/onaio/reveal-frontend/issues/973#issuecomment-656681240
- [x] add tests to ensure `getJurisdictions` does not get re-called for jurisdictions that are already in the redux store and whose geometry we already have
- [x] In AssignmentMapWrapper `fillColor`, `fillOutlineColor`, and `lineColor` should use colors imported from the colors module and not magic strings
- [x] Consider if indeed `fillColor`, `fillOutlineColor`, and `lineColor` should be properties that are stored to the redux store.  I argue that they should not be stored there because:
    - The reducer module in questions deals with OpenSRP jurisdictions and doe snot really care about mapbox-related properties
    - If these are persisted to redux then whether or not they are re-used across different plans inadvertently becomes unclear.  We should remove this uncertainty
- [x] Reduce [this snapshot](src/containers/pages/AssigmentMapWrapper/tests/__snapshots__/index.test.tsx.snap) size in consultation with https://kentcdodds.com/blog/effective-snapshot-testing/

KipSigei:
Fix #1050  and Fix #1054


KipSigei:
To Do:

- [ ] Shift + click - should select jurisdiction on map
- [ ] Click - should drill down jurisdictions on map
- [ ] Clicking on jurisdiction checkbox should update layer with blue boundary

src/components/page/HeaderBreadcrumb/HeaderBreadcrumb.tsx
```diff
@@ -12,23 +12,49 @@ export interface Page {
 export interface BreadCrumbProps {
   currentPage: Page;
   pages: Page[];
+  fetchUpdatedCurrentParentHandler: (currentParentId: string, isRootJurisdiction: boolean) => void;
```
moshthepitt, line 15: Could you describe some more either as a comment or in the PR what this is for?

src/containers/pages/AssigmentMapWrapper/index.tsx
```diff
@@ -32,18 +49,93 @@ export interface AssignmentMapWrapperProps {
...
+  selectNodeCreator: selectNode,
   serviceClass: OpenSRPService,
 };
 
+/**
+ * Handler to get current parent id from clicked jurisdiction on map
+ * @param props - AssignmentMapWrapper component props
+ */
+
+export const onJurisdictionClick = (props: any, setMapParent: any) => {
```
moshthepitt, line 79: Lets not use `any` here.

src/containers/pages/AssigmentMapWrapper/index.tsx
```diff
@@ -32,18 +49,93 @@ export interface AssignmentMapWrapperProps {
...
+    const { point, target, originalEvent } = e;
+    const {
+      fetchUpdatedCurrentParentActionCreator,
+      currentParentNode,
+      rootJurisdictionId,
+      selectNodeCreator,
+      deselectNodeCreator,
+    } = props;
+    // grab underlying features from map
+    const features: any = target.queryRenderedFeatures(point) as Dictionary[];
```
moshthepitt, line 91: Lets not use `any` here.

src/containers/pages/AssigmentMapWrapper/index.tsx
```diff
@@ -32,18 +49,93 @@ export interface AssignmentMapWrapperProps {
...
+    // dont drill down map if current jurisdiction is leafNode
+    if (!features.length) {
+      return false;
+    }
+    if (features[0]) {
```
moshthepitt, line 96: What's the significance of 0 here?  Can we avoid using a [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)) like this? I'd prefer either a named constant or a map.

moshthepitt: It seems that `features` is an array of something but we are only using the first element.  Why?  Would it make sense to structure it different if we only need one element?

src/containers/pages/AssigmentMapWrapper/index.tsx
moshthepitt: Can we use a better defined type than `Dictionary[]`?

src/containers/pages/AssigmentMapWrapper/index.tsx
moshthepitt, line 79: Can we move this function to a utils or helpers module? Something like `../AssigmentMapWrapper/{utils/helpers}.tsx `

src/containers/pages/FocusInvestigation/active/index.tsx
```diff
@@ -170,7 +170,7 @@ class ActiveFocusInvestigation extends React.Component<
   }
 
   public render() {
-    const breadcrumbProps: BreadCrumbProps = {
+    const breadcrumbProps: Partial<BreadCrumbProps> = {
```
moshthepitt, line 173: Why is this necessary in this PR?

src/containers/pages/JurisdictionAssignment/JurisdictionCell/index.tsx
```diff
@@ -33,7 +33,12 @@ const NodeCell = (props: NodeCellProps) => {
   }
 
   return (
-    <Link to={`${baseUrl}/${node.model.id}`} key={`${node.model.id}-span`}>
+    <Link
+      to={`${baseUrl}/${node.model.id}`}
+      key={`${node.model.id}-span`}
+      // tslint:disable-next-line: jsx-no-lambda
+      onClick={() => fetchUpdatedCurrentParent('', false)}
```
moshthepitt, line 40: Why does this have to happen as an onClick event to `Link`?  Can we find another way please?

src/containers/pages/JurisdictionAssignment/JurisdictionTable/index.tsx
```diff
@@ -73,14 +82,16 @@ export interface JurisdictionSelectorTableProps {
...
-  autoSelectNodesCreator: typeof autoSelectNodes;
-  fetchPlanCreator: typeof addPlanDefinition;
+  selectNodeCreator: ActionCreator<SelectNodeAction>;
+  deselectNodeCreator: ActionCreator<DeselectNodeAction>;
+  autoSelectNodesCreator: ActionCreator<AutoSelectNodesAction>;
+  fetchPlanCreator: ActionCreator<AddPlanDefinitionAction>;
+  fetchUpdatedCurrentParentActionCreator: typeof fetchUpdatedCurrentParent;
   selectedLeafNodes: TreeNode[];
-  leafNodes: TreeNode[];
+  leafNodes: TreeNode[] | undefined;
```
moshthepitt, line 91: This should be `leafNodes?: TreeNode[]`

src/containers/pages/JurisdictionAssignment/JurisdictionTable/index.tsx
```diff
@@ -116,11 +132,17 @@ const JurisdictionTable = (props: JurisdictionSelectorTableProps) => {
...
+    mapCurrentParent,
+    fetchUpdatedCurrentParentActionCreator,
     serviceClass,
     autoSelectionFlow,
     deselectAllNodesCreator,
   } = props;
 
+  function fetchUpdatedCurrentParentHandler(currentParentId: string, isRootJurisdiction: boolean) {
+    fetchUpdatedCurrentParentActionCreator(currentParentId, isRootJurisdiction);
+  }
```
moshthepitt, line 144: Is it necessary to wrap `fetchUpdatedCurrentParentActionCreator` like this?

src/containers/pages/JurisdictionAssignment/JurisdictionTable/index.tsx
```diff
@@ -169,7 +191,7 @@ const JurisdictionTable = (props: JurisdictionSelectorTableProps) => {
   // been set, if drilling down has begun.
   let derivedParentNode = currentParentNode;
   let derivedChildrenNodes = currentChildren;
-  if (!props.currentParentId) {
+  if (!props.currentParentId && !mapCurrentParent.currentParentId.length) {
```
moshthepitt, line 194: Why do we need both props.currentParentId and mapCurrentParent.currentParentId?  These should be the same value at all times, no?

Also, I prefer that we destructure `currentParentId` from props

src/store/ducks/opensrp/hierarchies/index.tsx
```diff
@@ -128,6 +146,16 @@ export function fetchTree(
...
+export function fetchUpdatedCurrentParent(
+  currentParentId: string,
+  isRootJurisdiction: boolean
+): FetchUpdatedCurrentParent {
+  return {
+    currentParentId,
+    isRootJurisdiction,
+    type: FETCH_MAP_CURRENT_PARENT,
+  };
+}
```
moshthepitt, line 158: I am struggling to see why this is necessary.  It seems that we are using this as a way to keep the map's current parent in the redux store.  However, do we not already have that current parent i.e. when you are drilling up and down the JurisdictionTable component, you know the current parent at all times.  Why not just pass that to the map instead of doing this?

src/containers/pages/AssigmentMapWrapper/index.tsx
KipSigei: The reason for this is sometimes when features on map are being queried it would return an array of duplicate feature objects

src/containers/pages/JurisdictionAssignment/JurisdictionTable/index.tsx
KipSigei: This was to avoid having to transform the `HeaderBreadCrumb` component to connect it to store and just have a handler passed to it as a prop to be able to dispatch the `fetchUpdatedCurrentParent` action.

moshthepitt: Why do we need to dispatch that action creator when you click on the link?  I would like us to find a way to avoid that.

src/store/ducks/opensrp/hierarchies/index.tsx
KipSigei: The reason i added this action is that when drilling up and down using the map the we need to know what jurisdiction children need to be rendered next on the table/breacrumb since `currentParentId` will remain the same

src/components/GisidaLite/tests/index.test.tsx
```diff
@@ -10,7 +10,7 @@ describe('components/GisidaLite/arePropsEqual', () => {
   it('returns false if layers length has changed', () => {
     const prevProps: GisidaLiteProps = {
       layers: [],
-      mapCenter: undefined,
+      mapCenter: [10, 10],
```
moshthepitt, line 13: Does this and line 38 below mean that the map center is by default defined to [10, 10]?  Why?

src/containers/pages/AssigmentMapWrapper/helpers/utils.tsx
```diff
@@ -0,0 +1,110 @@
...
+    if (!features.length) {
+      return false;
+    }
+    if (features[0]) {
+      const currentId =
+        (features[0].id && features[0].id.toString()) ||
+        (features[0] && features[0].properties && features[0].properties.externalId);
+      if (
+        originalEvent.altKey &&
+        PLAN_TYPES_WITH_MULTI_JURISDICTIONS.includes(getInterventionType.valueCodableConcept)
```
moshthepitt, line 59: At this point do we need to consider that `getInterventionType` might be undefined?

src/containers/pages/AssigmentMapWrapper/index.tsx
```diff
@@ -1,49 +1,97 @@
...
+  features: [],
+  type: 'FeatureCollection',
+};
+
+const defaultProps: AssignmentMapWrapperProps = {
+  autoSelectNodesActionCreator: autoSelectNodes,
+  autoSelectionFlow: false,
   currentChildren: [],
   currentParentId: undefined,
+  currentParentNode: undefined,
```
moshthepitt, line 74: I do not think it is necessary to declare `currentParentNode` as undefined, as that is already catered for in the interface.

src/containers/pages/FocusInvestigation/map/active/helpers/utils.tsx
```diff
@@ -480,3 +480,64 @@ export const buildOnClickHandler = (currentPlanId: string) => {
...
+ */
+
+export const buildMouseMoveHandler = () => {
+  function mouseHoverHandler(map: Map, event: EventData) {
+    const popup = new mapboxgl.Popup({
+      closeButton: false,
+    });
+    // Added this part to remove duplicate popup containers showing on map when hovering
+    const nodes: any = document.getElementsByClassName('mapboxgl-popup');
+    let node: any;
```
moshthepitt, line 497: Is it possible to get rid of `any` here?

src/containers/pages/FocusInvestigation/map/active/helpers/utils.tsx
```diff
@@ -480,3 +480,64 @@ export const buildOnClickHandler = (currentPlanId: string) => {
...
+    // grab underlying features from map
+    const features: any = event.target.queryRenderedFeatures(event.point) as Dictionary[];
+    //
+    if (!features.length) {
+      return;
+    }
+    // loop through features, setting the name property to content var
+    for (feature of features) {
+      if (feature && feature.properties && feature.properties.name) {
+        content += `<div><center>${feature.properties.name}</center></div>`;
```
moshthepitt, line 517: Let's give this div a class name so that in the future if we want to style it we can do so easily.

src/store/ducks/opensrp/jurisdictions/index.ts
```diff
@@ -146,12 +174,64 @@ export const getJurisdictionsArray = () =>
...
+    ],
+    (
+      jurisdictionsArray,
+      newFeatureProps,
+      currentChildren,
+      rootJurisdictionId,
+      planId,
+      metaDataByJurisdiction
+    ): FeatureCollection => {
+      const validFeatures = jurisdictionsArray.filter(item => 'geometry' in item) as Feature[];
```
moshthepitt, line 193: There's already a selector that does this (`filterGeom`), no?  Why not use it?

src/containers/pages/AssigmentMapWrapper/helpers/utils.tsx
```diff
@@ -0,0 +1,110 @@
...
+      ) {
+        let activeCurrentNode: TreeNode | any = {};
```
moshthepitt, line 61: Lets not use `any` here.

src/containers/pages/AssigmentMapWrapper/helpers/utils.tsx
```diff
@@ -0,0 +1,110 @@
...
+      deselectNodeCreator,
+      plan,
+      autoSelectionFlow,
+      getJurisdictionsMetadata,
+    } = props;
+
+    const { useContext } = plan;
+    const getInterventionType = useContext.find(
+      context => context.code === 'interventionType'
+    ) as UseContext;
```
p-netm, line 43: lets use [this](https://github.com/onaio/reveal-frontend/blob/bd679d507b221c883245ad095c144c263811583a/src/helpers/utils.tsx#L882)

src/containers/pages/AssigmentMapWrapper/index.tsx
```diff
@@ -1,49 +1,97 @@
...
+  currentParentId?: string;
+  jurisdictionsChunkSize: number;
+  getJurisdictionsMetadata: Dictionary<Dictionary<Dictionary<Meta>>>;
   currentChildren: TreeNode[];
   serviceClass: typeof OpenSRPService;
   fetchJurisdictionsActionCreator: typeof fetchJurisdictions;
   getJurisdictionsFeatures: FeatureCollection;
+  autoSelectNodesActionCreator: ActionCreator<AutoSelectNodesAction>;
+  selectNodeCreator: ActionCreator<SelectNodeAction>;
+  deselectNodeCreator: ActionCreator<DeselectNodeAction>;
```
p-netm, line 59: nitPick: `ActionCreator<Action>` is not as type-safe as `typeof actionCreator`

src/containers/pages/AssigmentMapWrapper/index.tsx
```diff
@@ -52,12 +100,17 @@ const AssignmentMapWrapper = (props: AssignmentMapWrapperProps) => {
     getJurisdictionsFeatures,
     currentParentId,
     rootJurisdictionId,
+    jurisdictionsChunkSize,
   } = props;
   const currentChildIds: string[] = currentChildren.length
```
p-netm, line 105: I believe the conditional check for this is unnecessary: `[].map(callback) => []`;

src/containers/pages/JurisdictionAssignment/AutoSelectJurisdictions/index.tsx
```diff
@@ -260,7 +260,10 @@ const mapStateToProps = (
 ): MapStateToProps => {
   const planId = ownProps.match.params.planId;
   const planObj = getPlanDefinitionById(state, planId);
-  const tree = treeByIdSelector(state, { rootJurisdictionId: ownProps.match.params.rootId });
+  const tree = treeByIdSelector(state, {
+    planId,
```
p-netm, line 264: why was planId included in this selector's filter.

src/containers/pages/JurisdictionAssignment/EntryView/JurisdictionAssignmentReRouting.tsx
```diff
@@ -144,7 +144,10 @@ const mapStateToProps = (
   ownProps: JurisdictionAssignmentReRoutingProps
 ): MapStateToProps => {
   return {
-    tree: treeSelector(state, { rootJurisdictionId: ownProps.rootJurisdictionId }),
+    tree: treeSelector(state, {
+      planId: ownProps.plan.identifier,
```
p-netm, line 148: why was planId included in this selector's filter.

src/containers/pages/JurisdictionAssignment/JurisdictionAssignmentView/index.tsx
```diff
@@ -219,7 +222,7 @@ const mapStateToProps = (
   const planObj = getPlanDefinitionById(state, planId);
   return {
     plan: planObj,
-    tree: treeByIdSelector(state, { rootJurisdictionId: ownProps.match.params.rootId }),
+    tree: treeByIdSelector(state, { rootJurisdictionId: ownProps.match.params.rootId, planId }),
```
p-netm, line 225: why was planId included in this selector's filter.

src/containers/pages/JurisdictionAssignment/JurisdictionTable/index.tsx
```diff
@@ -67,18 +75,18 @@ export interface JurisdictionSelectorTableProps {
...
   currentParentNode?: TreeNode;
   currentChildren: TreeNode[];
-  selectNodeCreator: typeof selectNode;
-  deselectNodeCreator: typeof deselectNode;
-  autoSelectNodesCreator: typeof autoSelectNodes;
-  fetchPlanCreator: typeof addPlanDefinition;
+  selectNodeCreator: ActionCreator<SelectNodeAction>;
+  deselectNodeCreator: ActionCreator<DeselectNodeAction>;
+  autoSelectNodesCreator: ActionCreator<AutoSelectNodesAction>;
+  fetchPlanCreator: ActionCreator<AddPlanDefinitionAction>;
```
p-netm, line 88: `ActionCreator<Action>` is not type-safe as `typeof actionCreator`. You can try removing one of the args from any of the `ActionCreator<Action>`-typed action creators and you will see that no error is thrown(or atleast my editor does not show the red scriggly lines.)

src/containers/pages/PlanAssignment/index.tsx
```diff
@@ -308,7 +308,10 @@ const mapStateToProps = (state: Partial<Store>, ownProps: any): DispatchedStateP
   const organizations = getOrganizationsArray(state);
   const assignments = getAssignmentsArrayByPlanId(state, ownProps.match.params.planId);
 
-  const tree = treeSelector(state, { rootJurisdictionId: ownProps.match.params.planId });
+  const tree = treeSelector(state, {
+    planId: (plan as PlanDefinition).identifier,
```
p-netm, line 312: why was planId included in this selector's filter.

src/store/ducks/opensrp/hierarchies/index.tsx
```diff
@@ -66,6 +69,12 @@ export interface DeforestAction extends AnyAction {
   treeByRootId: {};
 }
 
+export interface FetchUpdatedCurrentParent extends AnyAction {
```
p-netm, line 72: is this being used elsewhere?

src/store/ducks/opensrp/hierarchies/index.tsx
```diff
@@ -355,7 +363,7 @@ export interface WithoutMetaFilters {
 /** prop filters to customize selector queries */
 export interface Filters {
   rootJurisdictionId: string /** specify which tree to act upon */;
-  planId?: string /** get selections that are associated with this planId */;
+  planId: string /** get selections that are associated with this planId */;
```
p-netm, line 366: IMO making planId required will require one to supply it even for selectors like `getTreeById` that does not use it,

src/store/ducks/opensrp/hierarchies/index.tsx
```diff
@@ -547,7 +553,14 @@ export const getCurrentChildren = () => {
  * @param props -  the filterProps
  */
 export const getAllSelectedNodes = () =>
-  createSelector(getTreeWithMeta(), getLeafNodesOnly, computeSelectedNodes);
+  createSelector(
+    getTreeWithMeta(),
```
p-netm, line 557: heads up:  getTreeWithMeta will be removed.

src/containers/pages/AssigmentMapWrapper/helpers/utils.tsx
moshthepitt: Actually looks like you should use [this](https://github.com/onaio/reveal-frontend/blob/bd679d507b221c883245ad095c144c263811583a/src/helpers/utils.tsx#L894).

src/store/ducks/opensrp/hierarchies/index.tsx
moshthepitt: @p-netm how do you propose we resolve that in this PR?

p-netm: @moshthepitt by merging #1127

src/components/GisidaLite/tests/index.test.tsx
KipSigei: I added that value for testing purposes

moshthepitt: Cool

KipSigei:
> When i ran this locally, I encountered these issues:
> 
> * can drill down in the map past the operational area.
> * make cursor, pointer if clickable
> * there is no popup for Jurisdiction name on hover => I expected to see this since it looks like there is a mouseMoveHandler added in the pr

@p-netm These have been addressed you can check once more when you get time

src/store/ducks/opensrp/hierarchies/index.tsx
```diff
@@ -55,6 +55,9 @@ export const DESELECT_ALL_NODES = 'opensrp/locations/hierarchy/DESELECT_ALL_NODE
 // /** action to auto-append/ auto-modify the selected attribute of nodes in a tree */
 export const AUTO_SELECT_NODES = 'opensrp/locations/hierarchy/AUTO_SELECT_NODES';
 
+/** action to fetch updated children to ensure map and jurisdiction table are in sync */
+export const FETCH_MAP_CURRENT_PARENT = 'opensrp/locations/hierarchy/FETCH_MAP_CURRENT_PARENT';
```
p-netm, line 59: is this being used somewhere?

KipSigei: @p-netm  addressed via https://github.com/onaio/reveal-frontend/pull/1083/commits/6ca235ed7f320553ff768af85161ae54f5f39023

src/containers/pages/AssigmentMapWrapper/index.tsx
```diff
@@ -1,49 +1,93 @@
...
+    identifier: '',
+  } as PlanDefinition,
   rootJurisdictionId: '',
+  selectNodeCreator: selectNode,
   serviceClass: OpenSRPService,
 };
 
+/**
+ * This is a map HOC for the plan and jurisdiction assignment pages
+ * It is responsuble for fetching jurisdiction geojson and wiring down to
```
moshthepitt, line 86: typo here

src/containers/pages/FocusInvestigation/map/active/helpers/utils.tsx
```diff
@@ -480,3 +480,65 @@ export const buildOnClickHandler = (currentPlanId: string) => {
...
   return clickHandler;
 };
+
+/**
+ * Map component popup handler
+ * @param map - mapbox map instance
+ * @param event - map event object
+ */
+
+export const buildMouseMoveHandler = (map: Map, event: EventData) => {
```
moshthepitt, line 490: Let's move this either to GisidaLite utils or the AssignmentMapWrapper utils.  I feel that it doesn't have much to do with FocusInvestigation

Issue #1072, opened 2020-07-23
 the Jurisdiction Metadata page doesn't create CSVs with the proper extension when you click the download button


Naima-Bashir:
1. Login to the webapp
2. Click on Admin- Jurisdiction metadata and download the CVS templates. 
3. Click on the downloaded csv templates and observe if they have the proper extensions

![image](https://user-images.githubusercontent.com/61143692/88289929-a8902580-ccfe-11ea-8475-806d77097e15.png)
![image](https://user-images.githubusercontent.com/61143692/88380952-26f7d080-cdae-11ea-8a79-54ade1fc53ba.png)

I think we need to add the .csv to the filename.

moshthepitt:
Closed by https://github.com/onaio/reveal-frontend/pull/1087

PR #1087, opened 2020-07-29
fixes for csv download and upload on windows machines

### LINKED ISSUE #1008: CSV Upload on Windows Chrome, Microsoft Edge and Firefox does not work as expected -RVL-989/1018
1- Downloaded the Template CSV ; it downloaded successfully. However, I tried to find the downloaded file and I couldn't find it(it's there but when i try to upload it, it's not showing). See image of previous templates I had downloaded.
2- I tried to upload a csv file and I got this error below. Either because I cant find the recently downloaded file or that I am using a old csv.
3- I have tried to upload again, logged out and logged in again, still the same.
on Windows Chrome the upload functionality does not work but on Ubuntu chrome it does.
image
https://github.com/onaio/reveal-frontend/pull/970
Same with Edge and Firefox
image

gmwasi:
- Modification on the download process making the filename include the file extension instead of being derived from the MIME type.
- Slacked the file validation during upload and instead having the validation be for the file extension to ease validation for windows machines which have numerous MIME types for CSV files depending on programs installed
- Resolves #1008 #1072


src/components/forms/JurisdictionMetadata/index.tsx
```diff
@@ -39,23 +25,9 @@ import {
 } from '../../../helpers/utils';
 import { OpenSRPService } from '../../../services/opensrp';
 
-const SUPPORTED_FORMATS = [
```
moshthepitt, line 42: Why did we need to remove these?

gmwasi: On widows, CSV files have different MIME types depending on applications installed. I couldn't have an exhaustive list on this and so I decided just to validate by file extension instead. This Is sufficient because the user still has to download a template before uploading

PR #1089, opened 2020-07-29
Add reasons for not sprayed structures to map indicators list

### LINKED ISSUE #1069: RVL-1027 - Add Indicator with Reasons for not-sprayed structures
1. Go to the Zambia Reveal WebUI > Select Monitor > select IRS Reporting > Select a completed IRS Plan. On the dashboard table that comes up, drill down to the lowest level:
- On the indicators list to the right hand side of the map, add an indicator named `Reasons for not-sprayed structures` . This would read e.g. 7 out of 100 structures not sprayed. The different reasons given for not sprayed will then be given as aggregates e.g.   Sick - 1; Locked - 0,  Funeral - 0; Refused - 2; Adult not present - 4; Other - 0

ciremusyoka:
Closes #1069 
Add a list of reason why some structures were not sprayed and an over view of the number of structured not sprayed.


ciremusyoka:
ScreenShot
Screenshot from 2020-07-29 16-56-47

src/store/ducks/generic/jurisdictions.ts
```diff
@@ -58,6 +59,8 @@ export interface ZambiaIRSFocusArea extends GenericJurisdiction {
   spraytarg: number;
   spraysuccess: number;
   roomcov: number;
+  notsprayed_reasons: string[];
+  notsprayed_reasons_counts: Dictionary;
```
moshthepitt, line 63: I think that we can make this type more accurate.  Seems like it is Dictionary<number>

ciremusyoka:
> based on the screenshot, the number of not-sprayed structures should be `16 of 72`. You can arrive at 16 in one of two ways:

Have updated this calculation.

src/containers/pages/IRS/Map/index.tsx
```diff
@@ -255,6 +256,20 @@ const IRSReportingMap = (props: IRSReportingMapProps & RouteComponentProps<Route
 
   const gisidaWrapperProps = getGisidaWrapperProps(jurisdiction, structures, indicatorStops);
 
+  /**
+   * Create list elements from dictionary
+   * @param {Dictionary} dict
+   */
+  const processListDisplay = (dict: Dictionary | string) => {
+    const createList = (obj: Dictionary) =>
+      Object.keys(obj).map(key => (
```
moshthepitt, line 265: Can we use [Object.entries](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries)?  Is it better?

ciremusyoka: Yah! it is better.

src/containers/pages/IRS/Map/index.tsx
```diff
@@ -255,6 +256,20 @@ const IRSReportingMap = (props: IRSReportingMapProps & RouteComponentProps<Route
 
   const gisidaWrapperProps = getGisidaWrapperProps(jurisdiction, structures, indicatorStops);
 
+  /**
+   * Create list elements from dictionary
+   * @param {Dictionary} dict
+   */
+  const processListDisplay = (dict: Dictionary | string) => {
+    const createList = (obj: Dictionary) =>
+      Object.entries(obj).map(keyVal => (
```
p-netm, line 265: nitpick: `Object.entries().map(([key, value]) => {})`

Issue #1091, opened 2020-07-30
RVL-1033 - Setting a plan to "Active" during plan creation results in the save button hanging


samkanga:
Here are the steps to reproduce this:

1. Select Plan > Manage Plans > Create new Plan
2. Select the plan type as Dynamic IRS and the desired jurisdiction. Set its status to Active and move to the bottom of the page to save the plan.
3. The save button appears to freeze and does not complete.
4. However, when one moves to the assignment page the plan is listed with an active status.

See below the a screenshot on the frozen screen.
![image](https://user-images.githubusercontent.com/5908630/88903874-01ab0c80-d25d-11ea-8b7a-f586d634e5cf.png)

kelvin-muchiri:
@moshthepitt A 500 error is being thrown by the server

kelvin-muchiri:
I noticed that the form global error messages are being displayed on top of the form, that's why it seems the error message is not visible. Also, it does not have a distinctive colour. A usability problem that should be considered for fixing
![Screenshot from 2020-07-30 13-57-53](https://user-images.githubusercontent.com/19818819/88915421-e8aa5780-d26c-11ea-9255-4b4156532efd.png)

kelvin-muchiri:
The form is long and an error message at the top of the form might not be noticeable if the user has scrolled way down the form

moshthepitt:
@kelvin-muchiri 

- please get a sample payload that fails with an error 500 and document it here
- please change the error to be displayed on the page --> maybe make the error toast message stay around until the user manually clicks to close it?

kelvin-muchiri:
This request payload fails with 500
```
[...]

PR #1093, opened 2020-07-30
Display plan form global errors as toast errors

kelvin-muchiri:
Partially addresses #1091 

In order to improve the usability on the PlanForm, the global errors are now displayed as toast errors and not at the top of the form as before with the toast requiring to be closed manually


src/helpers/errors.ts
```diff
@@ -1,38 +1,50 @@
...
-import { toast } from 'react-toastify';
+import { toast, ToastOptions } from 'react-toastify';
 import { ACCESS_DENIED, NETWORK_ERROR, USER_HAS_NO_VALID_ASSIGNMENTS } from '../configs/lang';
 import { growl } from './utils';
 
 /** union of all ErrorTypes that we are working with */
 type ServiceError = HTTPError | Error | NetworkError;
 
+/** autoClose types */
+type AutoClose = 5000 | false;
```
ciremusyoka, line 10: Should we fix this 5000ms? I think it should be flexible for different use cases.

kelvin-muchiri: I was thinking in terms of not accepting any value. We define what values are acceptable. So that we can avoid a developer passing 1000 for instance which will make the toast very short or perhaps a value of 10000 which will make it very long. We can add more acceptable values in the type if need be

kelvin-muchiri: Though 5000 is the [default value](https://fkhadra.github.io/react-toastify/api/toast-container) used by `react-toastify`

ciremusyoka: My view is we don't need to fix it but to have 5000 as the default just the way you have put it.

p-netm: @kelvin-muchiri I will have to agree with Erick on this. i do not think we should have  a parameter that is in essence a constant.Also please bear in mind that we have an env that controls the auto close period across the whole app: `TOAST_AUTO_CLOSE_DELAY`.

src/helpers/errors.ts
```diff
@@ -1,38 +1,50 @@
...
+ * the notification needs to be closed manually
  */
-export const displayError = (error: ServiceError, customMessage: string = '') => {
+
+export const displayError = (
+  error: ServiceError,
+  customMessage: string = '',
+  autoClose: AutoClose = 5000
+) => {
   /** opensrp error object for plans after filtering plans by a user when the assignments are expired
```
p-netm, line 25: lets also remove this string. looks like it describes something that was removed

src/helpers/errors.ts
```diff
@@ -20,11 +18,8 @@ type AutoClose = 5000 | false;
 export const displayError = (
   error: ServiceError,
   customMessage: string = '',
-  autoClose: AutoClose = 5000
+  autoClose: TOAST_AUTO_CLOSE_DELAY | false = TOAST_AUTO_CLOSE_DELAY
```
moshthepitt, line 21: Lets have this as `number | false` please

Naima-Bashir:
image 
This has passed QA. @p-netm @samkanga

PR #1094, opened 2020-07-30
Refactor hierarchies reducer structure

p-netm:
Moves the meta field out of the tree

Meta is saved alongside the tree, we use the jurisdiction Id to know which meta object belongs to  which node in tree.

during selection the final tree is computed(the meta fields are added to thier respective nodes) such that the final tree is similar in structure to how it was previously stored

This means we can update the tree in several service calls without overriding meta information such as to which node has been selected.


moshthepitt:
@p-netm is this still [DNM]?  Could you fix tests please?

PR #1096, opened 2020-08-03
Fix map jankiness

kelvin-muchiri:
Partially address #930 
> Fix map jankiness(probably because it gets mounted with each of the component's rerenders.)
takes a while to show up,
rerenders the whole map when props change

Due to lack of superset data, the solution was implemented by using fixtures in `src/containers/pages/FocusInvestigation/map/active/tests/supersetFixtures`
 
File `src/containers/pages/FocusInvestigation/active/index.tsx` was refactored as follows

```
.....

import * as fixturesMap from '../map/active/tests/fixtures';

.....

 public componentDidMount() {
    const { userName, fetchPlansActionCreator, supersetService } = this.props;
    const supersetParams = superset.getFormData(2000, [
      { comparator: InterventionType.FI, operator: '==', subject: 'plan_intervention_type' },
    ]);
      .then((_: Plan[]) => {
        /*if (result) {
          fetchPlansActionCreator(result);
        } else {
          displayError(new Error(AN_ERROR_OCCURRED));
        }*/
        fetchPlansActionCreator(fixturesMap.processedPlansJSON);
      })
      .finally(() => {
        this.setState({
          loading: false,
        });
      })
      .catch(err => {
        displayError(err);
      });

    if (userName) {
      loadPlansByUserFilter(userName).catch(err => displayError(err));
    }
  }

....

```

File `src/containers/pages/FocusInvestigation/map/active/index.tsx` was refactored as below. To simulate delayed  async API calls that would result in a change in props, `setTimeout` was used

```
....

import * as fixturesMap from './tests/fixtures';

....

React.useEffect(() => {
    if (props.plan) {
      /**
       * Plans present in state e.g when accesing this view from a list of plans or
       * when the plans are expilcity fetched as in the above if block
       */
      /*fetchData(
        props.fetchGoalsActionCreator,
        props.fetchJurisdictionsActionCreator,
        props.fetchPlansActionCreator,
        props.fetchStructuresActionCreator,
        props.fetchTasksActionCreator,
        props.plan,
        props.supersetService
      ).catch((error: Error) => displayError(error));*/
      setTimeout(() => props.fetchPlansActionCreator(fixturesMap.processedPlansJSON), 3000);
      setTimeout(
        () => props.fetchJurisdictionsActionCreator(fixturesMap.processedJurisdictionJSON),
        5000
      );
      setTimeout(
        () => props.fetchStructuresActionCreator(fixturesMap.processedStructuresJSON),
        6000
      );
      setTimeout(() => props.fetchGoalsActionCreator(fixturesMap.processedGoalsJSON), 7000);
      setTimeout(() => props.fetchTasksActionCreator(fixturesMap.processedPlansTasksJson), 8000);
      setTimeout(
        () => props.fetchTasksActionCreator(fixturesMap.processedCaseConfirmationTasksJSON),
        9000
      );
    }
    /**
     * Only re-run effect if props.plan.plan_id changes
     */
  }, [props.plan && props.plan.plan_id]);

.....

```

Once the app is running, from the monitor, page select the second plan `Test_Cs_A1 Akros_1 2020-03-30 `

Without the [workaround](https://github.com/onaio/reveal-frontend/blob/930-map-jankiness/src/components/GisidaLite/index.tsx#L110) the layers will be rendered **INCORRECTLY** as follows

Screenshot from 2020-07-13 18-21-08

The **CORRECT** rendered map should be as follows

Screenshot from 2020-07-30 12-01-29


moshthepitt:
@kelvin-muchiri could you add some notes to the PR describing what part of the issue is fixed, and how?

kelvin-muchiri:
@moshthepitt To test this, you need to set up the fixtures as described above and test using the plan given above. Before the fix, the map used to flash everytime the props change because a [new instance](https://github.com/onaio/reveal-frontend/blob/233bcbbac5ad86e6d5563e0e4b27dc4a1c95fc76/src/components/GisidaLite/index.tsx#L83) of `ReactMapboxGl` was being recreated when the props change.

After the fix, this does not happen, you should not experience intermittent flashes props for component `GisidaLite` change

moshthepitt:
@kelvin-muchiri I mean that we should add tests for this

kelvin-muchiri:
@moshthepitt Seems difficult to test this. One way is to spy on `useState` for `GisidaLite` but this will be breaking the rules of tests since we shall be testing the implementation and not the behaviour.

moshthepitt:
@kelvin-muchiri is it possible to mock the `Mapbox` element and observe/assert what happens when props change?

kelvin-muchiri:
@moshthepitt I have tried your suggestion but it doesn't since the mock will not be the actual implementation seem to help

moshthepitt:
@kelvin-muchiri please create an issue to research this further

PR #1097, opened 2020-08-03
1079 jurisdiction assignment cleanup routing refactors

p-netm:
Refactors the containers that are using the previous implementation of the hierarchy reducer structure.

- The biggest change is that the manual and auto-selection pages will be rendered on routes that have the rootJurisdictionId param.
- they will be rendered at a time when we already have the rootId and the tree in store.
- Also the the service call to get the tree in these containers will not keep the page in a loading state while resolving; i.e ofcourse if we already have the tree in store.


moshthepitt:
@p-netm lets get tests passing

PR #1099, opened 2020-08-03
Make tree state immutable

### LINKED ISSUE #987: Make the tree structure in jurisdiction Tree reducer immutable
From [here](https://github.com/onaio/reveal-frontend/blob/714b426032f67c8d542625a8f72f9a6ec1e7eb7b/src/store/ducks/opensrp/hierarchies/index.ts#L157) the added trees objects are mutable, which seems to against our best practice of having state variables immutable. also currentParentIdsByRootId variable is now surplus, should be removed.

Part of https://github.com/onaio/reveal-frontend/issues/986

gmwasi:
resolves #987

### DIFF
--- a/src/store/ducks/opensrp/hierarchies/index.ts
+++ b/src/store/ducks/opensrp/hierarchies/index.ts
@@ -7,6 +7,7 @@
 import { Dictionary } from '@onaio/utils/dist/types/types';
 import { AnyAction, Store } from 'redux';
 import { createSelector } from 'reselect';
+import SeamlessImmutable from 'seamless-immutable';
 import { AutoSelectCallback, RawOpenSRPHierarchy, TreeNode } from './types';
 import {
   autoSelectNodesAndCascade,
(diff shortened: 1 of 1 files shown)


moshthepitt:
@gmwasi could you fix the conflicts please

gmwasi:
> @gmwasi I think we should also wrap the returned state in in the reducer i.e. per each action handled in the switch statement
> 
> Lets also add a test that tries to modify a tree in the store and fails.

@p-netm I tried wrapping the returned state and the tests throw an error 'maximum call stack exceeded' emitted by _SeamlessImmutable_. Had a chat with @moshthepitt and we decided to leave it as is for now until a solution is gotten

p-netm:
> > @gmwasi I think we should also wrap the returned state in in the reducer i.e. per each action handled in the switch statement
> > Lets also add a test that tries to modify a tree in the store and fails.
> 
> @p-netm I tried wrapping the returned state and the tests throw an error 'maximum call stack exceeded' emitted by _SeamlessImmutable_. Had a chat with @moshthepitt and we decided to leave it as is for now until a solution is gotte

hmm, the tree structure does have circular properties, I guess that might be the reason.

PR #1102, opened 2020-08-04
Fix irs reporting

p-netm:
Resolves issues reported [here](https://onaio.slack.com/archives/CBSU8L2SC/p1596514793200300)

### DIFF
--- a/src/containers/pages/GenericJurisdictionReport/helpers.ts
+++ b/src/containers/pages/GenericJurisdictionReport/helpers.ts
@@ -209,7 +209,15 @@ export const ZambiaFocusAreasColumns = [
     accessor: 'spraysuccess',
   },
   {
-    Cell: (cell: Cell) => getIRSThresholdAdherenceIndicator(cell),
+    Cell: (cell: Cell) => {
+      const { value } = cell;
+      const intValue = Number(value);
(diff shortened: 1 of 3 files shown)


moshthepitt:
@p-netm which of those issues does it resolve?  Could you get screenshots in this issue?

p-netm:
@moshthepitt  these are the addressed issues so far: 2 out of 3
| image | 
|:--:| 
| *the name column has overlapping values* |


| image (1)| 
|:--:| 
| *There's another issue with the percentages. It looks like they're calculating  "Structures remaining to spray to reach 90% SE. It looks like that's displaying a percentage, but it should be a whole number. * |

moshthepitt:
Thanks @p-netm.  Could you fix this "Structures remaining to spray to reach 90% SE"?

p-netm:
Tests for the changes included here:
- the drill-down in the genericJurisdictionReport component is able to consume `ZambiaFocusAreas2019` columns correctly
- an "integration test" for the IRS reporting map
    - the service calls.
    - the relevant props passed to the map wrapper component.
    - params during selector's invocation.
Both these tests are proving to be more involved than the time available at the moment allows for. I am moving them into a a different issue #1105

Issue #1124, opened 2020-08-07
Dynamic MDA plan cannot be assigned locations or teams 


isabelcshaw:
For NTD Zambia, we are using the Dynamic MDA plan type. When I log on to stage (user: onatest) and create a new plan type using the Zambia(vl 2020) hierarchy, I am able to create a draft plan. 

When I click back into that plan to assign locations, I cannot click on any of the check boxes to select districts, or operational areas, or even the entire country. 

Earlier, when try to assign users to a plan I was able to set to active, I get a "OpenSRPService read on location/hierarchy/plan failed, HTTP status 400" error. This plan was "Zambia IS MDA 8.20." (video linked from Slack below) 

I tried to troubleshoot this with @samkanga and he was able to create a plan and assign locations. But he could not assign locations to the plan I created. We tested with plan: "IS MDA Zambia Aug 2020." I logged in and out more than once and also opened it in an incognito browser. 

Because of this, one guess I have is that maybe it is not a problem with the hierarchy and maybe it is because I disabled the "MDA Adherence" task on plan creation - is it possible this is creating errors? @craigappl 

I am marking this as high priority because Zambia needs to be ready to go live on Friday. 

Video of the second error from Slack here: https://onaio.slack.com/archives/CBSU8L2SC/p1596819112463100?thread_ts=1596802330.446200&cid=CBSU8L2SC

moshthepitt:
@ciremusyoka please investigate what may be causing the problems described here, and then report back to this issue as a comment.

ciremusyoka:
> For NTD Zambia, we are using the Dynamic MDA plan type. When I log on to stage (user: onatest) and create a new plan type using the Zambia(vl 2020) hierarchy, I am able to create a draft plan.
>
> When I click back into that plan to assign locations, I cannot click on any of the check boxes to select districts, or operational areas, or even the entire country.

@moshthepitt on testing this have realised the [plan](https://reveal-stage.smartregister.org/opensrp/rest/plans/c593f15a-2592-5ff4-9a7b-f53d5b39f5e0) was saved without a `jurisdiction` being assigned which am not sure how it happened. The endless loading on the `Location` field is happening because the [request to get the jurisdiction associated to the plan](https://reveal-stage.smartregister.org/opensrp/rest/plans/findLocationNames/c593f15a-2592-5ff4-9a7b-f53d5b39f5e0) doesn't return any name.

[...]

Issue #1101, opened 2020-08-04
There should be an error if one tries to save a new plan without a jurisdiction 


Naima-Bashir:
Create a new plan without any jurisdictions on the planning tool.
Observe the results.
https://web.reveal-stage.smartregister.org/plans/planning/new

 The fix: check if there is at least one valid selected jurisdiction id

moshthepitt:
Fixed via https://github.com/onaio/reveal-frontend/pull/1130

Naima-Bashir:
![image](https://user-images.githubusercontent.com/61143692/89995430-1db5a180-dc92-11ea-9492-da3536100466.png)
@ciremusyoka  If you look at the plan title, it shows it's required, the jurisdiction should also have that error  if one tries to save it.

ciremusyoka:
@Naima-Bashir the reason you are seeing required is because you have deleted the plan title. Is that the case?

If you try selecting a location and and cancel an error message will be displayed

PR #1126, opened 2020-08-10
Improve create plan fields validation

ciremusyoka:
Part of #1124 and #1101
Disable create plan submit button if the form is not valid.

### DIFF
--- a/src/components/forms/PlanForm/index.tsx
+++ b/src/components/forms/PlanForm/index.tsx
@@ -279,7 +279,7 @@ const PlanForm = (props: PlanFormProps) => {
         }}
         validationSchema={PlanSchema}
       >
-        {({ errors, handleChange, isSubmitting, setFieldValue, values, touched }) => (
+        {({ errors, handleChange, isSubmitting, setFieldValue, values, touched, isValid }) => (
           <Form
             /* tslint:disable-next-line jsx-no-lambda */
             onChange={(e: FormEvent) => {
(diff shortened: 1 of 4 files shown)


src/components/forms/PlanForm/index.tsx
```diff
@@ -1061,7 +1061,7 @@ const PlanForm = (props: PlanFormProps) => {
               className="btn btn-block"
               color="primary"
               aria-label={SAVE_PLAN}
-              disabled={isSubmitting || Object.keys(errors).length > 0}
+              disabled={isSubmitting || Object.keys(errors).length > 0 || !isValid}
```
p-netm, line 1064: @ciremusyoka question? in what way is `Object.keys(errors).length > 0` different from `!isValid`

moshthepitt:
@ciremusyoka have you confirmed that this works to ensure that you cant create a plan with no jurisdiction?

src/components/forms/PlanForm/index.tsx
ciremusyoka: @p-netm from my observations sometimes before any field is touched `errors` are an empty object. That is why i decided to use the two checks.

ciremusyoka:
> @ciremusyoka have you confirmed that this works to ensure that you cant create a plan with no jurisdiction?

Yah! Currently the `Save Plan` button is disabled until the form is valid.

src/components/forms/PlanForm/index.tsx
p-netm: :+1:

PR #1127, opened 2020-08-10
Jurisdiction assignment page optimizations

### LINKED ISSUE #1103: The entire screen freezes when one clicks on the planning tool- then one plan, then checkbox  RVL-1048
Click on the planning tool
Select a plan https://web.reveal-stage.smartregister.org/manualSelectJurisdictions/6b82645a-da8b-5e7c-ad33-56d229c9c383/22bc44dd-752d-4c20-8761-617361b4f1e7
Drill down on a location. 
Click on the checkbox and observe if the page is responsive. 
RESULTS: The entire page freezes and becomes unresponsive.

image

image

p-netm:
close #1103 

 - replaced `...` object spread operator with `Object.assign` since its much faster, 
- refactored `.forEach` with the standard `for` loop
- instead of computing another whole tree where we add the metaData to respective nodes; meta is now applied to only nodes that will be returned to selectors.
- added a couple of stress tests along the way, not sure where to place them since they have no assertions.

### DIFF
--- a/src/store/ducks/opensrp/hierarchies/index.tsx
+++ b/src/store/ducks/opensrp/hierarchies/index.tsx
@@ -308,12 +308,7 @@ export default function reducer(state = initialState, action: TreeActionTypes) {
       if (!treeOfInterest) {
         return state;
       }
-      const metaByJurisdictionId = autoSelectNodesAndCascade(
-        treeOfInterest,
-        callback,
-        actionBy,
-        thisPlansMetaData
-      );
+      const metaByJurisdictionId = autoSelectNodesAndCascade(treeOfInterest, callback, actionBy);
 
       return {
         ...state,
--- a/src/store/ducks/opensrp/hierarchies/index.tsx
+++ b/src/store/ducks/opensrp/hierarchies/index.tsx
@@ -19,6 +19,7 @@ import {
   TreeNode,
 } from './types';
 import {
+  applyMeta,
   autoSelectNodesAndCascade,
   computeParentNodesSelection,
   computeSelectedNodes,
@@ -356,8 +357,6 @@ export interface Filters {
   nodeId?: string /** target node with this id */;
   leafNodesOnly?: boolean /** specified when requesting for selected nodes, truthy returns leaf nodes */;
   currentParentId?: string /** to use when filtering current children */;
-  selectedLeafNodes?: TreeNode[] /** to use when filtering from the selected leaves */;
-  parentNode?: TreeNode /** to use when filtering from the selected leaves */;
 }
 
 /** retrieve the rootJurisdiction value
@@ -438,37 +437,17 @@ export const getMetaForPlanInTree = () => {
   });
 };
 
-/** combine tree with metaData
- * @param state - the store
- * @param props - the filterProps
- */
-export const getTreeWithMeta = () => {
-  return createSelector(getTreeById(), getMetaForPlanInTree(), (origTree, metaForTree) => {
-    // walk the tree and update meta field.
-    if (!origTree) {
-      return;
-    }
-    const treeClone = cloneDeep(origTree);
-    treeClone.walk(node => {
-      node.model.meta = metaForTree[node.model.id] ? metaForTree[node.model.id] : {};
-      return true;
-    });
-
-    return treeClone;
-  });
-};
-
 /** retrieve a node given the id
  * @param state - the store
  * @param props -  the filterProps
  */
 export const getNodeById = () =>
-  createSelector(getTreeWithMeta(), getNodeId, (tree, nodeId) => {
+  createSelector(getTreeById(), getNodeId, getMetaForPlanInTree(), (tree, nodeId, metaData) => {
     if (!tree || !nodeId) {
       return;
     }
     const nodeOfInterest = tree.first(node => node.model.id === nodeId);
-    return nodeOfInterest;
+    return applyMeta(nodeOfInterest, metaData)[0];
   });
 
 /**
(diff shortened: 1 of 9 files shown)


src/store/ducks/opensrp/hierarchies/index.tsx
```diff
@@ -623,14 +569,24 @@ export const getSelectedHierarchy = () => {
...
+    // tslint:disable-next-line: prefer-for-of
+    for (let i = 0; i < allSelectedLeafNodes.length; i++) {
+      const thisNode = allSelectedLeafNodes[i];
+      const nodesInPath = thisNode.getPath();
+
+      // tslint:disable-next-line: prefer-for-of
+      for (let idx = 0; idx < nodesInPath.length; idx++) {
+        temp.push(nodesInPath[idx]);
+      }
+    }
```
moshthepitt, line 582: Is it at all possible to avoid the nested loops?

moshthepitt:
@p-netm does it now work for the plans it was failing?

src/store/ducks/opensrp/hierarchies/index.tsx
p-netm: yeah, Ill refactor that.

p-netm:
> @p-netm does it now work for the plans it was failing?

Yeah!,, though its a bit laggy for now, at least it does not freeze the page.

centered

moshthepitt:
> Yeah!,, though its a bit laggy for now, at least it does not freeze the page.

Do you think that we can show some indication that something is loading?

p-netm:
Made some changes to the selected Jurisdictions component it was calling an expensive selector for each row in the table which is a lot of times.

centered

src/store/ducks/opensrp/hierarchies/index.tsx
p-netm: @moshthepitt unfortunately I was not able to get around this. IMO I cannot avoid them here.

src/store/ducks/opensrp/hierarchies/index.tsx
```diff
@@ -438,37 +431,17 @@ export const getMetaForPlanInTree = () => {
...
  * @param state - the store
  * @param props -  the filterProps
  */
 export const getNodeById = () =>
-  createSelector(getTreeWithMeta(), getNodeId, (tree, nodeId) => {
+  createSelector(getTreeById(), getNodeId, getMetaForPlanInTree(), (tree, nodeId, metaData) => {
     if (!tree || !nodeId) {
       return;
     }
     const nodeOfInterest = tree.first(node => node.model.id === nodeId);
```
moshthepitt, line 443: What are the implications of using `.find` here?

src/store/ducks/opensrp/hierarchies/index.tsx
```diff
@@ -623,14 +564,24 @@ export const getSelectedHierarchy = () => {
...
+      }
+    }
+    allNodesInPaths = Object.assign(
+      {},
+      allNodesInPaths,
+      keyBy(temp, nd => nd.model.id)
+    );
 
-    // flatten it into an array in preparation of creating a nested structure
+    // // flatten it into an array in preparation of creating a nested structure
```
moshthepitt, line 584: Should we remove one set of `//`?

p-netm: yeah, will change.

src/store/ducks/opensrp/hierarchies/index.tsx
p-netm: .first is a method of the tree ADT we are using. I do not think it supports find.

src/store/ducks/opensrp/hierarchies/utils.tsx
```diff
@@ -202,82 +201,79 @@ export function computeParentNodesSelection(
...
+        metaByJurisdictionId = Object.assign(
+          metaByJurisdictionId,
+          createSingleMetaData(nd.model.id, actionBy, true)
+        );
         // removing this return cause a type error, that's its sole purpose
         return true;
       });
-      parentNodes.push(node);
     }
     return true;
```
moshthepitt, line 264: Is it possible to refactor this so that we only have one `return true;` statement?

src/store/ducks/opensrp/hierarchies/utils.tsx
```diff
@@ -300,27 +296,70 @@ export const preOrderStructureCountComputation = (node: TreeNode) => {
...
+
 /** find the parent node in a tree given the parentId
  * @param tree - the tree
  * @param parentId - id of the parent node
+ * @param metaData - the tree's metaData - used to know which nodes to selected
  */
-export const findAParentNode = (tree: TreeNode | undefined, parentId: string | undefined) => {
+export const findAParentNode = (
+  tree: TreeNode | undefined,
+  parentId: string | undefined,
```
moshthepitt, line 327: Should we change this to `tree?` and `parentId?`?

src/store/ducks/opensrp/hierarchies/utils.tsx
```diff
@@ -300,27 +296,70 @@ export const preOrderStructureCountComputation = (node: TreeNode) => {
...
+  return finalNode;
 };
 
 /** find the current children from  a parent node. This is defined here to avoid repetition
  * @param parentNode -  the parent Node
+ * @param metaData - the tree's metadata; used to know which nodes to select
  */
-export const getChildrenForNode = (parentNode: TreeNode | undefined) => {
+export const getChildrenForNode = (
+  parentNode: TreeNode | undefined,
```
moshthepitt, line 354: Should we change this to `parentNode?`?

moshthepitt:
@p-netm has this been integrated with the PR that made the tree reducer immutable?

src/store/ducks/opensrp/hierarchies/utils.tsx
p-netm: AFAIK, am afraid not.

src/store/ducks/opensrp/hierarchies/utils.tsx
p-netm: yeah

p-netm:
> @p-netm has this been integrated with the PR that made the tree reducer immutable?

yeah but the tree reducer is not immutable yet. There is the error that Mwasi pointed out which is caused by circular references of some properties in a tree, this is when wrapping the returned states with SeamlessImmutable. At the moment. the base selector that gets the tree from the store returns a deeply copied clone of the requested tree.

PR #1130, opened 2020-08-11
Allow saving of plans only when a jurisdiction is selected

ciremusyoka:
part of #1101 

Ensures a plan can not be saved without selecting a jurisdiction by:
- Disabling action buttons (`Save & Activate` and `Save Draft`) if no jurisdiction is selected
- Also adds a tool tip to indicate no jurisdiction selected when the action buttons are disabled.

It is also partial fix for #1124

### DIFF
--- a/src/configs/lang.ts
+++ b/src/configs/lang.ts
@@ -804,3 +804,8 @@ export const TARGETED_STATUS = translate('TARGETED_STATUS', 'Target Status');
 export const TARGETED = translate('TARGETED', 'Targeted');
 
 export const NOT_TARGETED = translate('NOT_TARGETED', 'Not Targeted');
+
+export const SELECT_JURISDICTION = translate(
+  'SELECT_JURISDICTION',
+  'Please select atleast one jurisdiction'
+);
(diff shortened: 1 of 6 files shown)


src/configs/lang.ts
```diff
@@ -804,3 +804,8 @@ export const TARGETED_STATUS = translate('TARGETED_STATUS', 'Target Status');
 export const TARGETED = translate('TARGETED', 'Targeted');
 
 export const NOT_TARGETED = translate('NOT_TARGETED', 'Not Targeted');
+
+export const SELECT_JURISDICTION = translate(
+  'SELECT_JURISDICTION',
+  'Please select atleast one jurisdiction'
```
moshthepitt, line 810: Please change it to "at least"

src/containers/pages/JurisdictionAssignment/JurisdictionTable/index.tsx
```diff
@@ -330,26 +343,50 @@ const JurisdictionTable = (props: JurisdictionSelectorTableProps) => {
...
+              onClick={onSaveAndActivate}
+              size="xs"
+            >
+              {SAVE_AND_ACTIVATE}
+            </Button>
+          </span>
+        </>
+      )}
+      {!hasSelectedNode && (
+        <>
```
moshthepitt, line 373: Do you think we should use the `React.Fragment` component instead of `<>`?

ciremusyoka:
Screen shots
Screenshot from 2020-08-11 17-01-45
Screenshot from 2020-08-11 17-01-50
Screenshot from 2020-08-11 17-02-03

Issue #1128, opened 2020-08-11
MDA Plans should allow multi jurisdiction selection


moshthepitt:
This is an env (config) change to ensure that MDA-Point and Dynamic-MDA plans allow selecting more than one jurisdiction.

- [x] Ensure the default value of the relevant env var reflects  this
- [x] Ensure deployment inventories reflect this

PR #1134, opened 2020-08-12
Allow multi jurisdiction selection for MDA plans

ciremusyoka:
part of #1128 

Adds `MDA-Point` and `Dynamic-MDA` plan intervention types to the list of plans allowed to have multi jurisdiction selection.


p-netm:
@ciremusyoka please update once the inventories' vars are updated.

Issue #1115, opened 2020-08-05
The save button freezes when one tries to create a new plan RVL-1051


Naima-Bashir:
Trying to set a plan as active during creation causes the Save button to freeze with the text Saving and an error pop's up
![image](https://user-images.githubusercontent.com/61143692/89379230-158fbc00-d6fe-11ea-9277-bf10c9c1c0ba.png)

moshthepitt:
@ciremusyoka please investigate to determine if there are any fixes that can be done on the frontend codebase.  Additionally, raise any issues on the OpenSRP server web repo.

ciremusyoka:
@Naima-Bashir am not able to replicate this bug.
@kelvin-muchiri @p-netm can you please try if you can replicate it.

Naima-Bashir:
Just retested this. The save button works as expected. One can now create new plans and save them.

ciremusyoka:
@moshthepitt i will enable the save button if an error occurs so that a user can try saving again. My assumption is the error will not be from the data we are posting but from API since the save button is only enabled when all fields are filled correctly.

moshthepitt:
@ciremusyoka not sure what you mean, could you rephrase that?

ciremusyoka:
@moshthepitt  Currently if an error occurs when submitting a plan the save plan button remains disabled. My thinking is if an errors occures we enable the save plan button so that a user can try saving the plan again.

PR #1136, opened 2020-08-12
Enable plans submit button after an error

ciremusyoka:
Part of #1115 
Enables submit button once an error occurs when creating or editing a plan.


moshthepitt:
@ciremusyoka is it possible to add a test for this?

moshthepitt:
@ciremusyoka how does this work with respect to https://github.com/onaio/reveal-frontend/pull/1126?

ciremusyoka:
> @ciremusyoka how does this work with respect to #1126?

@moshthepitt I think what we are trying to achieve on this PR is different. Here we want to avoid the submit button freezing (remaining disabled) with `Saving` text in case the API throws an error.

PR #1146, opened 2020-08-14
Display historical index cases above yellow structures

### LINKED ISSUE #953: RVL-1001 - Historical Index Case shows below yellow structure
We recently added the capability to view historical index cases on the Web UI map view. The white crosshair is below structures that are yellow, we would want it to be on top of locations that are yellow.

image
In this screenshot, you can see the yellow crosshair as the active index case confirmation. You can see one white crosshair of historical index case on top of the green point. I have clicked on the yellow point and you can see that the white crosshair is below the yellow structure.

Solution:
- Make sure the yellow crosshairs are at the highest level of the styling so they show on top.
- Make sure to account for the possibility that there may be both a white and yellow index case at the same location, so the yellow (current case confirmation) takes precedence. (put yellow on top of the white crosshairs)

kelvin-muchiri:
Closes #953 

Historical index case symbols were showing below yellow structures. In order to have these symbols show over yellow structures, symbol layers are built as the last layers after fill and line.

This is now how the map renders. The white crosshair is above the yellow structure

Screenshot from 2020-08-14 17-31-10


src/containers/pages/FocusInvestigation/map/active/helpers/utils.tsx
```diff
@@ -342,6 +328,59 @@ export const buildGsLiteLayers = (
...
+
+  if (polygonFeatureCollection && polygonFeatureCollection.features.length) {
+    if (goalIsWithSymbol) {
+      gsLayers.push(
+        <GeoJSONLayer
+          {...symbolLayerTemplate}
+          symbolLayout={{
+            ...symbolLayerTemplate.symbolLayout,
+            ...{ 'icon-image': iconGoal },
+            'icon-size': currentGoal === CASE_CONFIRMATION_GOAL_ID ? 0.045 : 0.03,
```
moshthepitt, line 340: could we have these as named constants?  I see they are re-used below

src/containers/pages/FocusInvestigation/map/active/index.tsx
```diff
@@ -298,12 +300,38 @@ const SingleActiveFIMap = (props: MapSingleFIProps & RouteComponentProps<RoutePa
...
+   */
   const gsLayers = [
     ...jurisdictionLayers,
     ...structureLayers,
     ...historicalIndexLayers,
     ...currentIndexLayers,
     ...otherLayers,
+    ...historicalIndexSymbolLayers,
+    ...currentIndexSymbolLayers,
+    ...otherSymbolLayers,
```
p-netm, line 334: lets have the index symbols take precedence over other symbol layers.

PR #1164, opened 2020-08-18
Update jurisdiction metadata payload

ciremusyoka:
Removes `teamId` and `locationId` keys and makes `providerId` key configurable when uploading jurisdictions metadata.


moshthepitt:
Nice work

PR #1166, opened 2020-08-18
1061 fetch keycloak users

### LINKED ISSUE #1061: Reveal stage Create Practitioner needs to transition from OpenMRS to Keycloak
We transitioned Reveal stage to Keycloak, but the Create Practitioner view still queries the list of users from OpenMRS. We need to get this list from Keycloak, which requires a special permission for that user. We want to retain the same functionality which only displays the list of users who are not already assigned as practitioners.

Screen:
image

KipSigei:
Add `source=Keycloak` param to fetch users from Keycloak API instead of OpenMRS

Fixes #1061


src/components/forms/PractitionerForm/UserIdSelect/index.tsx
```diff
@@ -81,27 +87,28 @@ export const UserIdSelect = (props: Props) => {
...
 
       // modify filter params to point to next page
-      const responseSize = response.results.length;
+      const responseSize = response.length;
       filterParams = {
         ...filterParams,
         start_index: filterParams.start_index + responseSize,
       };
-    } while (thereIsNextPage(response));
+    } while (filterParams.start_index < response.length);
```
p-netm, line 111: I am trying to think about this:
- seems like this will never be true:
```
- assuming there is more pages of data, then responseSize < len(totalData)
- but seems like we are using responseSize to know if there is more data.
- set start_index = 0
- say responseSize = 50
- set start_index = 50
- now start_index < responseSize => 50 < 50 => which is false
```

src/components/forms/PractitionerForm/UserIdSelect/index.tsx
```diff
@@ -81,27 +87,28 @@ export const UserIdSelect = (props: Props) => {
   };
 
   /** Pulls all users data */
-  const loadUsers = async (service: typeof OpenSRPService = OpenSRPService) => {
+  const loadUsers = async () => {
     let filterParams = {
       page_size: USERS_REQUEST_PAGE_SIZE,
```
p-netm, line 92: also please check that the deault `USERS_REQUEST_PAGE_SIZE` is set to a `1000`

src/components/forms/PractitionerForm/UserIdSelect/index.tsx
KipSigei: updated this logic via https://github.com/onaio/reveal-frontend/pull/1166/commits/a01e152355358f8c40a89d8d7d52a059c5e1a34a

src/components/forms/PractitionerForm/UserIdSelect/index.tsx
KipSigei: addressed via https://github.com/onaio/reveal-frontend/pull/1166/commits/5021264b457440efccc9f7e22ce84498b4b5e093

p-netm (approved):
:1st_place_medal:

src/components/forms/PractitionerForm/UserIdSelect/index.tsx
```diff
@@ -80,28 +87,40 @@ export const UserIdSelect = (props: Props) => {
...
+      source: OPENSRP_KEYCLOAK_PARAM,
       start_index: 0,
     };
-    const serve = new service(OPENSRP_USERS_ENDPOINT);
+    const serve = new props.serviceClass(OPENSRP_USERS_ENDPOINT);
     const allUsers = [];
-    let response: UserResponse;
+    let response: User[];
+    const usersCount: number = await fetchUsersCount();
```
moshthepitt, line 109: If `fetchUsersCount` encounters an error, will this break?

src/configs/env.ts
```diff
@@ -313,7 +313,7 @@ export const PRACTITIONER_FORM_NAMESPACE =
   process.env.REACT_APP_PRACTITIONER_FORM_NAMESPACE || '78295ac0-df73-11e9-a54b-dbf5e5b2d668';
 export type PRACTITIONER_FORM_NAMESPACE = typeof PRACTITIONER_FORM_NAMESPACE;
 
-const usersRequestPageSize = process.env.REACT_APP_USERS_REQUEST_PAGE_SIZE || '100';
+const usersRequestPageSize = process.env.REACT_APP_USERS_REQUEST_PAGE_SIZE || '1000';
```
moshthepitt, line 316: Could you add a test that shows this value being used?

src/components/forms/PractitionerForm/UserIdSelect/index.tsx
KipSigei: Refactored this to allow handling of failed requests to `user/count` endpoint via https://github.com/onaio/reveal-frontend/pull/1166/commits/4f8ec26fd253c9f38235913d0f35e043364545c9

src/configs/env.ts
KipSigei: addressed via https://github.com/onaio/reveal-frontend/pull/1166/commits/a27c53cb68556f68dd9da405285e6eaf0f121296

moshthepitt (approved):

moshthepitt (approved):

moshthepitt (approved):

PR #1187, opened 2020-08-24
Update getting of columns in GenericJurisdictionReport

### LINKED ISSUE #1144: Implement a better way to get columns in GenericJurisdictionReport
For jurisdiction reports, right now we are [using "focus area levels"](https://github.com/onaio/reveal-frontend/blob/master/src/containers/pages/GenericJurisdictionReport/index.tsx#L204) to determine which columns to show in the table.

We need to move away from this and instead implement a map/function to achieve the same thing.  Maybe a configurable function is better (with a nice default) so that each report can bring its own function, potentially.

Things to consider:

- [x] It must work for already existing IRS Zambia, IRS Namibia, MDA, etc reports
- [x] You should be able to potentially use different column for every single jurisdiction hierarchy level
- [x] You should be able to use the same column for a range or ranges of jurisdiction hierarchy levels

ciremusyoka:
Closes #1144

Moves the logic of getting columns to a function and adds capability of passing a custom function for getting columns.


src/containers/pages/GenericJurisdictionReport/index.tsx
```diff
@@ -197,14 +199,9 @@ const GenericJurisdictionReport = (
...
 
-  const currLevelData = data.filter(el => el.jurisdiction_parent_id === jurisdictionId);
-
-  let columnsToUse = get(plansTableColumns, jurisdictionColumn, null);
-  if (currLevelData && currLevelData.length > 0) {
-    if (currLevelData[0].jurisdiction_depth >= +focusAreaLevel) {
-      columnsToUse = get(plansTableColumns, focusAreaColumn, null);
-    }
-  }
+  const columnsToUse = customColumnsGet
```
moshthepitt, line 202: Are you avoiding getting this from props to maintain the way it currently works?

ciremusyoka: not really, my thought was we have the way it works as default and if one intends to change that is when a prop is passed to override

moshthepitt: why not have `getColumnsToUse` as the default value of the prop?

moshthepitt (approved):

p-netm (approved):

PR #1194, opened 2020-08-25
1180 plan team assignment cleanup

### LINKED ISSUE #1180: Plan team assignment page clean up
- [x] Add Page title and breadcrumbs to be before the map 
- [x] Remove page title from the part of the page that comes after the map
- [x] The breadcrumbs section that comes after the map should have a white background

NTD-Zambia-Test-Dynamic-MDA-2020-08-13-Reveal-Stage

### LINKED ISSUE #1205: Error in GisidaLite
Testing using this plan:
https://web.reveal-stage.smartregister.org/assignJurisdictions/4fb345b1-3319-5479-b069-e05ca53a33e6

steps to reproduce:
- open the above draft plan from the planning tool.
- drilldown 2 steps down  on the drill down table.
  ราชอาณาจักรไทย > ตาก (63) > 

Expected:
- the page should continue working normally, any errors you might get will be known issues, and should not break the app

Actual:
- the page breaks with a `GisidaLite issues`, see screenshot below:
Screenshot from 2020-08-26 15-31-22

The error suggests we may be breaking this one rule of hooks:
`you ensure that Hooks are called in the same order each time a component render`

KipSigei:
Fixes #1180  and fixes #1205


src/components/page/HeaderBreadcrumb/HeaderBreadcrumb.tsx
```diff
@@ -40,7 +41,9 @@ class HeaderBreadcrumb extends React.Component<BreadCrumbProps, {}> {
 
     return (
       <div>
-        <Breadcrumb className="reveal-breadcrumb">
+        <Breadcrumb
+          className={`reveal-breadcrumb${isPlanAssignmentPage ? ' plans-breadcrumb' : ''}`}
```
moshthepitt, line 45: What if instead of changing `Breadcrumb` we wrap it in some div in the assignment page (where the bg needs to be different), and then use that div to write targeted CSS to change the background color?

moshthepitt (changes requested):

src/components/page/HeaderBreadcrumb/helpers/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+    if (page.url && page.url.trim()) {
+      breadCrumbItem = (
+        <Link to={page.url} key={key}>
+          {page.label}
+        </Link>
+      );
+    } else {
+      breadCrumbItem = page.label;
+    }
+    return <BreadcrumbItem key={key}>{breadCrumbItem}</BreadcrumbItem>;
```
p-netm, line 21: To reduce the possibility of key's collision, the key string should be ideally namespaced

src/components/page/HeaderBreadcrumb/HeaderBreadcrumb.tsx
KipSigei: addressed https://github.com/onaio/reveal-frontend/pull/1194/commits/c0af93a306855cc6eedfaf09cb5e74b76f0c6da0

src/components/page/HeaderBreadcrumb/helpers/index.tsx
KipSigei: addressed https://github.com/onaio/reveal-frontend/pull/1194/commits/b47e4a81a9db95b07f166b4ea7e015f108edd882

src/containers/pages/AssigmentMapWrapper/index.tsx
```diff
@@ -166,14 +166,14 @@ const AssignmentMapWrapper = (props: AssignmentMapWrapperProps) => {
...
 
-  if (!hasValidGeoms) {
-    return <div>{MAP_LOAD_ERROR}</div>;
-  }
-
   if (loading) {
     return <Loading />;
   }
 
+  if (!hasValidGeoms || !structures.length) {
```
moshthepitt, line 173: I don't think we should prevent the map from loading if we have no structures.  For example, it is pretty common in IRS reports to only have the boundary (jurisdiction) but no structures.  In which case I would like the map to load and just show the boundary.

I think we can take this idea further and aim for the map being able to load as long as we have any valid layers.

Does that sound like it could work?

moshthepitt (changes requested):

src/containers/pages/AssigmentMapWrapper/index.tsx
KipSigei: For this case we are geting the boundaries(map bounds and center) from the jurisdiction features. So unless the bounds are pre defined in settings then im not really sure if that would work.

moshthepitt: Where do we get structures in this component?

To answer your question, I made a comment [here](https://github.com/onaio/reveal-frontend/issues/1205#issuecomment-681899700).

KipSigei: Got it, so i have added [this](https://github.com/onaio/reveal-frontend/pull/1194/commits/287308df0f830a874ad9186baa6205a9bb20f230) so that we can show an error message if features/layers are empty

moshthepitt (changes requested):
Health-Facility-Dynamic-MDA-2020-08-27-Reveal

src/components/page/HeaderBreadcrumb/helpers/index.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from 'react';
+import { Link } from 'react-router-dom';
+import { BreadcrumbItem } from 'reactstrap';
+import { BreadCrumbProps, Page } from '../HeaderBreadcrumb';
+
+const LinkList = (props: Partial<BreadCrumbProps>) => {
```
moshthepitt, line 6: Lets add a docstring here

src/containers/pages/PlanAssignment/index.tsx
```diff
@@ -268,8 +276,9 @@ const PlanAssignment = (props: PlanAssignmentProps) => {
 
   return (
     <>
```
moshthepitt, line 278: Lets change this to the React.Fragment component

moshthepitt (commented):
I am having trouble with us moving from defining the breadcrumbs like this:

```tsx
      <Breadcrumb className="reveal-breadcrumb plans-breadcrumb">
        <LinkList pages={pages} />
        <BreadcrumbItem active={true}>{currentPage.label}</BreadcrumbItem>
      </Breadcrumb>
```

Instead of like this:

```tsx
<HeaderBreadcrumb {...breadcrumbProps} className="whatever" />
```

Was the introduction of `LinkList` necessary, and worth us having to write more code to define breadcrumbs?

moshthepitt:
When the map is turned off, or when there is an error that prevents the map from showing, let us not show the bottom breadcrumbs.

NTD-Team-2-Dynamic-MDA-2020-08-27-Reveal

KipSigei:
> I am having trouble with us moving from defining the breadcrumbs like this:
> 
> ```tsx
>       <Breadcrumb className="reveal-breadcrumb plans-breadcrumb">
>         <LinkList pages={pages} />
>         <BreadcrumbItem active={true}>{currentPage.label}</BreadcrumbItem>
>       </Breadcrumb>
> ```
> 
> Instead of like this:
> 
> ```tsx
> <HeaderBreadcrumb {...breadcrumbProps} className="whatever" />
> ```
> 
> Was the introduction of `LinkList` necessary, and worth us having to write more code to define breadcrumbs?

That will involve us adding another prop to HeaderBreadCrumb which i thought was something we are trying to avoid?

moshthepitt:
> That will involve us adding another prop to HeaderBreadCrumb which i thought was something we are trying to avoid?

If it 100% involves another prop then we should avoid it.  

But it seems to me we can leave `HeaderBreadCrumb` as it was and instead just add a css class name when we need the styling to be different.  Can that work?

src/containers/pages/PlanAssignment/helpers/JurisdictionTableView/tests/__snapshots__/index.test.tsx.snap
```diff
@@ -133,43 +133,9 @@ Array [
...
-        }
-      }
-    >
-      Team Assignment
-    </th>
-    <th />
-  </tr>
-</thead>
-`;
+exports[`PlanAssignment/JurisdictionTableView works as expected: Table headers 1`] = `null`;
```
p-netm, line 136: is this intentionally null, g

src/containers/pages/PlanAssignment/helpers/JurisdictionTableView/helpers/utils.tsx
```diff
@@ -0,0 +1,51 @@
+import { ASSIGN_PLANS, HOME } from '../../../../../../configs/lang';
+import { ASSIGN_PLAN_URL, HOME_URL } from '../../../../../../constants';
+import { TreeNode } from '../../../../../../store/ducks/opensrp/hierarchies/types';
+
+/**
+ * Util function to build out pages for header breadcrumb component
+ * @param props - props required to build out pages
+ */
+export const pagesBuilder = (props: any) => {
```
p-netm, line 9: Do we have to use any here

src/containers/pages/PlanAssignment/helpers/JurisdictionTableView/tests/__snapshots__/index.test.tsx.snap
KipSigei: Table headers were moved to a different component, addressed via https://github.com/onaio/reveal-frontend/pull/1194/commits/4fe94d99db40c4867eb98b468f296f8244b5f286

src/containers/pages/PlanAssignment/helpers/JurisdictionTableView/helpers/utils.tsx
KipSigei: addressed via https://github.com/onaio/reveal-frontend/pull/1194/commits/4fe94d99db40c4867eb98b468f296f8244b5f286

src/components/page/HeaderBreadcrumb/HeaderBreadcrumb.tsx
```diff
@@ -12,41 +12,36 @@ export interface Page {
...
-          {linkList}
-          <BreadcrumbItem active={true}>{currentPage.label}</BreadcrumbItem>
-        </Breadcrumb>
-      </div>
-    );
-  }
-}
+  return (
+    <div>
+      <Breadcrumb className={`reveal-breadcrumb${props.className ? ' ' + props.className : ''}`}>
```
moshthepitt, line 39: Nitpick: lets get this prop with the rest while they are being destructured.

KipSigei: addressed https://github.com/onaio/reveal-frontend/pull/1194/commits/078fe9203dd24dc48f8dd1322b308855fc2214a3

moshthepitt (approved):

p-netm (approved):

PR #1221, opened 2020-08-31
1216 audit jurisdiction requests

### LINKED ISSUE #1216: Audit loading of jurisdictions in planning tool
We need to audit the API calls made by the planning tool e.g. https://web.reveal-stage.smartregister.org/assignJurisdictions/3ea2c755-e2b4-5bdd-88aa-c118e5ef44bc.  Looks like it is loading some jurisdictions more than once.

p-netm:
closes #1216 
Manual select Jurisdictions:
- getting tree hook is now dependent on changes to the rootId only, removed plan from dependency array
- selecting existing selections was thus moved to own useEffect that depends on both plan and tree.

### DIFF
--- a/src/containers/pages/JurisdictionAssignment/EntryView/utils.tsx
+++ b/src/containers/pages/JurisdictionAssignment/EntryView/utils.tsx
@@ -121,10 +121,6 @@ export function useGetRootJurisdictionId(
   return rootJurisdictionId;
 }
 
-const defaultCallback = () => {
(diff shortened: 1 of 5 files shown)


moshthepitt (changes requested):
Lets add tests that clear show the API calls being made, in what order, and with which parameters

moshthepitt (approved):

pld (approved):

PR #1234, opened 2020-09-01
Initial commit Nigeria SMC Dashboard

jinsthomas:
Please review for the initial cut for SMC Dashboard to production server


src/containers/pages/SMC/Map/index.tsx
```diff
@@ -0,0 +1,468 @@
...
+  fetchStructures: fetchGenericStructures,
+  focusArea: null,
+  jurisdiction: null,
+  plan: null,
+  service: supersetFetch,
+  structures: null,
+};
+
+/** The SMC Reporting Map component */
+const SMCReportingMap = (props: SMCReportingMapProps & RouteComponentProps<RouteParams>) => {
```
moshthepitt, line 126: Can you confirm that this will still load even if Superset queries fail?

src/store/ducks/generic/structures.ts
```diff
@@ -103,7 +106,7 @@ export const fetchGenericStructures = (
         return structure;
       }
     ),
-    'id'
+    'structure_id'
```
moshthepitt, line 109: Should we really make this change?  Won't it affect things like IRS reporting?

moshthepitt: This is because it is possible to have the same structure id present in different plans.  That is why we have another field, `id`, that is absolutely unique

moshthepitt (changes requested):

src/containers/pages/SMC/Map/index.tsx
moshthepitt: @ciremusyoka ^^

ciremusyoka: @moshthepitt addressed by [this commit](https://github.com/onaio/reveal-frontend/pull/1234/commits/2026844d2ccd16682b6a00ffb433a5f36ed7fdf0)

moshthepitt (approved):

kahummer (approved):

PR #1235, opened 2020-09-02
Fix for SSO-like logout behaviour on signout

p-netm:
close #1235 

- invoke opensrp logout url `logout.do` as a fetch request so that we can append the token.
- we then need to clear keycloak cookies: therefore;
  - redirect to keycloak with a redirect-uri of the apps home route


p-netm:
using authorization code flow:

logoutWorks

moshthepitt (approved):
Nice work

KipSigei (commented):
Could we also test that a growl message is displayed when logout fails?

src/App/tests/logout.test.tsx
```diff
@@ -42,13 +43,74 @@ describe('src/app.logout', () => {
...
+      wrapper.update();
+    });
+
+    // expect(fetch.mock.calls).toEqual();
+
+    // keycloack url was not set
+    expect(hrefMock).not.toHaveBeenCalled();
+
+    // TODO - determine why this are 3 calls instead of 1
+    expect(errorSpy.mock.calls).toEqual([[err], [err], [err]]);
```
p-netm, line 111: @KipSigei I tested that the `displayError` function was called with the error object that  the fetch was rejected with.

I thought `displayError` is tested elswhere to call the growl function, so I did not have to do that.

KipSigei (approved):

PR #1248, opened 2020-09-04
Remove metadata from jurisdiction name when creating setting payload

### LINKED ISSUE #1239: RVL-1126 - Jurisdiction names in jurisdiction metadata has metadata appended to it
- [ ] On the Namibia Production, when a user download the csv file from the "Download Jurisdiction Metadata" section, the jurisdiction names come with ad appended "metadata" addition See screenshot below.
image

p-netm:
close #1239

This fix will not apply for aleady uploaded metadata,


src/components/forms/JurisdictionMetadata/tests/index.test.tsx
```diff
@@ -100,3 +105,58 @@ describe('components/forms/JurisdictionMetadata', () => {
     });
   });
 });
+
+describe('uploads form', () => {
```
moshthepitt, line 109: Sorry if I missed it, but does this test assert that the metadata values (the actual data payload) is as we expect it?

p-netm: yes

moshthepitt (approved):

PR #1256, opened 2020-09-07
Update helper that removes plans with invalid jurisdiction_id values

### LINKED ISSUE #1251: RVL-1133 - Monitoring Tab not loading plans
- [ ] The monitoring tab on the Thailand WebUI Preview environment, is not loading plans. See screenshot below:
image

p-netm:
the mentioned helper should now also:
- filter out plans whose jurisdiction_id value is falsy

close #1251

### DIFF
--- a/src/helpers/utils.tsx
+++ b/src/helpers/utils.tsx
@@ -458,15 +458,15 @@ export function getColor(taskObject: InitialTask): Color {
   }
 }
 
-/** filters out plans whose jurisdiction id is null
+/** filters out plans whose jurisdiction id is null or has a value of 'null'
  * @param {Plan []} plansArray - a list of pre-filtered plans
  *
  * @return {Plan []} - A list of filtered plans
  */
 export const removeNullJurisdictionPlans = (plansArray: Plan[]): Plan[] => {
   return plansArray.filter((plan: Plan) => {
-    const jurisdictionID = plan.jurisdiction_id.toLowerCase().trim();
-    return !jurisdictionID.includes('null');
+    const jurisdictionId = plan.jurisdiction_id?.toLowerCase().trim();
+    return !!jurisdictionId && !jurisdictionId.includes('null');
   });
 };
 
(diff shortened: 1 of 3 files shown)


src/helpers/utils.tsx
```diff
@@ -458,15 +458,15 @@ export function getColor(taskObject: InitialTask): Color {
...
  * @param {Plan []} plansArray - a list of pre-filtered plans
  *
  * @return {Plan []} - A list of filtered plans
  */
 export const removeNullJurisdictionPlans = (plansArray: Plan[]): Plan[] => {
   return plansArray.filter((plan: Plan) => {
-    const jurisdictionID = plan.jurisdiction_id.toLowerCase().trim();
-    return !jurisdictionID.includes('null');
+    const jurisdictionId = plan.jurisdiction_id?.toLowerCase().trim();
+    return !!jurisdictionId && !jurisdictionId.includes('null');
```
moshthepitt, line 469: Is it completely safe to remove `.trim()`?

p-netm: I don't think so. If I remember correctly the values that were originally handled by this looked sth like ~  '  NULL'

moshthepitt: yeah I recall that as well.  Should we add trim back to this code and ensure we have a test that covers the `' NULL'` cases?

ciremusyoka (approved):

src/helpers/utils.tsx
ciremusyoka: The existing function [test](https://github.com/onaio/reveal-frontend/blob/4b8394d538e2cb9c7fe77e792e32849792ea67ef/src/helpers/tests/utils.test.tsx#L326-L330) covers `'  NULL'`  cases ([fixtures](https://github.com/onaio/reveal-frontend/blob/4b8394d538e2cb9c7fe77e792e32849792ea67ef/src/store/ducks/tests/fixtures.ts#L187-L217)).

moshthepitt (approved):

p-netm:
@moshthepitt @ciremusyoka This pr seems to have been overriden by a force push, i.e. after it was merged into the parent branch. going to re-open this

PR #1260, opened 2020-09-08
default plan user filter

### LINKED ISSUE #1172: RVL-1016 - User filter to have by default the username of the user logged in
When logging in to the Web, in the manage plan screens and monitoring, the filter by "username" should be set to default to the user who is logged in.

In the below picture, would be set to "SuperSetTester"
image

p-netm:
close #1172 

plans are filtered by user by default (once one accesses on the pages with userFilter)


moshthepitt:
@p-netm some screenshots, perhaps?

src/components/forms/UserFilter/index.tsx
```diff
@@ -60,6 +63,20 @@ export const BaseUserSelectFilter = (props: BaseUserSelectFilterPropTypes) => {
...
 
+  useEffect(() => {
+    if (!(defaultUserNameValue === SELECT_USERNAME && ENABLE_DEFAULT_PLAN_USER_FILTER)) {
+      return;
+    }
+
+    const { username } = getUser(store.getState());
+    const option = {
+      label: username,
+      value: '',
```
moshthepitt, line 74: Why is the value empty?

p-netm: its supposed to be the id of the user. which we don't have at that point, fortunately we can and are using the `username` to get the plans assigned to a user, so the userId is optional.

moshthepitt: if we are using the username to search, would it make sense for the value to be username?

p-netm: the username is also used as the in react-select's dropdown,

moshthepitt: I mean this eventually creates something similar to HTML options ie `<option value="value">label</option>` right?

So here we'll end up with something like `<option value="">superset-user</option>` - which is weird from a HTML point of view, right?  I am asking is there any harm in having `<option value="superset-user">superset-user</option>`?

Additionally, if as you say we are using the username to filter, does it mean that we filter using the label of the react-select option?

p-netm: >  I am asking is there any harm in having `<option value="superset-user">superset-user</option>`?

I suppose ther is no harm in having the `value` and the `label` being the same value.

moshthepitt: @p-netm could you confirm what is used when we are doing the filtering?  Coz if it is `label` then I would want it changed if possible (in a different issue) because I prefer to stay as close to HTML standards/principles as possible - in this case that would be that for select options, it is the `value` that is important and submitted to be used, and the `label` is just for display - despite this being js-land where we can do whatever we want.

moshthepitt: > I suppose ther is no harm in having the value and the label being the same value.

It is actually pretty common for HTML

ciremusyoka: There no any harm

p-netm: @moshthepitt @ciremusyoka Thanks for the review, I will look at this more closely. for the now, I do not hink it will make it for the release.

p-netm:
userplanFilter

src/components/forms/PractitionerForm/UserIdSelect/index.tsx
```diff
@@ -25,6 +25,10 @@ export interface Props {
   showPractitioners: boolean /** show users that are already mapped to a practitioner */;
   className: string;
   ReactSelectDefaultValue: Option;
+  /** if we are only interested in the userName -> affects how a single option
```
moshthepitt, line 28: Is it possible to just pick one of these to use instead of making it interchangeable?

p-netm: yeah, though in that case we would have the userName as the label,even if we intend to use it as the value

p-netm: > Additionally, if as you say we are using the username to filter, does it mean that we filter using the label of the react-select option

moshthepitt: What do we use the `id` for?

p-netm: its used in the PractionerForm when creating a practitioner @moshthepitt

moshthepitt: @p-netm I mean when doing user filters like here, how is the id used?

p-netm: its not being used for filtering in this useCase, but I would rather pass in a different prop than have different schema in the data passed to the callback, thoughts?

moshthepitt: @p-netm  Could you elaborate on what you mean?  What is causing the differences in schema?

p-netm: hmm,.. stepping back here:
> Is it possible to just pick one of these to use instead of making it interchangeable?
how did mean, I though you referred to how the selected option is passed back to the callback, where in this case, you would call the callback with just the value you need which would be a string

p-netm: by the way; I removed the use of the userId from the filtering functionality, the passed back option has the username as both the value of `label` and `value`

moshthepitt: I meant that if we are not using the id at all when filtering, why do we need to have a mechanism to conditionally use either id or username as value?  Or is it because this is shared between filtering and the practitioner form?

p-netm: ooh ok. yeah, this is a shared component.

moshthepitt (approved):
Nice work

ciremusyoka (approved):

Issue #1262, opened 2020-09-08
Hide Assign Plan and Admin Pages on Thailand Preview


samkanga:
- [x] Hide Assign Plan Page on Thailand Preview

- [x] Hide admin page on Thailand Preview

PR #1263, opened 2020-09-09
Add conditions to when Admin nav is shown

p-netm:
Refactor condition of when Admin nav is shown
relates to #1262


src/components/page/Header/index.tsx
```diff
@@ -238,7 +238,13 @@ export class HeaderComponent extends React.Component<HeaderProps, State> {
                 </UncontrolledDropdown>
               )}
 
-              {(ENABLE_TEAMS || ENABLE_PRACTITIONERS || ENABLE_USERS || ENABLE_MDA_POINT) && (
+              {(ENABLE_TEAMS ||
```
moshthepitt, line 241: Does it make sense to now create a variable to hold this value?

p-netm: yeap

moshthepitt (approved):

ciremusyoka (approved):

PR #1277, opened 2020-09-14
get Dynamic FI plans too in Focus investigation pages

### LINKED ISSUE #1272: Not all plans are shown qwhen filtering Monitor screen by user RVL-1160
Login as bvbd_mhealth
Go to Plan screen - see that there are more than 30 plans assigned, many in 'In progress' state
Go to Monitor screen - see that there is only 1 Routine plan displayed

### LINKED ISSUE #1276: Routine plans created in the WebUI are not showing in Monitor page RVL 1162
2 plans created in the webUi show on the device but not in the Monitor section of the webui

USER bvbd_mhealth 

The plans created are Dynamic FI plans:

Dynamic-FI 2020-09-14 Test Targets 1
Dynamic-FI 2020-09-14 Test Targets 2
image

p-netm:
close #1276 close #1272 
adds a filter to get `Dynamic-FI` plans too in the reportin> activeFI pages

### DIFF
--- a/src/containers/pages/FocusInvestigation/active/index.tsx
+++ b/src/containers/pages/FocusInvestigation/active/index.tsx
@@ -58,7 +58,10 @@ import {
   ROUTINE,
   ROUTINE_QUERY_PARAM,
 } from '../../../../constants';
-import { loadPlansByUserFilter } from '../../../../helpers/dataLoading/plans';
+import {
+  loadPlansByUserFilter,
+  supersetFIPlansParamFilters,
+} from '../../../../helpers/dataLoading/plans';
 import { displayError } from '../../../../helpers/errors';
 import { renderClassificationRow } from '../../../../helpers/indicators';
 import '../../../../helpers/tables.css';
@@ -146,9 +149,7 @@ class ActiveFocusInvestigation extends React.Component<
 
   public componentDidMount() {
     const { userName, fetchPlansActionCreator, supersetService } = this.props;
-    const supersetParams = superset.getFormData(2000, [
-      { comparator: InterventionType.FI, operator: '==', subject: 'plan_intervention_type' },
-    ]);
+    const supersetParams = superset.getFormData(2000, supersetFIPlansParamFilters);
     supersetService(SUPERSET_PLANS_SLICE, supersetParams)
       .then((result: Plan[]) => {
         if (result) {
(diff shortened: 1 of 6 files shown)


src/helpers/dataLoading/plans.ts
```diff
@@ -84,3 +95,13 @@ export async function loadOpenSRPPlan(
       displayError(err);
     });
 }
+
+/** Added to superset requests to specify only FI and Dynamic-Fi plans are required. */
+export const supersetFIPlansParamFilters = [
```
moshthepitt, line 100: Will this work like we need it to?  i.e. will it filter for plans that are either FI **or** DynamicFI?  Or will it attempt to filter for plans that are FI **and** DynamicFi?

In any case, I vote that we switch to using an **IN** filter.  SQL **IN** is usually faster for these kinds of filters i.e. where you are checking if a variable is in a number of values.  An added benefit is that it would potentially make this code much simpler by checking if the intervention type is in enabled plan types without the IFs.

Please see this comment for the available filters: https://github.com/onaio/canopy/issues/228#issuecomment-506699679

moshthepitt (changes requested):

src/helpers/dataLoading/plans.ts
p-netm: > An added benefit is that it would potentially make this code much simpler by checking if the intervention type is in enabled plan types without the IFs

@moshthepitt 
does this mean we can have sth like `in [FI,IRS,MDA,MDA-Point,Dynamic-FI,Dynamic-IRS,Dynamic-MDA]`

moshthepitt: yeah the equivalent SQL will be something like `select ... from {the table} where intervention_type in (FI, DynamicFI)`;

p-netm: `57c3068 `

p-netm: https://github.com/onaio/reveal-frontend/pull/1277/commits/57c30684e6bfef33ecba4dc4ac1b40dfe7f9f828

src/containers/pages/FocusInvestigation/active/index.tsx
```diff
@@ -146,9 +149,7 @@ class ActiveFocusInvestigation extends React.Component<
 
   public componentDidMount() {
     const { userName, fetchPlansActionCreator, supersetService } = this.props;
-    const supersetParams = superset.getFormData(2000, [
-      { comparator: InterventionType.FI, operator: '==', subject: 'plan_intervention_type' },
-    ]);
+    const supersetParams = superset.getFormData(2000, supersetFIPlansParamFilters);
```
ciremusyoka, line 152: Use the `SUPERSET_MAX_RECORDS` env instead of `2000` for number of records to be fetched.

moshthepitt (approved):

ciremusyoka (approved):

PR #1284, opened 2020-09-16
Map fixes

### LINKED ISSUE #1273: Monitor Plan view zooms in whenever an indicator is selected RVL-1163
The expectation is that no zooming occurs 
Go to monitor page- select different types of indicators
Observe how the map view is already zoomed in whenever an indicator is selected
This means that one has to zoom out of each indicator in order to see the full map

### LINKED ISSUE #1274: Index case targets disappear with zoom- RVL 1159
Login to the webUI  as bvbd_mhealth
Navigate to https://web.reveal-th-preview.smartregister.org/focus-investigation/map/9cf6f809-8729-54d3-a526-d000c2985416
See a number of green dots  - as you zoom in and out, the white target indicators appear and disappear randomly

p-netm:
close #1273 close #1274 

Remove default zoom value

Add an onLoad prop that can supply callbacks called when `load` event is emitted

will be using the onLoad to set a zoom based on  a combination of the map's element canvas size and the mapBounds.

Also add icon-allow-overlap to symbo layout property

when symbols collide they will now overlap

also tweaked the sizes of the icons to scale with zoom

### DIFF
--- a/src/containers/pages/FocusInvestigation/map/active/helpers/utils.tsx
+++ b/src/containers/pages/FocusInvestigation/map/active/helpers/utils.tsx
@@ -438,10 +438,14 @@ export const getMapBounds = (jurisdiction: Jurisdiction | null) => {
  * ref: https://github.com/mapbox/mapbox-gl-js/issues/1970#issuecomment-297465871
  */
 export const setMapViewPortZoomFactory = (mapBounds: BoundingBox) => (map: Map) => {
+  const zoomAnimationDuration = 1000;
+  const minZoom = 0;
+  const maxZoom = 20;
+  const mapBoxTileSize = 512;
   const mapEl = map.getCanvas().getBoundingClientRect();
   const mapDim: [number, number] = [mapEl.height, mapEl.width];
-  const newBounds = viewport(mapBounds, mapDim, 0, 20, 512);
-  map.zoomTo(newBounds.zoom, { duration: 1000 });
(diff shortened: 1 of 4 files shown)


src/containers/pages/FocusInvestigation/map/active/helpers/utils.tsx
```diff
@@ -432,6 +434,16 @@ export const getMapBounds = (jurisdiction: Jurisdiction | null) => {
...
 };
 
+/** dynamically get the bounds and zoom for the maps viewport.
+ * ref: https://github.com/mapbox/mapbox-gl-js/issues/1970#issuecomment-297465871
+ */
+export const setMapViewPortZoomFactory = (mapBounds: BoundingBox) => (map: Map) => {
+  const mapEl = map.getCanvas().getBoundingClientRect();
+  const mapDim: [number, number] = [mapEl.height, mapEl.width];
+  const newBounds = viewport(mapBounds, mapDim, 0, 20, 512);
+  map.zoomTo(newBounds.zoom, { duration: 1000 });
```
moshthepitt, line 444: Could we turn these numbers into named constants?  Don't think we should move to the constants module (yet) but lets name them for easier reading of the coe

p-netm:
Screenshot from 2020-09-16 17-41-08
with overlapping symbols.

moshthepitt (approved):

ciremusyoka (approved):

Issue #1298, opened 2020-09-21
RVL-1185 - No Case Triggered plans are showing in the WebUI


samkanga:
- [ ] On Thailand preview WebUI Case triggered Plans are not showing up under the Monitor Page

PR #1305, opened 2020-09-21
Decouple configuration for displayed plans and plans that can be created.

ciremusyoka:
Fixes: #1298 
Use different envs to control plans to be displayed and plans which can be created:
- REACT_APP_DISPLAYED_PLAN_TYPES
- REACT_APP_PLAN_TYPES_ALLOWED_TO_CREATE


src/components/forms/PlanForm/helpers.ts
```diff
@@ -776,7 +777,15 @@ export function getGoalUnitFromActionCode(actionCode: PlanActionCodesType): Goal
...
 export const isPlanTypeEnabled = (planType: InterventionType): boolean =>
-  ENABLED_PLAN_TYPES.includes(planType);
+  DISPLAYED_PLAN_TYPES.includes(planType);
+
+/**
+ * Check if plan type should be created
+ * @param {InterventionType} planType - plan type
+ * @param {boolean} isEditMode - are we editing or creating a plan
+ */
+export const shouldPlanTypeBeCreated = (planType: InterventionType, isEditMode: boolean): boolean =>
```
moshthepitt, line 787: Maybe we should rename this?  Seems like it is not accurate? i.e. you cannot create plans in edit mode.  So this means something other than "create"

moshthepitt (approved):

pld (approved):

PR #1309, opened 2020-09-23
Disable high level jurisdictions team assignment

p-netm:
Disable high level jurisdictions team assignment

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -135,9 +135,6 @@ REACT_APP_PLAN_TYPES_WITH_MULTI_JURISDICTIONS='IRS,Dynamic-IRS,MDA-Point,Dynamic
 # options displayed when downloading jurisdiction metadata
 REACT_APP_ENABLED_JURISDICTION_METADATA_IDENTIFIER_OPTIONS=POPULATION,RISK,COVERAGE,TARGET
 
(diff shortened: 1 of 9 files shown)


moshthepitt:
@p-netm how do you feel about leaving the code "as is" and ensuring this `REACT_APP_SHOW_TEAM_ASSIGN_ON_OPERATIONAL_AREAS_ONLY` is turned to the right value?

p-netm:
> @p-netm how do you feel about leaving the code "as is" and ensuring this `REACT_APP_SHOW_TEAM_ASSIGN_ON_OPERATIONAL_AREAS_ONLY` is turned to the right value?

depends on if the intention was to still have `assignments at non-leaf nodes levels` behaviour configurable for certain envirionments.

moshthepitt:
signments at non-leaf nodes levels isn't expected to happen at any of our environments.  But I wonder whether it is something that we should keep because there is a chance that we may want it in the future, or some other non-reveal implementation may want it

p-netm:
> signments at non-leaf nodes levels isn't expected to happen at any of our environments. But I wonder whether it is something that we should keep because there is a chance that we may want it in the future, or some other non-reveal implementation may want it

ok, closing this. will switch the env everywhere to false

PR #1319, opened 2020-09-30
Resolve all promise as a single promise

### LINKED ISSUE #1318: Show loading status in table
[This page](https://web.reveal-zm.smartregister.org/intervention/irs/report/d40c1227-644b-55d6-b3f3-bed42380322f) takes a while to actually load.  But the entire time shows "No Data".  Lets fix this by adding a loading indicator.

p-netm:
close #1318 

use promise.all to resolve the superset calls <del>as a single promise</del>

### DIFF
--- a/src/containers/pages/GenericJurisdictionReport/index.tsx
+++ b/src/containers/pages/GenericJurisdictionReport/index.tsx
@@ -141,7 +141,7 @@ const GenericJurisdictionReport = (
       });
       await Promise.all(allPromises);
     } catch (e) {
-      // todo - handle error https://github.com/onaio/reveal-frontend/issues/300
+      displayError(e);
     } finally {
       setLoading(false);
     }
--- a/src/containers/pages/GenericJurisdictionReport/index.tsx
+++ b/src/containers/pages/GenericJurisdictionReport/index.tsx
@@ -114,12 +114,16 @@ const GenericJurisdictionReport = (
           { comparator: planId, operator: '==', subject: 'plan_id' },
         ]);
       }
+      const allPromises: Array<Promise<unknown>> = [];
 
-      await service(reportingPlanSlice, fetchPlansParams).then((result: GenericPlan[]) => {
-        fetchPlans(result);
-      });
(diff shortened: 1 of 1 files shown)


moshthepitt (commented):
Possible to get a test for this?

p-netm:
> Possible to get a test for this?

@moshthepitt this might require some looking up not sure if maybe you or @ciremusyoka  have ideas around this. 
so to test this I am thinking I would probably need to find a way to resolve the promises at different times, or rather in intervals, not sure `jest-fetch-mock` does this, or maybe how to work it so that it does. currently when we resolve promises we do so in a wholesale-kind of manner.

moshthepitt:
@p-netm we shouldn't test that the promise works/resolves.  We should limit the test to ensuring that we have a promise that is as expected.  Does it make sense?

p-netm:
> @p-netm we shouldn't test that the promise works/resolves. We should limit the test to ensuring that we have a promise that is as expected. Does it make sense?

@moshthepitt could you say that in another way.

moshthepitt:
> @moshthepitt could you say that in another way.

@p-netm I mean that the test would not be trying to prove that Javascript promises work as expected.  Instead it is trying to prove to us that our code creates a promise in the way that we expect.

p-netm:
@moshthepitt there is an existing test that checks what calls are made to superset [here](https://github.com/onaio/reveal-frontend/blob/0eb0bfcf19414a850903345709ac3deeedbe4bfa/src/containers/pages/IRS/JurisdictionsReport/tests/index.test.tsx#L158), not sure if this suffices,  its still passes meaning there was no regression.

The change in behaviour is that;
 I think previously we were not awaiting on all promises before calling the `finally` block.
now, we run all all promises at the same place and await that they all resolve( or dont) before getting to the `finally` bock.

so I did not make any changes to how the Promises are constructed, just where and how they are awaited.

moshthepitt:
@p-netm cool, I think that covers it

moshthepitt (approved):

ciremusyoka (approved):

PR #1339, opened 2020-10-08
Update on teams and task generation statuses on plan

ciremusyoka:
Fixes: https://github.com/onaio/canopy/issues/1551
Adds the following:
- Configuration to ignore taskGenerationStatus when creating plans
- Don't update taskGenerationStatus on plan update
- Don't update teamAssignmentStatus on plan update


src/containers/pages/InterventionPlan/UpdatePlan/tests/fixtures.ts
```diff
@@ -83,6 +84,7 @@ export const updatePlanFormProps = {
     start: parseISO('2019-07-10T00:00:00.000Z'),
     status: 'active',
     taskGenerationStatus: 'False',
+    teamAssignmentStatus: '',
```
moshthepitt, line 87: What are the possible values of this?

ciremusyoka: I don't think we need them because we are only updating based on the values that we get from the server that is why have used `string`.

moshthepitt: Understood.  But what are the possible values of this for plan objects in general?

kelvin-muchiri (dismissed):

src/containers/pages/InterventionPlan/UpdatePlan/tests/fixtures.ts
ciremusyoka: `False` and `True`

moshthepitt:
@ciremusyoka I believe that the desired way this works is:

1. When creating plans, we should have an env that controls what value we set for this.
2. When creating plans and the env is not set, this should be completely not set and absent in the plan object JSON
3. When updating plans, we should keep whatever value was set without changing it at all
4. When updating plans that do not have this, we should do nothing.  We should not add it to the plan at all.

Is the above correct?

Could you please point me to tests that handle each of the four scenarios above?

src/components/forms/PlanForm/tests/helpers.test.ts
```diff
@@ -131,6 +137,62 @@ describe('containers/forms/PlanForm/helpers', () => {
...
+    expect(generatePlanDefinition(planFormValues3 as PlanFormFields, null, true)).toEqual(
+      expectedDynamicPlan
+    );
+
+    // on edit when taskGenerationStatus is not present
+    const { taskGenerationStatus, ...planFormValues3Copy } = planFormValues3;
+    expect(generatePlanDefinition(planFormValues3Copy as PlanFormFields, null, true)).toEqual(
+      expectedDynamicPlanCopy
+    );
+  });
```
ciremusyoka, line 194: > @ciremusyoka I believe that the desired way this works is:
> 
> 1. When creating plans, we should have an env that controls what value we set for this.
> 2. When creating plans and the env is not set, this should be completely not set and absent in the plan object JSON
> .....

@moshthepitt this tests.

src/containers/pages/InterventionPlan/UpdatePlan/tests/dynamicFi.test.tsx
```diff
@@ -77,6 +77,7 @@ describe('components/InterventionPlan/UpdatePlan', () => {
       { code: 'fiStatus', valueCodableConcept: 'A1' },
       { code: 'fiReason', valueCodableConcept: 'Routine' },
       { code: 'taskGenerationStatus', valueCodableConcept: 'True' },
+      { code: 'teamAssignmentStatus', valueCodableConcept: 'True' },
```
ciremusyoka, line 80: > @ciremusyoka I believe that the desired way this works is:
> 
> 1. When creating plans, we should have an env that controls what value we set for this.
> 2. When creating plans and the env is not set, this should be completely not set and absent in the plan object JSON
> .....

@moshthepitt and also this tests.

src/components/forms/PlanForm/helpers.ts
```diff
@@ -757,6 +772,11 @@ export function getPlanFormValues(planObject: PlanDefinition): PlanFormFields {
       ? taskGenerationStatuses[2]
       : taskGenerationStatuses[1];
 
+  const teamAssignmentStatus: string =
+    teamAssignmentStatusContext.length > 0
+      ? taskGenerationStatusContext[0].valueCodableConcept
+      : '';
```
moshthepitt, line 778: 1. Should this be `undefined`?
2. If this is not set (i.e. it is ''), what happens to the plan object?  Does it end up with this useContext value being ''?

ciremusyoka: Am using `formValue.teamAssignmentStatus && formValue.teamAssignmentStatus.trim()` to not set for empty strings

moshthepitt (approved):

src/configs/env.ts
```diff
@@ -459,7 +459,7 @@ export const ENABLE_DEFAULT_PLAN_USER_FILTER =
   process.env.REACT_APP_ENABLE_DEFAULT_PLAN_USER_FILTER === 'true';
 export type ENABLE_DEFAULT_PLAN_USER_FILTER = typeof ENABLE_DEFAULT_PLAN_USER_FILTER;
 
-export const TASK_GENERATION_STATUS = process.env.REACT_APP_TASK_GENERATION_STATUS;
+export const TASK_GENERATION_STATUS = process.env.REACT_APP_TASK_GENERATION_STATUS || 'ignore';
```
pld, line 462: is setting as `ignore` here a convention?

pld: if so, can we use the constant here

pld (changes requested):
also unclear on the resolution of @moshthepitt's comments

pld:
> also unclear on the resolution of @moshthepitt's comments

nevermind, these were revolved before I refreshed the pr

src/configs/env.ts
ciremusyoka: setting ignore is not a convention.

pld (approved):

Issue #1250, opened 2020-09-07
RVL-1002 - Active FI Map tooltip


moshthepitt:
Render a tool tip upon click of the index case point, include: Case number, Case Classification, and Date of Diagnosis

moshthepitt:
Related https://github.com/onaio/reveal-frontend/issues/930

p-netm:
Elsewhere where we have similar functionality: `PlanForm > CaseDetails`.
we get extract the eventId from the plan by looking for the `opensrpEventId` `useContext` object, we then make a get request to opensrp `event/findBy` endpoint.

for this useCase however; we are getting the plans from superset, and they do not have the eventIds

moshthepitt:
The best solution would be to refactor the Canopy views used for Thailand to include the missing information.

mberg:
After 1023 and 1024

lincmba:
@moshthepitt 
Only events of `'eventType': 'Case Details'` have the fields `case_classification`, `case_number` and `investigtion_date` in the `raw_events` table. However, these events, unlike other events don't have a `task_id` or any other foreign key that would link to a `plan`, `goal` or `goal_target`. 

It's therefore a challenge to have the these fields in the `task_structures_materialized_view` . 
Could the events `full_json` be updated with task_id? Or would it be okay to have an independent query for this.

Sample event full_json of eventType `Case Details`:
```
json
{'id': '7c4ec36f-391a-46c8-ac3f-4fc6288aeee7',
 'obs': [],
 'type': 'Event',
 'teamId': 'bd2efe2b-7dcf-4d48-b2ac-27fced048713',
 'details': {'age': '27',
  'species': 'M',
  'surname': 'Doe',
  'focus_id': '3951',
  'first_name': 'Holly',
  'focus_name': 'Akros_1',
  'case_number': '151311000001089190221104722501',
  'family_name': 'Holly',
  'focus_reason': 'Investigation',
  'focus_status': 'B1',
  'house_number': '',
  'ep1_create_date': '2019-09-06T08:41:49.237+0000',
  'ep3_create_date': '2019-09-06T08:41:49.237+0000',
  'investigtion_date': '2019-09-06T00:00:00.000+0000',
  'case_classification': 'A'},
 'version': 1557860282617,
 'duration': 0,
 'revision': 'v3',
 'eventDate': '2019-09-06T02:00:00.000+02:00',
 'eventType': 'Case Details',
[...]

PR #1357, opened 2020-10-14
Add details on index cases map tool tip

ciremusyoka:
Part of #1250
 Includes Case number, Case Classification, and Date of Diagnosis details on index cases map tool tip


moshthepitt:
@ciremusyoka lets get a screenshot if possible please

ciremusyoka:
Screenshot
map-tool-tip

src/store/ducks/opensrp/indexCasesDetails/index.ts
```diff
@@ -0,0 +1,154 @@
+import {
```
moshthepitt, line 1: I think the intention was to have the opensrp directory include only ducks that have entities that come from the OpenSRP API.  I would move this into a "data-warehouse" directory.  We may need to audit the opensrp ducks directory to see what we else we should move - could you create an issue for that please?

ciremusyoka: Created #1358

moshthepitt (approved):

pld (approved):

PR #1369, opened 2020-10-16
Remove default task generation status

ciremusyoka:
Fixes task generation status being given disabled value by default when editing.
Ensures task generation status is only added to plans which have the status and not those which didn't have the status.

### DIFF
--- a/src/components/forms/PlanForm/helpers.ts
+++ b/src/components/forms/PlanForm/helpers.ts
@@ -57,7 +57,6 @@ import {
   FAMILY_REGISTRATION_ACTIVITY_CODE,
   FI_REASON_CODE,
   FI_STATUS_CODE,
-  IGNORE,
   INTERVENTION_TYPE_CODE,
   IRS_ACTIVITY_CODE,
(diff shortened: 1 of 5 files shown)


src/components/forms/PlanForm/helpers.ts
```diff
@@ -769,7 +773,7 @@ export function getPlanFormValues(planObject: PlanDefinition): PlanFormFields {
     taskGenerationStatusContext.length > 0
       ? (taskGenerationStatusContext[0].valueCodableConcept as taskGenerationStatusType)
       : isDynamicPlan(planObject)
-      ? taskGenerationStatuses[2]
+      ? taskGenerationStatuses[4]
       : taskGenerationStatuses[1];
```
moshthepitt, line 777: Lets use a map instead of these indices.  It's easier to read that way.

moshthepitt (approved):
Nice

src/components/forms/PlanForm/helpers.ts
craigappl: + 1 on this

moshthepitt (approved):

PR #1382, opened 2020-10-29
Show map errors

### LINKED ISSUE #1210: Show map error messages
When an error occurs that causes the map on the planning tool or team assignment page to not load, we currently show an error on the page in a div.

In addition to the above, we should [display a growl message](https://github.com/onaio/reveal-frontend/blob/f952af04dea65437efb5962e078cd5b67f1f8562/src/helpers/errors.ts#L18) with the specific problem that caused the map to not load.

ciremusyoka:
Closes #1210 
Display specific errors causing a map not to load.


src/configs/lang.ts
```diff
@@ -898,3 +898,17 @@ export const PLAN_CHANGES_HAVE_NOT_BEEN_SAVED = translate(
...
+);
+export const PLAN_NOT_FOUND = translate('PLAN_NOT_FOUND', 'Map failed to load: plan not found');
+export const JURISDICTION_NOT_FOUND = translate(
+  'JURISDICTION_NOT_FOUND',
+  'Map failed to load: Jurisdiction not found'
+);
+export const FOCUS_AREA_NOT_FOUND = translate(
+  'FOCUS_AREA_NOT_FOUND',
+  'Map failed to load: no focus area found'
+);
```
moshthepitt, line 914: What is the difference between these two?

ciremusyoka: My bad, I think they both refer to same thing. will drop the focus area in favour of jurisdiction.

ciremusyoka: on second thought since we use different slices for jurisdiction and focus areas if both fail we may have same error being flashed twice when we use the `JURISDICTION_NOT_FOUND` to represent both.

moshthepitt: Yeah I think they are the same thing to the web user

moshthepitt (approved):

p-netm (approved):

PR #1423, opened 2020-11-24
Update practitioner API calls when adding practitioner.

### LINKED ISSUE #1381: RVL-1301 Update Add practitioner list to account for practitioner pagination
It looks like the dropdown list that allows a user to add a practitioner does a comparison between the OpenSRP server and the list of users that are returned from Keycloak. We need to update this to account for pagination that will be added to the practitioner endpoint in https://github.com/OpenSRP/opensrp-server-web/issues/591. We also need to make sure we don't get into pagination issues from the Keycloak API.

This work can't be done until the OpenSRP server web work cited above is complete and deployed for development.

### LINKED ISSUE #1422: Fix FI Map text
There seems to be an unwanted apostrophe.

unwanted

ciremusyoka:
Updates how API call for practitioners is made to cater for the pagination functionality added on the endpoint.
The number of records fetched at once are controlled by an env, `REACT_APP_PRACTITIONER_REQUEST_PAGE_SIZE`.

Fix #1381   
Fix #1422


src/components/forms/PractitionerForm/UserIdSelect/index.tsx
```diff
@@ -132,6 +132,37 @@ export const UserIdSelect = (props: Props) => {
...
+    let filterParams = {
+      pageNumber: 1,
+      pageSize: PRACTITIONER_REQUEST_PAGE_SIZE,
+    };
+    const allPractitioners = [];
+    let response: Practitioner[];
+    let serverError = false;
+    const service = new props.serviceClass(OPENSRP_PRACTITIONER_ENDPOINT);
+
+    do {
```
moshthepitt, line 146: I wonder if there is a way to do this with chained promises to avoid awaiting?  Some recursion would need to be involved, I think.  Do you think it is worth it to pursue that?

src/components/forms/PractitionerForm/UserIdSelect/tests/index.test.tsx
```diff
@@ -12,6 +12,7 @@ import { practitioners, sortedUsers, users } from './fixtures';
 // tslint:disable-next-line: no-var-requires
 const fetch = require('jest-fetch-mock');
 jest.mock('../../../../../configs/env.ts', () => ({
+  PRACTITIONER_REQUEST_PAGE_SIZE: 1000,
```
moshthepitt, line 15: Should we raise the default for practitioners to be more than 1000?  We want to strike a balance between the number of requests we make, and the size of the response we receive from OpenSRP. I think if we can figure out what page length results in about 100Kb of practitioners then we should use that page size. The figure of 100Kb is loosely based on [this discussion](https://www.postgresql.org/message-id/4D2F2C71.8080805%40dndg.it) - not scientific.

moshthepitt: Just to make it clear, not suggesting that you try and determine the number of practitioners in 100Kb programmatically.  Just do some manual testing with the API.

src/components/forms/PractitionerForm/UserIdSelect/index.tsx
```diff
@@ -132,6 +132,37 @@ export const UserIdSelect = (props: Props) => {
     return allUsers;
   };
 
+  /** pull all practioners */
+  const getAllPractitioners = async () => {
```
p-netm, line 136: should we move this to https://github.com/onaio/reveal-frontend/blob/master/src/containers/pages/PractitionerViews/helpers/serviceHooks.ts

src/components/forms/PractitionerForm/UserIdSelect/index.tsx
ciremusyoka: I think it can be possible if we had the endpoint for returning total counts ready. Currently i have no way of knowing when to stop unless i wait for the results.

src/components/forms/PractitionerForm/UserIdSelect/tests/index.test.tsx
ciremusyoka: 1000 records are more than 100kb. 100kb is about 650 records

src/components/forms/PractitionerForm/UserIdSelect/index.tsx
moshthepitt: I think its probably possible to construct the promise chain and in the `then` block of each promise check if there are any records returned.  You'd then stop chaining when there are no records returned.

The advantage here is that everything is async and we have the store getting populated as soon as the first promise resolves.

src/components/forms/PractitionerForm/UserIdSelect/index.tsx
```diff
@@ -34,6 +35,7 @@ export interface Props {
 /** default props for UserIdSelect component */
 export const defaultProps = {
   ReactSelectDefaultValue: { label: '', value: '' },
+  allPractitioners: [] as any,
```
moshthepitt, line 38: Does it have to be any?

src/containers/pages/PractitionerViews/helpers/serviceHooks.ts
```diff
@@ -40,3 +41,45 @@ export const loadPractitioner = async (
...
+      return service
+        .list(filterParamss)
+        .then((res: Practitioner[]) => {
+          if (res && res.length) {
+            store.dispatch(fetchPractitioners(res));
+          }
+          if (getAll) {
+            const newPage = ++currentPage;
+            // tslint:disable-next-line: no-floating-promises
+            getPractitionersByPage(newPage, res);
```
moshthepitt, line 75: What will happen if an exception occurs here?

ciremusyoka: will through an error.

moshthepitt: will the error be caught nicely?

src/components/forms/PractitionerForm/UserIdSelect/index.tsx
ciremusyoka: No. Made it more specific

src/containers/pages/PractitionerViews/helpers/serviceHooks.ts
ciremusyoka: Yeah, this [line](https://github.com/onaio/reveal-frontend/blob/54d7912fad75c2d88f897b4fec01d8336941941e/src/containers/pages/PractitionerViews/helpers/serviceHooks.ts#L78) catches any exceptions.

moshthepitt (approved):

src/components/forms/PractitionerForm/UserIdSelect/index.tsx
```diff
@@ -138,25 +140,14 @@ export const UserIdSelect = (props: Props) => {
    */
   const loadData = async () => {
     const allUsers = await loadUsers();
-    if (props.showPractitioners && isMounted.current) {
-      // setState with all unfiltered users if component is mounted
-      setUsers(allUsers);
-      setSelectIsLoading(false);
-    }
+    // setState with all unfiltered users if component is mounted
```
p-netm, line 143: does this comment still apply?

p-netm (approved):

Issue #1442, opened 2020-12-07
Don't show `NaN` on planning tool


moshthepitt:
![nonan](https://user-images.githubusercontent.com/372073/101330215-6584b800-3883-11eb-86e7-1b674ba9748a.png)

This happens [here](https://web.reveal-stage.smartregister.org/auoSelectJurisdictions/d36e06b3-c556-5f86-801b-d0a686114bcb/ecfbf048-fb7a-47d8-a12b-61bf5d2a6e7b) - it should always show an integer

Issue #1441, opened 2020-12-04
Uploaded Risk and Structures Data not available When Creating plans through Planning Tool


samkanga:
Akros are doing tests on reveal stage on the upload of risk files and structures from a csv. The file is downloading and uploding successfully with the values entered into the risk and target columns of the the csv. However, when they create a plan through the planning tool, the following are the observations:

1. There structures count on the risk slider view remain at zero when the slider is moved from 0 to 100%.
2. On the map view a message appears to the right of the map reading `No descendant jurisdictions are targeted`
3. On the table below the map the risk and structures columns are all empty and do not show any of the uploaded values.

The expectation was that with the slider, one would be able to narrow down to the locations that match the uploaded risk values and be able to select them or deselect them for inclusion in an intervention.

moshthepitt:
@samkanga is it possible to get an example of a plan for which this happens?  That is crucial missing information

samkanga:
@moshthepitt you can use the plan [here](https://web.reveal-stage.smartregister.org/auoSelectJurisdictions/d36e06b3-c556-5f86-801b-d0a686114bcb/ecfbf048-fb7a-47d8-a12b-61bf5d2a6e7b).

moshthepitt:
Blocked by https://github.com/OpenSRP/opensrp-server-web/issues/655

ciremusyoka:
Also depends on https://github.com/OpenSRP/opensrp-server-web/issues/705

ciremusyoka:
The structure counts are changing when the slider is moved. 
> Also depends on OpenSRP/opensrp-server-web#705

Will close this ticket in favour of the above OpenSRP ticket

PR #1443, opened 2020-12-07
Fix planning tool auto-selection

moshthepitt:
This PR addresses some issues on the planning tool:

- Ensures that the planning tool can work even when the API is broken:
    - it works even when the jurisdiction hierarchy API does not return a `structureCount` property
    - it now works even when the jurisdiction metadata API does not return valid `value` properties
- Adds information to the risk slider to show the number of selected jurisdictions.  I think that this is necessary so that the
  user can see how many jurisdictions are selected.  It was possible in the past to move the slider and have the structure count
  not change, but the jurisdiction count would be changing.
- Use `>=` instead of `>` when evaluating if a jurisdiction meets criteria for selection based on risk.  This would cater for jurisdictions whose values are 0 or 100, which would otherwise not always be selected when they should. 

Fix: #1442 
Part of fixing: #1441

add-jurisdiction-count


src/store/ducks/opensrp/hierarchies/index.tsx
```diff
@@ -541,7 +541,7 @@ export const getStructuresCount = () =>
     const allSelectedLeafNodes = computeSelectedNodes(tree, true, metaByJurisdiction);
     const reducingFn = (acc: number, value: number) => acc + value;
     return allSelectedLeafNodes
-      .map(node => node.model.node.attributes.structureCount)
+      .map(node => node.model.node.attributes.structureCount || 0)
```
moshthepitt, line 544: Ensure that we don't fail when `structureCount` is not defined.

src/containers/pages/JurisdictionAssignment/helpers/Slider/index.tsx
```diff
@@ -131,6 +136,13 @@ export const JurisdictionSelectionsSlider = (props: Props) => {
               <p>{structuresCount}</p>
             </Col>
           </Row>
+          <Row className="auto-target-row mb-3">
+            <Col xs="12" md="12" className="info-section py-0 px-5">
+              <p className="text-center mt-3">
+                {jurisdictionsCount} {JURISDICTIONS_SELECTED}
+              </p>
+            </Col>
+          </Row>
```
moshthepitt, line 145: Added a way to display to the user how many jurisdictions are selected.  See commit message: https://github.com/onaio/reveal-frontend/pull/1443/commits/73d347ad35de49bc80f38f6511d046641f28fd03

src/containers/pages/JurisdictionAssignment/helpers/Slider/tests/slider.test.tsx
```diff
@@ -39,31 +67,111 @@ describe('JurisdictionAssignment/Slider', () => {
...
+    // simulate another change to deselect all jurisdictions
+    /** we really should not be doing this, but I am currently unable to simulate
+     * the change event
+     */
+    act(() => {
+      (wrapper.find('InputRange').props() as any).onChange(81);
+    });
+    expect(wrapper.find('.risk-label').text()).toMatchInlineSnapshot(`"81%"`);
     (wrapper.find('InputRange').props() as any).onChangeComplete(81);
+    // we should now have two jurisdictions selected
```
p-netm, line 125: I think this should read `no selected jurisdictions`

p-netm (commented):
:+1:

src/containers/pages/JurisdictionAssignment/helpers/Slider/tests/slider.test.tsx
moshthepitt: thanks for catching that @p-netm - now fixed

p-netm (approved):

ciremusyoka (approved):

PR #1452, opened 2020-12-11
Fix endless loading on irs map

### LINKED ISSUE #1451: IRS map loads forever when Superset error is encountered
e.g. https://web.reveal-zm.smartregister.org/intervention/irs/report/4cbe10b9-231e-5084-9646-0a923dcfa0da/04f4a5b9-31a5-48e9-bcd0-c389be95c35d/map

ciremusyoka:
This pr fixes endless loading on the map component when there is an error on the points to be loaded. In case of an error the error page will be loaded.

Fixes #1451

### DIFF
--- a/src/components/GisidaLite/helpers.tsx
+++ b/src/components/GisidaLite/helpers.tsx
@@ -1,6 +1,10 @@
+import { viewport } from '@mapbox/geo-viewport';
+import GeojsonExtent from '@mapbox/geojson-extent';
 import { center as turfCenter, FeatureCollection as TurfFeatureCollection } from '@turf/turf';
 import { Style } from 'mapbox-gl';
 import { DIGITAL_GLOBE_CONNECT_ID } from '../../configs/env';
+import { StructureFeatureCollection } from '../../store/ducks/generic/structures';
+import { Jurisdiction } from '../../store/ducks/jurisdictions';
 
 /** Get the centre coordinates from a feature collection */
 export const getCenter = (fc: TurfFeatureCollection): [number, number] | undefined => {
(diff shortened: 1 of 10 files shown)


moshthepitt (changes requested):
Could we ensure, by way of a test, that the map will load if the IRS jurisdiction is present but the IRS structures are missing or their request to Superset errors out?

src/components/GisidaLite/helpers.tsx
```diff
@@ -90,3 +94,34 @@ export const gsLiteStyle: Style | string = DIGITAL_GLOBE_CONNECT_ID
...
+  let zoom;
+  if (structures?.features?.length) {
+    mapBounds = GeojsonExtent(structures);
+  }
+  if (!mapBounds && jurisdiction?.geojson) {
+    mapBounds = GeojsonExtent(jurisdiction.geojson);
+  }
+  // get map zoom and center values
+  if (mapBounds) {
+    const centerAndZoom = viewport(mapBounds, [600, 400]);
```
moshthepitt, line 118: could we make these magic numbers params in the function? Thinking of this from an [open/closed](https://stackify.com/solid-design-open-closed-principle/) perspective.

src/components/GisidaLite/helpers.tsx
```diff
@@ -90,3 +94,34 @@ export const gsLiteStyle: Style | string = DIGITAL_GLOBE_CONNECT_ID
...
+) => {
+  let mapCenter;
+  let mapBounds;
+  let zoom;
+  if (structures?.features?.length) {
+    mapBounds = GeojsonExtent(structures);
+  }
+  if (!mapBounds && jurisdiction?.geojson) {
+    mapBounds = GeojsonExtent(jurisdiction.geojson);
+  }
```
moshthepitt, line 115: This means that if both the jurisdiction and the structures are defined then we will use the jurisdiction to get the center and zoom.  Could we make it the other way round (which is how it is currently).  i.e. that the structures will override the jurisdiction if both exist

src/components/GisidaLite/helpers.tsx
```diff
@@ -90,3 +94,34 @@ export const gsLiteStyle: Style | string = DIGITAL_GLOBE_CONNECT_ID
...
+  // get map zoom and center values
+  if (mapBounds) {
+    const centerAndZoom = viewport(mapBounds, [600, 400]);
+    mapCenter = centerAndZoom.center;
+    zoom = centerAndZoom.zoom;
+  }
+  return {
```
moshthepitt, line 122: what will you return if you have neither structures nor a jurisdiction?  should we make the jurisdiction mandatory for this function?

moshthepitt (changes requested):

src/components/GisidaLite/helpers.tsx
ciremusyoka: we will only use jurisdiction if structures are not defined. That is why on line 113 am first checking if `mapBounds`  is defined.

src/components/GisidaLite/helpers.tsx
ciremusyoka: When we have no structures or jurisdiction all those value will be returned as undefined. 
Because we are using jurisdiction geojson which is optional to generate the map bounds even if we make jurisdiction mandatory and geojson isn't we will have same results.

moshthepitt (approved):

kelvin-muchiri (approved):

PR #1502, opened 2021-02-19
Export student list at all levels

### LINKED ISSUE #1491: Allow user to export student list at any level of the location hierarchy
The "Export Student List" feature is accessed by navigating to "admin" --> "clients." 

Right now, when a user tries to download the student list, they are forced to click all the way through the location hierarchy in the dropdown to choose only the school level. 

Screen Shot 2021-02-09 at 8 17 58 AM

The expected behavior is for the user to be able to select any level of the hierarchy in the dropdown, whether it is a region, a district, or a school. If they want to be able to export all of the children in their campaign, we don't want them to have to go school by school in order to do so. 

@rkodev has already completed the backend work on his side for allow for these changes to be made. He'll add some more context in the comments below.

ciremusyoka:
Closes #1491 
Allows users to export student at any jurisdiction level.


ciremusyoka:
ScreenShot
export-student-list

moshthepitt:
@ciremusyoka would it have been possible to use the drill-down-table component that is already included in the project instead of an entirely new one?  I think that each new package that we add needs to absolutely necessary.

ciremusyoka:
> @ciremusyoka would it have been possible to use the drill-down-table component that is already included in the project instead of an entirely new one? I think that each new package that we add needs to absolutely necessary.

@moshthepitt drill down table could not be applicable here. To me i felt it was necessary

src/components/forms/LocationSelect/index.tsx
```diff
@@ -1,11 +1,123 @@
...
+  const [loadingId, setLoadingId] = useState<string | null>(null);
+  const [openNodes, setOpenNodes] = useState<string[]>([]);
+
+  const getJurisdiction = (
+    parentId: string = '',
+    key: string = '',
+    clickedNodes: string[] = []
+  ) => {
+    if (loadingId) {
+      return;
```
moshthepitt, line 35: Is this what we want to happen when it is loading?

ciremusyoka: Yes

ciremusyoka: Generally it ensures we load features of one location at a time

moshthepitt: What I meant is: do we need any indication that something is loading?

src/components/forms/LocationSelect/index.tsx
```diff
@@ -1,11 +1,123 @@
...
+      {({ items }) => (
+        <React.Fragment>
+          {items.map(({ key, ...itemProp }) => (
+            <Row key={`row-${key}`}>
+              <Col>
+                <ItemComponent key={key} {...itemProp} />
+              </Col>
+              <Col className="download-list">
+                {loadingId === itemProp.id ? (
+                  <div className="loading-text">{LOADING}...</div>
```
ciremusyoka, line 101: @moshthepitt we have a loading indicator here

moshthepitt (approved):

kelvin-muchiri (approved):

Issue #1492, opened 2021-02-09
Plan assignment failing for MDA Point Plans 


isabelcshaw:
Right now, the assignment feature is broken for MDA Point plans. This was working fine when tested about two or so weeks ago. I am not sure if it is relevant, but a new hierarchy for Eswatini was uploaded by Akros last week. 

Here is the workflow that was tested: 

1. Created a MDA Point plan and assigned it to Eswatini  (MDA-Point IS 2021-02-09) 
2. Activated the plan 
3. Went to the Assign page and selected the plan 
4. The assignment page loads, but you are only able to assign a user to Eswatini - you are not able to click through to lower levels of the hierarchy to assign a plan to any other geography. Also, as the page loads, you see an error that says: "Map failed to load: jurisdiction not found" 

Here is a video of that interaction: 
https://user-images.githubusercontent.com/12836913/107395143-67e8d580-6ab1-11eb-8091-3c041201db68.mov

ciremusyoka:
@isabelcshaw Plans created from planning page should be assigning lower jurisdictions before activating them. The plan you mentioned above is only assigned top jurisdiction only i.e. Eswatini that is why you can't drill down the hierarchy.

PR #1508, opened 2021-02-22
Fix error appearing sometimes when assigning teams

ciremusyoka:
Part of #1492 
We expect filtered out jurisdictions to have corresponding nodes which are filtered from the location hierarchy. In some cases when matching the nodes and the jurisdictions we end up  having some jurisdictions without corresponding nodes. And since the assumption is every jurisdiction has a corresponding node we try to access the values of missed nodes and end up with type error(Can't get value x of undefined). 
This pr fixes this bug by filtering out jurisdictions without corresponding nodes.


moshthepitt:
@ciremusyoka could we get some more explanation on how this solves the problem?  Basically I would like to read in this PR or a commit message that: X is the problem, caused by Y and this is how it is being fixed.

ciremusyoka:
Update the description

moshthepitt:
@ciremusyoka thanks.  The description is pretty good.  Next step is to get these into the commit messages whenever possible. :)

moshthepitt (approved):

p-netm (approved):

Issue #1507, opened 2021-02-22
Enable form configuration for TH-Preview


Naima-Bashir:
https://smartregister.atlassian.net/browse/RVL-339
Requirement

Enable form configurability for all Thailand Focus forms (in English and in Thai) in TH preview first
Once we’ve test form configuration in Preview we can enable form configurability for Thailand production
Dependency

Relevance files upload bug is fixed - https://github.com/OpenSRP/opensrp-server-web/issues/688

Also depends on this: 

https://github.com/OpenSRP/web/issues/379  

The updates so far are:
When a user creates a plan-changes the definition url for family registration task-saves and activates the plan- assigns his team-logs into android - registers a family- the changes he had made aren't shown.
From @githengi investigation it ​looks like the web Ui did not update the definition uri on the plan. The plan is still pointing to the default family_register.json 

![image](https://user-images.githubusercontent.com/61143692/108679608-d8561480-74fd-11eb-81bb-53e3ec05439f.png)

Naima-Bashir:
a meeting with the team from both Reveal and Opensrp to discuss the way forward on resolving a bug that was caused by some implementation on the OpenSRP side

Naima-Bashir:
Ask client if he published the draft.

Naima-Bashir:
Updates on the JSON uri bug: 

looks like the web Ui did not update the definition uri on the plan. The plan is still pointing to the default family_register.json

kelvin-muchiri:
It was agreed in order to resolve the bug, Reveal would cease to use its redux store and use the opensrp store. Since this is not the ideal solution, we'll research how to make changes on package `@opensrp/form-config` so that it does not require any other project that uses the package to use the opensrp store

kelvin-muchiri:
@Naima-Bashir We encountered another bug on Friday, things did not go as smoothly as planned...We are still trying to find a solution for it with @p-netm

kelvin-muchiri:
@Naima-Bashir The bug was fixed and I have a PR open for review

kelvin-muchiri:
I will be handling the second part which  has to do with translations

kelvin-muchiri:
@Naima-Bashir I have added the strings to be  translated on the doc 41-72

samkanga:
@kelvin-muchiri please share the link with the strings to be translated on this ticket.

kelvin-muchiri:
[...]

PR #1512, opened 2021-02-24
Update form config

kelvin-muchiri:
Addresses #1507 

Update form configuration to use the latest version of `@opensrp/form-config`, `@opensrp/store`

The latest version of `@opensrp/form-config` currently requires the store of used to be `@opensrp/store` in order to get the configs for multilanguage support. The Reveal store is refactored to use `@opensrp/store`


ciremusyoka (commented):
@kelvin-muchiri should the configs for the translations on the `form-config` package be dispatched somewhere?

kelvin-muchiri:
@ciremusyoka Right...which languages does Reveal support?

ciremusyoka:
> @ciremusyoka Right...which languages does Reveal support?

This package should support Thai

yarn.lock
```diff
@@ -2,6 +2,41 @@
 # yarn lockfile v1
 
 
+"@ant-design/colors@^6.0.0":
+  version "6.0.0"
+  resolved "https://registry.yarnpkg.com/@ant-design/colors/-/colors-6.0.0.tgz#9b9366257cffcc47db42b9d0203bb592c13c0298"
+  integrity sha512-qAZRvPzfdWHtfameEGP2Qvuf838NhergR35o+EuVyB5XvSA98xod5r4utvi4TJ3ywmevm290g9nsCG5MryrdWQ==
+  dependencies:
+    "@ctrl/tinycolor" "^3.4.0"
```
moshthepitt, line 10: @kelvin-muchiri must these be installed?

moshthepitt (changes requested):
Currently not comfortable with intalling antd here.

src/App/App.tsx
```diff
@@ -143,6 +146,14 @@ import { oAuthUserInfoGetter } from '../helpers/utils';
...
 import './App.css';
 
+/** register reducer for configs */
+reducerRegistry.register(configsSliceName, configsReducer);
+const opensrpLanguageConfig = {
+  languageCode: LANGUAGE as LanguageCode,
+};
+/** Dispatch language to use for opensrp packages */
+store.dispatch(addConfigs(opensrpLanguageConfig));
+
```
p-netm, line 156: @kelvin-muchiri here unfortunately the workflow here, requires we dispatch the configs before importing the components, we should hoist this functionality to the very top of this file.

yarn.lock
kelvin-muchiri: @moshthepitt Sure, I had added `antd` as a peer dependency, but for seems like at some point I might have moved it again as dependency....

kelvin-muchiri: @moshthepitt We had a discussion with @p-netm & @KipSigei to resolve this, we are faced with the following quick options:

1. Reveal to allow this since Antd is used by `@opensrp/form-config` and some of its opensrp dependencies
2. Move Antd in `@opensrp/config` and its opensrp dependencies into peer depencies

For options that will take time:

1. Move shared utilities into a new package `@opensrp/form-config-core` and have separate packages for the bootstrap and Antd components
2. Use one package with either bootstrap or Antd components and modify the look in the imported project through css

moshthepitt: Wasn't the initial intention for this to work without polluting Reveal with antd?

> Move shared utilities into a new package @opensrp/form-config-core and have separate packages for the bootstrap and Antd components

I was actually under the impression that this is what had been done all along especially because we knew form config was going to be used in Reveal sooner or later.

> Move Antd in @opensrp/config into peer dependencies. Do this too for opensrp packages installed by @opensrp/form-config particularly @opensrp/notifications

What's the reason that these packages do not already have antd as a peer dependency?

kelvin-muchiri: > Wasn't the initial intention for this to work without polluting Reveal with and?

@moshthepitt   Reveal components were not polluted...I added Antd components in a new directory but moved shared code into a utilities file within the same package

moshthepitt: > Reveal components were not polluted...I added Antd components in a new directory but moved shared code into a utilities file within the same package

I meant polluted in the sense that every Reveal user would have to pay the cost of downloading antd when interacting with the web app for no (obvious or not) benefit.

kelvin-muchiri: > I was actually under the impression that this is what had been done all along especially because we knew form config was going to be used in Reveal sooner or later.

Moving Antd to peer dependencies seemed straight forward. But other opensrp packages have it installed as a dependency, in this case `@opensrp/notifications` installed in `@opensrp/form-config`. Therefore every opensrp package should have Antd as a peer dependency and not as a dependency?

moshthepitt: I don't have an answer for that, but I would say:

> Therefore, should every opensrp package have Antd as a peer dependency and not as a dependency?

1. Is there a reason why not?
2. Doing re-usable components stop making sense if using just one such component means pulling down a huge dependency tree.  If we have fully committed to reusable packages then this is something we must keep in mind.

kelvin-muchiri: @moshthepitt Ok...Then for now looks like we shall take 

> Move Antd in @opensrp/config into peer dependencies. Do this too for opensrp packages installed by @opensrp/form-config particularly @opensrp/notifications

moshthepitt: Cool.  Let's pass it by @p-netm and @KipSigei for a sanity check.

kelvin-muchiri: @moshthepitt This will not work either since the components using Antd are failing.....Seems like the next option is 
> Move shared utilities into a new package @opensrp/form-config-core and have separate packages for the bootstrap and Antd components

moshthepitt: @kelvin-muchiri what's an ETA for that?  I want to loop in @samkanga

kelvin-muchiri: @moshthepitt  I'd give it 1 day

yarn.lock
```diff
@@ -9274,6 +9358,16 @@ immer@1.10.0:
   resolved "https://registry.yarnpkg.com/immer/-/immer-1.10.0.tgz#bad67605ba9c810275d91e1c2a47d4582e98286d"
   integrity sha512-O3sR1/opvCDGLEVcvrGTMtLac8GJ5IwZC4puPrLuRj3l7ICKvkmA0vGuU9OW8mV9WIBRnaxp5GJh9IEAaNOoYg==
 
+immer@^8.0.0:
```
moshthepitt, line 9361: Where is this used @kelvin-muchiri ?

moshthepitt (approved):

ciremusyoka (approved):

yarn.lock
kelvin-muchiri: @moshthepitt `@opensrp/pkg-config` uses `@reduxjs/toolkit` which depends on `immer`

Issue #1500, opened 2021-02-17
IRS progress bar color should still be green if > 100%


moshthepitt:
![MACEPA-IRS-Lite-Western-Live-Plan-Lubosi-UHC-Reveal](https://user-images.githubusercontent.com/372073/108171514-aa8f5b00-710c-11eb-8ded-aa8660598173.png)

PR #1547, opened 2021-04-06
IRS progress bar color should still be green if > 100%

haseemVD:
This PR addresses #1500


haseemVD:
Created new PR with signed commit
#1548

PR #1548, opened 2021-04-06
IRS progress bar color should still be green if > 100%

### LINKED ISSUE #1500: IRS progress bar color should still be green if > 100%

haseemVD:
this PR fixes #1500

### DIFF
--- a/src/containers/pages/IRSLite/MapLite/index.tsx
+++ b/src/containers/pages/IRSLite/MapLite/index.tsx
@@ -375,6 +375,7 @@ const IRSLiteReportingMap = (
                   {!row.listDisplay && <p className="indicator-description">{row.description}</p>}
                   {!row.listDisplay && (
                     <ProgressBar
+                      lineColor={indicatorThresholdsIRSLite.GREEN_THRESHOLD.color}
                       lineColorThresholds={indicatorThresholdsIRSLite || null}
                       value={row.value}
                     />
--- a/src/containers/pages/IRSLite/MapLite/tests/__snapshots__/index.test.tsx.snap
+++ b/src/containers/pages/IRSLite/MapLite/tests/__snapshots__/index.test.tsx.snap
@@ -103,7 +103,7 @@ Array [
     cssClass="progress-bar-striped"
     decimalPoints={0}
     height="10px"
-    lineColor="#0000FF"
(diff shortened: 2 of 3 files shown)


src/containers/pages/IRSLite/MapLite/index.tsx
```diff
@@ -375,6 +375,7 @@ const IRSLiteReportingMap = (
                   {!row.listDisplay && <p className="indicator-description">{row.description}</p>}
                   {!row.listDisplay && (
                     <ProgressBar
+                      lineColor={indicatorThresholdsIRSLite.GREEN_THRESHOLD.color}
```
moshthepitt, line 378: IIRC this will set the default line color to Green - right?  I'm not sure that is what we want in this case.  Instead, we should leave the default "as is" and instead ensure that when the value is > 100%, the color rendered is green.

haseemVD: @moshthepitt I have tested this implementation, 
Issue: 
- Before reaching 100, it shows gray color, and when >100 it shows blue color which was set to default.

So changing default color worked.

moshthepitt: Thanks @haseemVD - is it possible to capture that in a test? i.e. that it is grey before 100 and then green when > 100?

moshthepitt (approved):
Nice work

src/containers/pages/IRSLite/MapLite/index.tsx
haseemVD: @moshthepitt Added test,

Just a thought:
- We are using `ProgressBar` from `@onaio/progress-indicators`, So these color changing tests should be written in that package to check if thats working fine or not when passing `lineColor` as a porp.

moshthepitt: From the Real point of view, we care about the color being grey or green or whatever else.  So that's how I would justify writing these tests on this side.  The tests on the `@onaio/progress-indicators` side should be to ensure that that component works as intended in general.

ciremusyoka (approved):

Issue #1524, opened 2021-03-08
Missing Translations on WebUI Mapview


samkanga:
There are some missing translations on the WebUI Map View, for the historical index case issue. See the screenshot below:

![image](https://user-images.githubusercontent.com/5908630/110280361-bb384000-7feb-11eb-8219-6cf5715f2257.png)

haseemVD:
@samkanga Added point 73 in [file](https://docs.google.com/spreadsheets/d/1ONHpPpuPxNW9_-vP8gTTcrVHTbKTT3WaN-cfdySeMCs/edit?folder=1MN5J-PK6KUOHC6gw3KKmf12vV-GQFPkr#gid=209283218), 
Need translation to close this issue.

moshthepitt:
fixed by #1553

Naima-Bashir:
Hi @ciremusyoka cc @samkanga 
i've tested this and it has passed QA. 
![image](https://user-images.githubusercontent.com/61143692/116988885-b20cdc00-acd9-11eb-8991-e4a768311228.png)

Quick question, the word RACD is still in english, should it be translated or?

PR #1553, opened 2021-04-15
remove hardcoded string and use from lang file

haseemVD:
This PR addresses #1524


ciremusyoka (changes requested):
This string should also be translated to Thai. Right?

haseemVD:
> This string should also be translated to Thai. Right?

Added comment on [issue](https://github.com/onaio/reveal-frontend/issues/1524#issuecomment-820270941), Will update PR after getting translation.

ciremusyoka (approved):

moshthepitt (approved):

PR #1556, opened 2021-04-20
MDA Point Reporting: Child level views showing impossible data

### LINKED ISSUE #1555: MDA Point Reporting: Child level views showing impossible data
At the child level of the MDA Point reporting views, there are indicators about each individual child's drug administration. These should indicate if a child received drugs and which ones - or, if they opted out and why.

Right now, these views are showing "Yes" for indicators/columns 3-6, even though this is impossible from a data perspective. First, we know not all of those children have had an mda task completed, second, they can't have an mda task completed AND have refused the drug or opted out bc of a condition. And the ADR reported is an edge feature that will not have high usage and is its own form. Finally, these indicators don't match with the mda coverage counts on the higher levels of the same reports - many of them show 0 for number of drugs dispensed. This was originally reported in issue: [#1908](https://github.com/onaio/canopy/issues/1908)

Here is a screenshot of what this looks like:
image

After discussion with @baokiddy and @gstuder-ona it is possible that this is happening because there are null values in the base view. @moshthepitt is it possible that we are showing "yes" as a response to "null"? @baokiddy null is the correct value for no information has yet been synced, correct?

Comment from @ciremusyoka : @moshthepitt on the web we have yes as long as the value is not equal 0. See [here](https://github.com/onaio/reveal-frontend/blob/f5211c508be3f95de81e4f824a8c3c3b3c2762b3/src/containers/pages/MDAPoint/ChildReports/index.tsx#L198-L202).

haseemVD:
This PR closes #1555


moshthepitt:
@haseemVD could we have a screenshot please?

moshthepitt (changes requested):
Could we add a test please?

haseemVD:
> @haseemVD could we have a screenshot please?

Screenshot 2021-04-20 at 4 52 34 PM

moshthepitt:
@haseemVD is it possible to get a screenshot example where we have both Yes and No values?  @ciremusyoka @samkanga can you help identify a plan that has data like this?

haseemVD:
> @haseemVD is it possible to get a screenshot example where we have both Yes and No values? @ciremusyoka @samkanga can you help identify a plan that has data like this?

@moshthepitt Here it is, 
Also adding tests for this.
Screenshot 2021-04-20 at 5 23 15 PM

src/containers/pages/MDAPoint/ChildReports/tests/index.test.tsx
```diff
@@ -106,4 +106,28 @@ describe('components/MDA Reports/MDAPlansList', () => {
...
+        mmaalbgiven: 0,
+        mmanodrugadminreason: null,
+        mmapzqdosagegiven: 'test',
+        plan_id: '40357eff-81b6-4e32-bd3d-484019689f7c',
+        sactacurrenroll: 0,
+        sactanationalid: '2345',
+        school_location_id: '154153',
+      },
+    ];
+    const tableData = [['test user', '2345', 'No', 'No', 'No', 'No', 'No', 'test', 0]];
```
moshthepitt, line 130: Lets show a Yes in the test

moshthepitt (approved):

moshthepitt (changes requested):
@haseemVD @ciremusyoka please verify that this PR adheres to the directions here https://github.com/onaio/reveal-frontend/issues/1555#issuecomment-823434451

src/containers/pages/MDAPoint/ChildReports/index.tsx
```diff
@@ -68,6 +68,27 @@ const tableHeaders = [
...
+ */
+export const extractChildData = (data: ChildReport[]) => {
+  return data.map((sch: ChildReport) => {
+    return [
+      `${sch.client_first_name} ${sch.client_last_name}`,
+      sch.sactanationalid ? sch.sactanationalid : '',
+      sch.sactacurrenroll ? YES : NO,
+      sch.mmadrugadmin ? YES : NO,
+      sch.mmanodrugadminreason ? YES : NO,
+      sch.mmanodrugadminreason ? YES : NO,
```
ciremusyoka, line 84: Let's confirm if we need to use same data for two columns. I.e `SACs Refused` and `SACs Sick/Pregnant/Contraindicated`

ciremusyoka:
> @haseemVD @ciremusyoka please verify that this PR adheres to the directions here [#1555 (comment)](https://github.com/onaio/reveal-frontend/issues/1555#issuecomment-823434451)

@moshthepitt @haseemVD we need to add a check to covere `We should never have a case where you show either both "yes" or both "no" to "MMA drug administered" and
"SACs Refused"`

moshthepitt:
> we need to add a check to covere We should never have a case where you show either both "yes" or both "no" to "MMA drug administered" and "SACs Refused"

@ciremusyoka this should ideally be done in the database queries.  The frontend is not meant to hold significant logic for the reports.  Please work with @samkanga to get this to @baokiddy  or @gstuder-ona

src/containers/pages/MDAPoint/ChildReports/index.tsx
haseemVD: @isabelcshaw @samkanga

haseemVD: @baokiddy @gstuder-ona

ciremusyoka (approved):

haseemVD:
Updated ss:

Screenshot 2021-04-22 at 6 15 17 PM
Screenshot 2021-04-22 at 6 15 34 PM

@moshthepitt

moshthepitt (approved):

Issue #1574, opened 2021-05-05
Enforce Automated Selection of Subvillage Classification When Creating Routine Plans does not work 


Naima-Bashir:
Go to thailand preview 
Create new plan https://web.reveal-th-preview.smartregister.org/plans/new 
Select the OA
Observe the Action plan name
Previously it used to automatically have a name 
Now until one clicks on it, nothing happens 
However, after one starts typing in, it automatically fills in the name 
Connected to https://github.com/onaio/reveal-frontend/issues/1482

![image](https://user-images.githubusercontent.com/61143692/117118440-6e79a700-ad99-11eb-8ab5-1f0212893305.png)

ciremusyoka:
@samkanga @Naima-Bashir The name was auto populating on selecting the `Local Term`. Now that it is being updated automatically i think that is what is causing this issue.

samkanga:
@ciremusyoka can we make it auto populate when the classification is selected?

PR #1594, opened 2021-05-19
Enforce Automated Selection for Plan title

haseemVD:
This PR addresses issue #1574 

https://user-images.githubusercontent.com/68631902/118799699-774d9b00-b8b8-11eb-91e3-928047afee4e.mov


haseemVD:
> This PR addresses issue #1574
> 
>  Screen.Recording.2021-05-19.at.3.34.07.PM.mov

@samkanga Please confirm if the shown behaviour is what we are expecting.

samkanga:
@haseemVD I checked this out and it works as expected. Thank you.

src/components/forms/JurisdictionSelect/index.tsx
```diff
@@ -208,6 +210,26 @@ export const promiseOptions = (
...
   });
 
+/**
+ * Sets value in name and title field
+ * @param e Form event
+ * @param form Formik form Object
+ */
+export const onSelectionComplete = (e: FormEvent<Element>, form: FormikProps<any>) => {
+  const target = e.target as HTMLInputElement;
+  const fieldsThatChangePlanTitle = ['interventionType', 'fiStatus', 'date', 'jurisdictions[0].id'];
```
moshthepitt, line 220: Is there a way to have this code shared between this module and the [similar thing](https://github.com/onaio/reveal-frontend/blob/08d57148cc60210d60c5b10718970bd5e508a372/src/components/forms/PlanForm/index.tsx#L397) in the PlanForm component?

src/components/forms/PlanForm/index.tsx
```diff
@@ -204,6 +204,26 @@ export interface PlanFormProps {
   beforeSubmit: BeforeSubmit /** called before submission starts, return true to proceed with submission */;
 }
 
+/**
+ * Sets value in name and title field
+ * @param e Form event
+ * @param form Formik form Object
+ */
+export const handlePlanTitleChange = (e: FormEvent<Element>, form: FormikProps<any>) => {
```
moshthepitt, line 212: Let's move this to the helpers module if possible.

src/components/forms/PlanForm/helpers.ts
```diff
@@ -943,3 +943,23 @@ export const onSubmitSuccess = (
   setSubmitting(false);
   setAreWeDoneHere(true);
 };
+
+/**
+ * Sets value in name and title field
+ * @param e Form event
+ * @param form Formik form Object
+ */
+export const handlePlanTitleChange = (e: FormEvent<Element>, form: FormikProps<any>) => {
```
ciremusyoka, line 952: Lets do away with `any` please.

haseemVD: updated

src/components/forms/JurisdictionSelect/index.tsx
```diff
@@ -269,6 +271,14 @@ export const handleChangeWithOptions = (
           if (fiStatusFieldName) {
             const val = optionVal.fiStatus || NOT_AVAILABLE;
             form.setFieldValue(fiStatusFieldName, val);
+            form.setFieldTouched(fiStatusFieldName, true);
+            const updatedForm = { ...form };
+            updatedForm.values.jurisdictions[0] = { id: optionVal.value, name: optionVal.label };
```
moshthepitt, line 276: Could you explain this line?  It seems fragile to me in that it specifically targets element 0.  What if there are more, or less?

haseemVD: Confirmed with @ciremusyoka, There can be multiple jurisdictions, Updating in next commit.

src/components/forms/JurisdictionSelect/index.tsx
```diff
@@ -267,8 +269,22 @@ export const handleChangeWithOptions = (
           form.setFieldValue(field.name, optionVal.value);
           form.setFieldTouched(field.name, true);
           if (fiStatusFieldName) {
+            const jurisdictionIdx = field.name
+              .split('[')[1]
+              .split(']')[0]; /** get jurisdiction index from field name prop */
```
moshthepitt, line 274: What are we trying to do here?  I'm not sure I like how this looks and doubt if it will be not fragile.  We should take a step back and think if this can be solved/achieved in a different way.

haseemVD: @moshthepitt The name prop of the field contains the index of jurisdiction, So we can get index from name prop and change its value.

ciremusyoka:
Closed by #1614

PR #1617, opened 2021-06-11
Guard navigation to disabled routes

### LINKED ISSUE #1551: Audit routes and ensure users can't navigate to disabled routes
Currently users are able to navigate to to disabled routes which should not be the case.

ciremusyoka:
Closes #1551
This pr updates `@onaio/connected-private-route` package to limit navigation to only enabled routes and adds 404 page.

### DIFF
--- a/src/App/App.tsx
+++ b/src/App/App.tsx
@@ -25,6 +25,23 @@ import Loading from '../components/page/Loading';
 import {
   BACKEND_ACTIVE,
   DISABLE_LOGIN_PROTECTION,
+  ENABLE_ASSIGN,
+  ENABLE_CONFIG_FORM,
+  ENABLE_DYNAMIC_MDA,
+  ENABLE_FI,
+  ENABLE_IRS,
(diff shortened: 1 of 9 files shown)


src/App/App.tsx
```diff
@@ -231,34 +250,44 @@ const App = () => {
                 <ConnectedPrivateRoute
                   redirectPath={APP_CALLBACK_URL}
                   disableLoginProtection={DISABLE_LOGIN_PROTECTION}
+                  routerEnabled={ENABLE_MDA_POINT}
+                  routerDisabledRedirectPath={PAGE_NOT_FOUND_URL}
```
moshthepitt, line 254: Should we create an issue to look into making this more DRY?  Maybe by using a provider or some other thing?

moshthepitt (approved):

src/App/App.tsx
ciremusyoka: #1618

p-netm (approved):

Issue #1595, opened 2021-05-24
Dockerize Reveal Web


bennsimon:
To deploy reveal web on kubernetes a docker image of the setup is required.

PR #1623, opened 2021-06-21
Dockerize Reveal Web

ciremusyoka:
Part of #1595

Initial work to dockerize web app to support deployment using k8s


docker/confd_env.toml
```diff
@@ -0,0 +1,145 @@
+[template]
+src = "config.js.tmpl"
+dest = "/usr/src/web/config.js"
+keys = [
+  "/generate/sourcemap",
+  "/skip/preflight/check",
```
moshthepitt, line 6: How did you populate this?

ciremusyoka: @moshthepitt they are generated from the envs we have. e.g for env `REACT_APP_NAME` we have `/react/app/name`.

moshthepitt (approved):

Dockerfile
```diff
@@ -0,0 +1,70 @@
+FROM alpine/git AS sources
+
+RUN git clone --depth=1 --branch=config-sessions https://github.com/onaio/express-server.git /usr/src/express-server
```
KipSigei, line 3: use master branch

KipSigei (dismissed):

Dockerfile
moshthepitt: @KipSigei @ciremusyoka why not create a release tag and use that?

KipSigei (dismissed):

Dockerfile
ciremusyoka: @moshthepitt does this mean i have to find a way of passing the tag when building the image and maybe in case a tag is not specified default to master?

moshthepitt: It means you'll replace `--branch=config-sessions` with `--branch=your-tag-here`

moshthepitt (approved):

KipSigei (approved):

